Issue #11, opened 2015-06-16
dataPath of additionalProperties is empty


shachr:
i would like to remove all additional properties but dataPath is always empty when err type is additionalProperties!

epoberezkin:
I suspect I may have recently fixed the issue with incorrect dataPath reporting in version 0.4.9 (it was issue #9).

Please try using the latest version (0.4.10 - there is another important fix). If you are still seeing the issue please post your schema, data and results (errors) here.

Thanks for helping me making it better ;)

epoberezkin:
Ok, I've figured what it was about.
The choice I've made was to put the object path in dataPath rather than the property, because it is the object that is being validated rather than its property.
So if you had additionalProperty keyword in some sub-object, you would have it's path in dataPath.
Although this is more consistent with the rest of the validation keywords, I agree that it is useful to have the actual additional property in error.dataPath.

So I added it in 0.4.15.

```
var schema = {
  properties: {
    foo: {},
    bar: {}
  },
  additionalProperties: false
};

var data = { foo: 1, bar: 2, baz: 3, quux: 4 };

var ajv = Ajv({allErrors: true});
var valid = ajv.validate(schema, data);
console.log(ajv.errors);
```

There will be two errors in the array, with dataPaths `"['baz']"` and `"['.quux']"`. (It would have been easy to replace them with `".baz"` and `".quux"`, but as it's not a compile time replacement, I decided to keep it faster - you can map/slice your properties out of these errors.

So if you want to remove additional properties you could do smth like this:

```
ajv.errors
.filter(function(err) { return err.keyword == 'additionalProperties'; })
.map(function(err) { return err.dataPath.slice(2,-2); })
.forEach(function(p) { delete data[p]; });
```

It'll be just a bit more complex for properties in sub-objects.

PR #23, opened 2015-07-28
Add ability to filter an object based on the schema

andyscott:
How reasonable would it be fully support a flag that would filter any offending properties when `additionalProperties` is set to false? In short-- instead of returning an error, I want the value(s) removed. Ideally a 'warning' would be emitted instead of an error.

I took a very preliminary stab at getting this working for my immediate needs. Consider this PR a starting point for discussion.

Thanks!

### DIFF
--- a/lib/dot/definitions.def
+++ b/lib/dot/definitions.def
@@ -1,11 +1,11 @@
 {{## def.setup:_keyword:
-  {{
+  {{ 
     var $lvl = it.level
       , $dataLvl = it.dataLevel
       , $schema = it.schema[_keyword]
       , $schemaPath = it.schemaPath + '.' + _keyword
-      , $breakOnError = !it.opts.allErrors
-      , $removeAdditional = it.opts.removeAdditional;
+      , $breakOnError = !it.opts.allErrors;
+
 
     var $data = 'data' + ($dataLvl || '')
       , $valid = 'valid' + $lvl
--- a/lib/dot/definitions.def
+++ b/lib/dot/definitions.def
@@ -1,11 +1,11 @@
 {{## def.setup:_keyword:
-  {{ 
+  {{
     var $lvl = it.level
       , $dataLvl = it.dataLevel
       , $schema = it.schema[_keyword]
       , $schemaPath = it.schemaPath + '.' + _keyword
-      , $breakOnError = !it.opts.allErrors;
-
+      , $breakOnError = !it.opts.allErrors
+      , $removeAdditional = it.opts.removeAdditional;
 
     var $data = 'data' + ($dataLvl || '')
       , $valid = 'valid' + $lvl
--- a/lib/dot/properties.jst
+++ b/lib/dot/properties.jst
@@ -48,9 +48,12 @@ var valid{{=$it.level}} = true;
       }}
       {{? $noAdditional }}
         valid{{=$it.level}} = false;
-
+        {{? $removeAdditional }}
+        delete {{=$data}}[key{{=$lvl}}];
+        {{??}}
         {{# def.error:'additionalProperties' }}
         {{? $breakOnError }} break; {{?}}
+        {{?}}
       {{??}}
         {{ /* additionalProperties is schema */
           $it.schema = $aProperties;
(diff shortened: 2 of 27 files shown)


epoberezkin:
:+1: 

I think it's very cool, it's quite common feature for validators. I was reluctant to implement it thinking that it is probably out of scope, and you can remove additional properties by analysing errors (see #11 e.g.)... But analysing errors is a hassle and it's definitely easier to do it in the process of validation as your code shows.

Could you please do a couple of things (or I can):
1. move $removeAdditional var to properties.jst, so it's not defined for all keywords
2. remove additional properties if additionalProperties is a schema (it's below in the same properties.jst) and if there is no additional properties keyword at all.
3. recompile dots using 0.10 node as it doesn't insert these /**/ - I still hope it's some V8 fluke that will go away.
4. update readme
5. update version in package.json
6. Ideally, some test(s) for the option - in options.spec.js (nonexistent at the moment :).

andyscott:
Sounds good. I went ahead and added the tests.

I haven't been able to figure out how to support the schema values for `additionalProperties`.

In [properties.jst at line 69](https://github.com/epoberezkin/ajv/blob/65ccf49edfb0ebdeb06990b51a8c6cab57e35a1c/lib/dot/properties.jst#L69) the call to validate the nested schema modifies the `errors` and `validate.errors` variables. I don't quite know how to captures those nested errors and instead emit them as warnings.

epoberezkin:
Cool, I've noticed :)
I think warnings is a complication, don't you think? I would simply remove quietly with maybe optional logging (say if removeAdditional is set as 'log').

If you skip the test that's failing, I can merge and finish.

Or you can simply move this logic a bit outside - removing will be instead of what is happening between https://github.com/andyscott/ajv/blob/feature/filter-object-to-schema/lib/dot/properties.jst#L47 and https://github.com/andyscott/ajv/blob/feature/filter-object-to-schema/lib/dot/properties.jst#L73 - essentially removing will take precedence over failing in all cases - no error reporting will happen in this case (apart from logging maybe)

epoberezkin:
Also I think there is no point validating nested schema if we are going to remove the property anyway

andyscott:
Don't you have to first validate it to see if the property is offending and
needs removal?

On Wed, Jul 29, 2015 at 2:48 PM, Evgeny Poberezkin <notifications@github.com

epoberezkin:
yes, it will be inside `if (isAdditional{{=$lvl}}) { ... }`. With removeAdditional there will be removal code, without - validating. It doesn't matter if it is offending or not - if it is additional - it will be removed.

Or you only want to remove additional properties if there is `additionalProperties: false` in schema?

I thought it should always remove them with this option regardless of the schema for addtionalProperties. Although it is possible to differentiate this behaviour, if there is some use case, and remove only failing additional properties if the option value is 'failing' and remove all additional if the option is true. In this case you would have to validate nested schema first indeed if there is schema...

But I think it is an unnecessary complication, to be honest. I would simply always remove until you or somebody else has some specific use case where differentiation is needed. What do you think?

epoberezkin:
Looking at your test you do want to differentiate. Anyway, I merged and added 'all' for removeAdditional (see readme). 

You can add 'failing' if you want, but not necessary. Maybe I'll add it later. Essentially you have to preserve the value of errors (the counter) before schema validation and then restore it if schema failed and you want to remove additional property rather than fail validation. Array should also be reset. See anyOf keyword for example - it does that.

Thanks!

andyscott:
Awesome!

And you're right-- I think I was a bit confused (or just didn't explain myself properly). Originally I set out to just remove additional fields that failed validation. I should have named the property `removeInvalidAdditionalFields` or something more descriptive.

Either way, removing all additional fields covers my use case, since I only set `additionalFields = false` in my schemas.

epoberezkin:
it now can do both - either all or only those where additionalProperties is false. I didn't implement only removing properties for the failing schemas.

Thanks

epoberezkin:
@andyscott see #26 - funny issue came out of this feature. Make sure to use 0.6.11 - it works correctly at   least when your $schema points to draft 4 meta-schema (or absent).

PR #28, opened 2015-08-06
Escape schema strings in output

blakeembrey:
Run into this while trying to benchmark the library. Need to make sure any strings have the single quotes escaped.

### DIFF
--- a/lib/dot/definitions.def
+++ b/lib/dot/definitions.def
@@ -126,13 +126,13 @@
 
 
 {{## def._errorMessages = {
-  $ref:            "'can\\\'t resolve reference {{=$schema}}'",
+  $ref:            "'can\\\'t resolve reference {{=it.util.escapeQuotes($schema)}}'",
   additionalItems: "'should NOT have more than {{=$schema.length}} items'",
   additionalProperties: "'additional properties NOT allowed'",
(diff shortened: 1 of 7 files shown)


lib/dot/definitions.def
```diff
@@ -144,21 +144,21 @@
   multipleOf:      "'should be multiple of {{=$schema}}'",
   not:             "'should NOT be valid'",
   oneOf:           "'should match exactly one schema in oneOf'",
-  pattern:         "'should match pattern \"{{=$schema}}\"'",
+  pattern:         "'should match pattern \"{{=it.util.escapeQuotes($schema)}}\"'",
   required:        "'property {{=$missingProperty}} is required'",
-  type:            "'should be {{? $isArray }}{{= $typeSchema.join(\",\") }}{{??}}{{=$typeSchema}}{{?}}'",
+  type:            "'should be {{? $isArray }}{{= it.util.escapeQuotes($typeSchema.join(\",\")) }}{{??}}{{=it.util.escapeQuotes($typeSchema)}}{{?}}'",
```
epoberezkin, line 149: no need escaping types - they don't have quotes (schema validation will fail if they do)

epoberezkin:
:+1: 

types don't need escaping though - removed

Thanks!

PR #29, opened 2015-08-10
Add test coverage command

### LINKED ISSUE #7: Add test coverage

blakeembrey:
Closes #7.

Subjective note: I did change it to bail the tests when a failure occurs. I find it's easier to work with, but I wouldn't be offended if you don't want that :smile: 

Edit: This is the standard test format I use. The command `test-spec` is a mirror of `test-cov` without code coverage - useful because istanbul rewrites the source code and it can be harder to debug at times.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,9 @@
   "description": "Another JSON Schema Validator",
   "main": "lib/ajv.js",
   "scripts": {
-    "test": "./node_modules/.bin/mocha --reporter=spec spec/*.spec.js"
+    "test-spec": "mocha spec/*.spec.js -R spec --bail",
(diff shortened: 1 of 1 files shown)


epoberezkin:
Thank you very much!
Yeah, I usually let it shows all errors :)
But it's not because I prefer it - I actually just alwayS did it this way :)
So I'll leave with bail, probably it's easier indeed.

So now I just need to push the coverage data to https://codeclimate.com/github/epoberezkin/ajv to get a badge ;)

epoberezkin:
Removed --bail :)
It's easier to see all the things that break in this case rather than just the first one that breaks - it suggests what could be changed to fix.

PR #33, opened 2015-08-21
Latest schema v4

blakeembrey:
Some unrelated things in the PR, but all separate commits so you can cherry pick if needed.

Summary: The specification lists four different URLs that should be validated against the latest v4 specification, as of today. I've added the other three for compatibility with current JSON schemas out there.

### DIFF
--- a/lib/ajv.js
+++ b/lib/ajv.js
@@ -259,8 +259,17 @@ function Ajv(opts) {
 
 
     function addInitialSchemas() {
-        if (self.opts.meta !== false)
-            addMetaSchema(require('./refs/json-schema-draft-04.json'), META_SCHEMA_ID, true);
+        if (self.opts.meta !== false) {
+            var metaSchema = require('./refs/json-schema-draft-04.json');
+
+            addMetaSchema(metaSchema, META_SCHEMA_ID, true);
+
+            // Add other schemas according to JSON specification:
+            // http://json-schema.org/latest/json-schema-core.html
+            addMetaSchema(metaSchema, 'http://json-schema.org/schema', true);
+            addMetaSchema(metaSchema, 'http://json-schema.org/hyper-schema', true);
+            addMetaSchema(metaSchema, 'http://json-schema.org/draft-04/hyper-schema', true);
+        }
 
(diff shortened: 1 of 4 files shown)


lib/ajv.js
```diff
@@ -259,8 +259,17 @@ function Ajv(opts) {
...
-            addMetaSchema(require('./refs/json-schema-draft-04.json'), META_SCHEMA_ID, true);
+        if (self.opts.meta !== false) {
+            var metaSchema = require('./refs/json-schema-draft-04.json');
+
+            addMetaSchema(metaSchema, META_SCHEMA_ID, true);
+
+            // Add other schemas according to JSON specification:
+            // http://json-schema.org/latest/json-schema-core.html
+            addMetaSchema(metaSchema, 'http://json-schema.org/schema', true);
+            addMetaSchema(metaSchema, 'http://json-schema.org/hyper-schema', true);
```
epoberezkin, line 270: hyper meta-schema is different from meta-schema - it has additional keywords (and they are not implemented). I suggest not adding it, because it will increase the size of browser bundle. It can be manually added using addMetaSchema by users when needed.

epoberezkin:
Cool. Please see my comment in code.

lib/ajv.js
blakeembrey, line 270: You're right, feel free to remove the `hyper-schemas`. By the way, maybe you could make a note about `draft-03` compatibility in the README? I tried it (it was a pipe dream), but it failed to compile or work. I ended up using `json-schema-compatibility` to get it to work instead (https://github.com/mulesoft-labs/osprey-method-handler/blob/master/osprey-method-handler.js#L301-L305)

epoberezkin:
Thanks for all that - that's great. Added you to collaborators, hope you don't mind, - you can commit directly to a branch of this repo and merge PRs from them.

blakeembrey:
No worries. I'll definitely stick to PRs though, the conversations are valuable :smile:

epoberezkin:
Looked at this compatibility. Yes, can be mentioned in README.

Issue #32, opened 2015-08-18
JSON Pointers


blakeembrey:
I'm curious, but why did you use a custom notation instead of using [JSON Pointers](https://tools.ietf.org/html/rfc6901) in error objects?

epoberezkin:
I don't think I was choosing really. I just used JavaScript property access syntax as it can be easily used in the code that processes errors (either with eval or with a library like object-path). And the same (or similar) syntax is also used internally in generated code.

I can see the use case where errors are consumed outside of JavaScript environment (node.js server accessed by non-javascript client). While I don't think it's common, I don't mind having an option say `errorsJsonPointer` or smth like this that would make generated code report errors using JSONPath.

blakeembrey:
That would be cool, I'll look into it if I get a chance. For reference, http://jsonapi.org/ and other formats use pointers as the primary notation format.

Edit: Having a consistent format can be important and, at least, the errors I'm generating is for an API and, most likely, will be consumed outside of JS.

epoberezkin:
That's a relatively easy change - wherever `it.errorPath` or `$it.errorPath` is appended (in doT folder) with path segment using JS syntax, they should be appended using JSONPointer syntax if `it.opts.errorsJSONPointer` is present.

`it.errorPath` maintains the code for "validation-time" (as opposed to compile-time) expression that will evaluate to the path of the data currently validated. It is only used to report errors, nothing else relies on it. So changing it to a different format won't affect anything.

I've just discovered and removed unused `it.dataPath` from everywhere - at some point it was used but not anymore, so if you decide to do it please use the latest commit.

Thanks!

epoberezkin:
@blakeembrey I've found one more case where jsonPointers should be used - fixed.

By the way, with json pointers it reports error message slightly differently in this case (see https://github.com/epoberezkin/ajv/blob/master/spec/errors.spec.js#L93). I think it's not important to be honest.

epoberezkin:
Thanks!

PR #36, opened 2015-08-21
Support JSON pointers

blakeembrey:
Attempting #32. Please review this carefully, I'm not 100% if this is the right approach (took me long enough to the hang of the templating and what was happening).

Edit: Got to do some more tidy up to add the `.replace('" + "', '')` behaviour back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ Options can have properties `separator` (string used to separate errors, ", " by
 - _unicode_: calculate correct length of strings with unicode pairs (true by default). Pass `false` to use `.length` of strings that is faster, but gives "incorrect" lengths of strings with unicode pairs - each unicode pair is counted as two characters.
 - _beautify_: format the generated function with [js-beautify](https://github.com/beautify-web/js-beautify) (the validating function is generated without line-breaks). `npm install js-beautify` to use this option. `true` or js-beautify options can be passed.
 - _cache_: an optional instance of cache to store compiled schemas using stable-stringified schema as a key. For example, set-associative cache [sacjs](https://github.com/epoberezkin/sacjs) can be used. If not passed then a simple hash is used which is good enough for the common use case (a limited number of statically defined schemas). Cache should have methods `put(key, value)`, `get(key)` and `del(key)`.
-
+- _jsonPointers_: Output `dataPath` using JSON Pointers instead of JS path notation.
 
 ## Tests
 
--- a/lib/compile/util.js
+++ b/lib/compile/util.js
@@ -15,7 +15,10 @@ module.exports = {
   cleanUpCode: cleanUpCode,
   cleanUpVarErrors: cleanUpVarErrors,
   schemaHasRules: schemaHasRules,
-  stableStringify: require('json-stable-stringify')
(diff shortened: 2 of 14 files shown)


blakeembrey:
@epoberezkin Ok, tidied up the output. Strings are maximally joined together at compilation time by using single quotes consistently (I think previously `missingProperty` would not have joined because it used the escape method which uses single quotes).

Edit: I used the key `useJsonPointers` as it seemed more natural.

epoberezkin:
The option can be even shorter by the way: `{ jsonPointers: true }`.

It looks like I added all the comments in your branch, rather than in PR... It got them here anyway, but without code context. 

Do you want to me to fix all those issues or you'd rather look at them yourself? I don't mind either way.

epoberezkin:
Actually maybe it does differentiate - just noticed that you call two different functions - `getPath` and `getPathExpr` :) Let me have a look again

epoberezkin:
It's impressive that you've figured how it works by the way... I get lost almost every time :)

epoberezkin:
Ok, so apart from ~~two places~~ one place it uses `getPathExpr` for validation time variable names and `getPath` for compile time key/index. Am I right? Is it an error in ~~those two places~~ this one place :) or am I missing something?

epoberezkin:
@blakeembrey please comment on my comments - I've finished reviewing. Great stuff.

blakeembrey:
> Ok, so apart from one place it uses getPathExpr for validation time variable names and getPath for compile time key/index. Am I right?

Yes, pretty much. I just based the code on the type of code it was replacing though, so depending on which syntax you used I replaced it with the the runtime or compile time path expression.

> The option can be even shorter by the way

Using that now.

> it will escape quotes twice. is it intentional here?

I don't think that was intentional. I should get some tests for that later.

> We should keep both the current test (without json pointers) and the new one

I guess you could add another test with it enabled. I just changed only `fullAjv` tests since those errors would have been tested already using the regular `ajv` object.

> I would rename it to toQuotedString; toString

Fixed.

> Also the behaviour when normal properties were appended as .key instead of ['key'] is lost.

All the same behaviour is there for keys. Some more complex nested tests probably wouldn't hurt though.

I'm guessing all the comments related to compile time and validation time keys is ok now? If I missed something or the logic is wrong, feel free to fix it up. I mostly based it on what was already there.

epoberezkin:
Yes, it's great. Can I merge or you are going to make another commit?

epoberezkin:
I see that I can :)

epoberezkin:
@blakeembrey should we maybe add "#" in front of JSON pointers?

blakeembrey:
@epoberezkin Why's that? That would be an implementation detail that doesn't have anything to do with JSON pointers.

epoberezkin:
They usually start with #, no?

blakeembrey:
@epoberezkin Nope, never. You might be thinking of when it's used as a URL in a reference within JSON schemas? That's only because it's the URL fragment.

epoberezkin:
Ah, I see. Thanks for explaining...

Issue #40, opened 2015-08-25
Custom formats - asynchronous validation


simon-p-r:
Hi

Can custom formats be async?  

Thanks
Simon

epoberezkin:
Not really. Validation is synchronous. What's the use case?

simon-p-r:
Ok I use z-schema and it provides these hooks for you, if you are performing any meaningful validation you have to do IO to lookup up values

epoberezkin:
I will think about asynchronous...

I think that "format" is something that can be clearly and independently defined. If something requires IO it is not a "format", it is some other custom validation.
I don't think custom validation should be done via JSON-schema. It undermines schema platform independence.

I can reconsider under pressure though :)

simon-p-r:
What about a format that requires a webservice to validate it?  Such as http://ec.europa.eu/taxation_customs/vies/faq.html

epoberezkin:
I assume you are talking about VAT identification number, correct me if I am wrong. There are two parts in its validation. The first is "format" validation - it is clearly defined in Q11 and can be defined via regex or a function. The second is establishing that it is valid based on some proprietary algorithm or based on whether it is assigned to some company. I think it should be validated in a different application layer from the one that receives the data and establishes its validity to the JSON scheme, because it belongs to the business logic. Mixing IO / data validation with business logic validation is not the most efficient thing to do. JSON-schema is not designed for business logic validation.

epoberezkin:
Credit card number is similar, by the way. Format validation will just look for 16 digits with optional spaces. Business logic validation can look up database or access the web service. The first should be in the schema, the second should not be.

simon-p-r:
I see your unwilling to make change so will close issue, what is Q11?

epoberezkin:
Question 11 in the document you've posted. It has format specifications for VAT identification number in different countries. I think this is what the schema should validate against.

At best the change is not very easy to make - changing compiled functions to be asynchronous is quite an effort. So it won't happen very soon. If you really need async, z-schema is a very good choice, because although z-schema is much slower, with async the validation speed is not that important - it will be a tiny fraction of I/O anyway. At the same time z-schema is a very mature and battle-tested validator.

I've just tried to suggest you to consider an alternative approach, but it is up to you of course...

But keep it open; as I said, I will think about it, I am not 100% certain about the right approach to asynchronous and custom validations. I may implement them at some point.

simon-p-r:
[...]

PR #43, opened 2015-09-01
Add run time lookup of schemas

abalmos:
Adds the ability to register functions that resolve a schema, on-the-fly, when a reference is not found.

In my particular use case, there are a large number of available schemas but only a small subset is needed for any one instantiation of the application.

It certainly makes sense to allow asynchronous lookups depending on the outcome of #40. One obvious use case is downloading a remote schema definition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,17 @@ Although `addSchema` does not compile schemas, explicit compilation is not requi
 
 By default the schema is validated against meta-schema before it is added, and if the schema does not pass validation the exception is thrown. This behaviour is controlled by `validateSchema` option.
 
+##### .registerSchemaLookup(Function lookup)
+
+Register a synchronous `lookup` function that takes the desired reference (id) and returns the matching schema at runtime. This is useful when one has a large number of schemas and loading them at once results in too large of an overhead.
+
+The `lookup` function should return either:
+
+- the schema in a format accepted by .addSchema()
+- a falsey value indicating the schema was not found
+
+If more than one `lookup` function is registered, each will called in the order they where registered until one returns a schema.
+
 
 ##### .addMetaSchema(Object schema [, String key])
 
--- a/lib/ajv.js
+++ b/lib/ajv.js
@@ -30,12 +30,14 @@ function Ajv(opts) {
     this._refs = {};
     this._formats = formats(this.opts.format);
     this._cache = this.opts.cache || new Cache;
+    this._lookups = [];
 
     // this is done on purpose, so that methods are bound to the instance
     // (without using bind) so that they can be used without the instance
     this.validate = validate;
     this.compile = compile;
     this.addSchema = addSchema;
+    this.registerSchemaLookup = registerSchemaLookup;
     this.addMetaSchema = addMetaSchema;
     this.validateSchema = validateSchema;
     this.getSchema = getSchema;
@@ -100,6 +102,15 @@ function Ajv(opts) {
         schemaObj.meta = _meta;
     }
 
+    /**
+     * Registers a function that can be used to add a schema on demand
+     * @param {Function} lookup A function that translates a `key` to a schema on demand
+     */
+    function registerSchemaLookup(lookup) {
+        if (typeof lookup !== 'function') throw new Error('lookup should be a function');
+
+        self._lookups.push(lookup);
+    }
 
     /**
      * Add schema that will be used to validate other schemas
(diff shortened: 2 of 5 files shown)


epoberezkin:
Cool stuff. Several questions.
1. Why it is needed to have multiple lookup functions? I think one would suffice per ajv instance, in cases when not, the loop can exist outside of ajv, inside the supplied function.
2. I would pass it via options, if anything.
3. What about asynchronous loading? This is not covering it.
4. What about referenced schemas that are missing? I think that is more common use case. I was going to solve it with throwing exceptions that contain missing schema inside.

I think the whole logic, apart from part 4, can exist outside of ajv:

```
var validate = ajv.getSchema(schemaRef);
if (validate) {
    var valid = validate(data);
    continue(valid);
} else {
    loadSchema(schemaPef, function(err, schema) {
        validate = ajv.compile(schema);
        valid = validate(data);
        continue(valid);
    });
}
```

Not sure why this should be inside of ajv to be honest.

abalmos:
> What about referenced schemas that are missing? I think that is more common use case. I was going to solve it with throwing exceptions that contain missing schema inside.

This is actually the my motivating use case. Somehow, after all that, I didn't write a test case for missing refs and completely missed that my changes don't do what I really wanted after all...

I think I actually prefer your suggestion of just throwing an exception that has the refs of the missing schema(s).. I'm not that comfortable with the code lib/compile yet. Is there an obvious place to add it?

Below are responses to your other questions for perspective.

> Why it is needed to have multiple lookup functions? I think one would suffice per ajv instance, in cases when not, the loop can exist outside of ajv, inside the supplied function.

My particular use case is a library that has an unknown number of additional packages, each loaded at the will of the application programmer, that may or may not bring in additional schemas (some packages may even $ref schemas from other packages). There needs to be some way of loading them after the AJV object is instantiated.

> I would pass it via options, if anything

The same issues as above

> What about asynchronous loading? This is not covering it.

It doesn't appear asynchronous validation is supported by ajv. It is not clear to me how I could block ajv in an acceptable way while waiting for an async on-the-fly load.  If #40 results in asynchronous validating then I would gladly extend the lookup to support it as well.

abalmos:
@epoberezkin It looks like I can already sort of do what I need using the scheme you proposed. With options: { missingRefs: 'fail', verbose: true } ajv.validate returns false on a missing reference and ajv.errors indicates what schema id was missing.

A couple of issues/questions:
1. It would be nice if all the errors encountered were in the ajv.errors array. It looks like maybe that was the intention but only the last errors seems to survive?
2. It would be nice if the ajv.errors array was attached to the Error object before it is thrown (assuming missingRefs is not set). 
3. {missingRefs: 'fail'} causes ajv to console.log the error message. With 2. this isn't an issue.
4. The errors array is apparently lost (?) when calling the validate function returned by getSchema.

My apologizes if I am missing something obvious. I am new to dot.

epoberezkin:
No problem, thank you for the suggestions.
1. there is options allErrors: true that will force ajv to continue validation and collect all errors. The default behaviour is to return false on the first error.

2, 3. I think you are confusing schema compile time with validation time. Or I didn't understand what you meant. My intention was (I can actually do it today) during compilation to throw an error with the first missing schema (that will be the default behaviour). Then you can handle asynchronous loading with the code like this:

```
function compileSchema(schema, callback) {
    try { validate = ajv.compile(schema); }
    catch(e) {
        if (e.missingSchema) resolveMissingSchema(e.missingSchema);
        else callback(e);
        return;
    }
    callback(null, validate);

    function resolveMissingSchema(missingSchema) {
          loadSchema(missingSchema, function(err, sch) {
              if (err) callback(err);
              else {
                  compileSchema(sch, function(err) {
                      if (err) callback(err);
                      else compileSchema(schema, callback);
                  });
              }
          });
    }
}
```

Probably it will be a bit more elegant with promises. Also it needs some protection against situation when loadSchema returns a wrong schema that doesn't help compiling original schema so the same error is thrown again and again. I may actually include it in ajv - it's a bit not obvious....

 4 I am not sure I understand what you mean by lost. `ajv.error` will be overwritten by the result of the next validation, either `null` or the new errors array. If you want to use it you need to save it to some variable. Or did you mean something else?

epoberezkin:
I've added `missingRef` and `missingSchema` properties to the thrown error. I am going to either add callback and loadSchema parameters (functions) to compile method or create another method that will do asynchronous compilation loading missing schemas on the way (it will be similar to the code above - try compiling, see what's missing, try again...). It's not directly related to asynchronous validation in #40 though. Compiled schema will be synchronous.

epoberezkin:
@abalmos If `compileAsync` that I've done in 1.3 is covering it, I will close this PR

abalmos:
@epoberezkin It looks like everything I need. I should get back to this soon and when I do I will report back.

Thanks for the getting it in for me.

epoberezkin:
Thanks, you're welcome.

abalmos:
With #51 and #52 `compileAsync` does exactly what I need!

PR #46, opened 2015-09-06
Equal fail when order of keys is different

YChebotaev:
In section [3.6](http://json-schema.org/latest/json-schema-core.html#anchor9) equality of objects are defined as they are have the same set of property names, but not same order.
The equal function and it's test will fail if order of keys are different, but javascript [does not](http://stackoverflow.com/a/5525820) guarantee order of keys in object.

So i fixed this.

### DIFF
--- a/lib/compile/equal.js
+++ b/lib/compile/equal.js
@@ -17,8 +17,17 @@ module.exports = function equal(a, b) {
   if (arrA != arrB) return false;
 
   if (a && b && typeof a === 'object' && typeof b === 'object') {
+    var keysSet = {};
     var keys = Object.keys(a);
-    if (!equal(keys, Object.keys(b))) return false;
+    var keys2 = Object.keys(b);
+    if (keys.length !== keys2.length) return false;
+    for (i = 0; i < keys.length; i++) {
+      keysSet[keys[i]] = 1;
+    }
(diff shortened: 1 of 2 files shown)


blakeembrey:
Just a note, but I think this would be much simpler solved with a loop that exits once a key is in one object and not the other. However, even a `sort` would work here though there'd be a minor overhead. E.g.

``` js
var keys = Object.keys(a);

if (keys.length !== Object.keys(b).length) {
  return false;
}

for (var i = 0; i < keys.length; i++) {
  if (!Object.prototype.hasOwnProperty.call(b, keys[i]) {
    return false;
  }
}
```

YChebotaev:
@blakeembrey Yes. I think it's better solution. Thanks.

epoberezkin:
Thanks for that! Funny bug that was. I don't know what I was thinking :)  I guess because it's only used in uniqueItems and it's kind of inefficient to use it with objects, nobody hit it so far.

epoberezkin:
I optimised a bit. It's safe (and much faster) to compare with undefined here as the data comes from JSON. I thought to merge two loops in one but it will be only marginally faster when there are the same properties and much slower in all other cases, so I kept them separate.

Issue #18, opened 2015-06-29
Errors for 'required' validation should include missing property


cdm6a:
Hi, I'm trying to get some clarity on a potential issue I'm seeing. In running the following:

```
const ajv = Ajv({verbose: false, allErrors: true});

ajv.addSchema([
   require('./schemas/Currency.json'),
   require('./schemas/PositiveFloatString.json'),
   require('./schemas/Amount.json'),
   require('./schemas/Party.json'),
])

function isValid(object, schemaName) {
  return ajv.validate(schemaName, object);
}

function validate(object, schemaName) {
  if (!isValid(object, schemaName)) {
    return {
      isValid: false,
      errors: ajv.errors
    };
  }
  return {
    isValid: true
  };
}

console.log(validator.validate({
        entity: 'acct:conner@ropple.com',
        amount: {
          currency: 'USD'
        }
      }, 'Party'));
```

I get the following:

```
{ 
  isValid: false,
  errors: [{ 
    keyword: 'required',
     dataPath: '.amount',
     message: 'properties value, currency are required' 
  }]
}
```

Ideally I would like to see something more along the lines of:

```
{ 
  isValid: false,
  errors: [{ 
    keyword: 'required',
    dataPath: '.amount.value',
    message: 'property value is required' 
  }]
}
```

I was wondering if this was an issue with my implementation / schemas or if it is a shortcoming of ajv?

Thanks in advance.

My schemas are as follows:
Party.json

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "Party",
  "title": "Party",
  "description": "Represents a single party to a payment.",
  "type": "object",
  "properties": {
    "name": {
      "$ref": "Account"
    },
    "amount": {
      "description": "How much this party will send or receive in the payment.",
      "$ref": "Amount"
    }
  },
  "additionalProperties": false,
  "required": ["name", "amount"]
}
```

Amount.json

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "Amount",
  "title": "Amount",
  "type": "object",
  "properties": {
    "value": {
      "$ref": "PositiveFloatString"
    },
    "currency": {
      "type": "string",
      "$ref": "Currency"
    }
  },
  "additionalProperties": false,
  "required": ["value", "currency"]
}
```

PositiveFloatString.json

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "PositiveFloatString",
  "title": "PositiveFloatString",
[...]

PR #55, opened 2015-09-25
remove redundant property in required error message

marbemac:
For required properties, the error message is currently:

"data.name should have required property .name"

This works for other validations, where the property is present, but required is a special case that indicates the property is not there. The above message is somewhat confusing. I've simply updated the required errors to read:

"data should have required property .name"

### DIFF
--- a/lib/dot/definitions.def
+++ b/lib/dot/definitions.def
@@ -150,7 +150,7 @@
   not:             "'should NOT be valid'",
   oneOf:           "'should match exactly one schema in oneOf'",
   pattern:         "'should match pattern \"{{=it.util.escapeQuotes($schema)}}\"'",
-  required:        "'should have required property {{=$missingProperty}}'",
+  required:        "'is a required property'",
   type:            "'should be {{? $isArray }}{{= $typeSchema.join(\",\") }}{{??}}{{=$typeSchema}}{{?}}'",
   uniqueItems:     "'should NOT have duplicate items (items ## ' + j + ' and ' + i + ' are identical)'"
 } #}}
--- a/lib/dot/required.jst
+++ b/lib/dot/required.jst
@@ -31,7 +31,7 @@
           , $missingProperty = '\' + ' + $propertyPath + ' + \'';
         it.errorPath = it.opts.jsonPointers
                         ? it.util.getPathExpr($currentErrorPath,  $propertyPath, true)
-                        : $currentErrorPath;
+                        : $currentErrorPath + ' + ' + $propertyPath;
       }}
       {{# def.error:'required' }}
     } else {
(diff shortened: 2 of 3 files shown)


marbemac:
I see the failing assertion.. if this change is accepted, then I suppose that assertion can be removed for the required validation. Happy to do that.

epoberezkin:
this is contradictory to the wish to have a separate error for each required property and to have missing property in dataPath (e.g. https://github.com/epoberezkin/ajv/issues/18). I think the better solution is to change the message to `data.name is a required property`, i.e. to change the message rather than the dataPath

marbemac:
Fair enough! All set - let me know if I've missed anything.

epoberezkin:
Thanks!

You've done it all correctly. I am just still thinking how to address this issue better... The old messages are better in case they are used separately (that's how I use them). In case they are connected using `errorsText` they are obviously wrong - that's what your fix is improving... To be honest I wasn't fully sure about the changes that addressed #11 and #18. But I didn't think about the `errorsText` at the time (not sure if it even existed).

Your first solution may be better and in this case missingProperty can be made available as some additional parameter to those who needs them. There already exists an additional property of the error object `params` that is added if `i18n` option is used - these params are used to construct internationalised messages. So maybe I should have dataPath pointing to the object rather than to the property and rename `i18n` option to `params` - they can be useful for other use cases too.

Reverting dataPath to point to object for both `required` and `additionalProperties` keywords will be a change that breaks backward compatibility for some users, but probably it is worth it because it is more correct - for these keywords the error is in the object, not in the property.

What do you and others (@blakeembrey, @shachr, @cdmcnamara) think?

blakeembrey:
I had trouble parsing everything, so I might follow up, but some initial thoughts. I prefer `required` and `additionalProperties` and everything else to use the exact path to the issue (the property that is missing, not the object). I think it's more useful in generating error messages and parsing - in any case, a consumer can always split it and go backwards but it's trickier to get the path from the message itself. 

I guess I'll have to relook into `params` now, I haven't updated to use the latest `i18n` code to evaluate it properly.

epoberezkin:
I merged the message change while I am thinking what to, message is wrong anyway in the current situation. I am thinking more and more that dataPath should point to the object (that's where the problem is) and some extra property should have the name of the missing (or additional) property.

epoberezkin:
@blakeembrey another argument for switching `dataPath` to point to object rather than to property is that at the moment `dataPath` and `data` in error point to different things (`data` - object, `dataPath` - property). In all cases apart from required and additionalProperties they point to the data currently validated... So I think in 2.0 I'll preserve existing behaviour with some option (for backwards compatibility) and the default behaviour will be to make `dataPath` point to the object and report the property in params.

Still not sure what's the better thing to do :)

PR #60, opened 2015-10-14
Do not force two slashes for uris

johanlelan:
To allow relative urls (conformed to https://tools.ietf.org/html/rfc1808), URI regex do not force two slashes after scheme.


epoberezkin:
It would be good to add some test that would fail without this change here: https://github.com/epoberezkin/ajv/tree/master/spec/tests/issues

epoberezkin:
The standard for URI is [RFC3896](http://tools.ietf.org/html/rfc3986) according to this http://json-schema.org/latest/json-schema-validation.html#anchor123

johanlelan:
First of all, thanks for this useful library! 
I just want that your regular expression takes every URIs values. 
In the section 3 "Syntax component" of RFC3896, the author underlines "When authority is not present, the path cannot begin with two slash characters ("//")."

epoberezkin:
Thank you. It's merged :) I will publish in the evening

epoberezkin:
@johanlelan It's on npm

PR #130, opened 2016-02-24
use browserified & uglified version

bearmini:
Hi, I have modified bower.json to use browserified & uglified version of source code.


epoberezkin:
makes sense, thanks

epoberezkin:
pushed the tag, you should have it soon

PR #132, opened 2016-02-26
add dist for bower/npm users

bearmini:
Currently dist/ directory and its contents only generated by running `npm bundle`. This PR includes the artifacts to the repo and allows bower/npm users to use dist/ajv.min.js etc. without running the command.


epoberezkin:
npm includes dist already, it is generated during `npm publish`. For bower it should be done in a similar way - there should be post-install command that would generate bundles. I don't like the idea of including bundles in the repo because it completely distorts changelog and stats.

bearmini:
OK, I'll drop this PR and create another with post-install way.

bearmini:
Reopen this PR.

In order to build bundle files, all users who want to install this package via bower should have npm installed on their environment. Can we assume all the bower users have npm command?

I found this SO:
http://stackoverflow.com/questions/17170500/how-make-bower-build-the-package

According to this, 

> Building on install is an anti-pattern and is strongly recommended against in Node. Like Node, Bower packages should be pre-built. This is because the end-user shouldn't have to care what preprocessor or build-system a package requires.

Should we go 'post-install' option though? Just want to clarify.

bearmini:
I found an answer for my own question. Bower can be installed only using npm, so all bower users must have npm installed. We can go post-install option safely.

bearmini:
Reopen again.

As [bower document](http://bower.io/docs/config/#hooks), 

> These hooks are not intended to provide a post-installation build step for component authors. As such, the configuration for these hooks is provided in the .bowerrc file in the parent project’s directory.

The postinstall script was not triggered even if I have .bowerrc in ajv repository. The 'postinstall' is for projects that uses ajv.

epoberezkin:
> Building on install is an anti-pattern and is strongly recommended against in Node.

npm recommends limiting the use of post-install script to environment-specific compilation and it provides an alternative mechanism for all other build steps - building when publishing.

Bower has post-install hook but they advise against using them for building packages... Unfortunately, although they may have plans to provide storage for compiled packages (according to the same answer by one of bower contributors that you quote) the only viable alternative to using the hook they offer is building manually. I am sorry that it causes you some inconvenience, I don't think that putting compiled code under source control is a better alternative.

We will have to decide whether we simply show the compiling instructions when the package is installed (post-install script will be some "echo ...") or build automatically disregarding their recommendation. I think just showing instruction can be a bit better, but I am happy with any of two options that you think is more convenient. If bower creates storage for packages we can change it.

The other alternatives you can consider are:
- using [compiled ajv bundles on cdnjs](https://cdnjs.com/libraries/ajv).
- using npm/require/browserify to build your application - we used it with angular apps without any problems.

epoberezkin:
Ah, so you say it doesn't even trigger post-install... Sorry, I missed it. We can't even show the instruction then. I guess that inconvenience may be the reason in some decline in bower usage...

epoberezkin:
Is it possible to publish to bower from the branch somehow? And the branch would be pushed to from travis build process whenever master branch changes.

Tags exist without connection to branches and bower seems to pick up the latest tag anyway... But maybe it is somehow possible to limit bower to the branch? Not sure.

bearmini:
Thanks for your comments. For a short-term solution, I'll add instructions how to build bundles manually. I'll also look into whether we can publish a branch to bower or not. 

By the way, we probably move to npm/require/browserify for our Angular project.

epoberezkin:
Thank you. I also thought some instructions are needed.

PR #148, opened 2016-03-21
Minimal typescript definition file for ajv.js

mrchief:
This would allow `ajv.js` to be used in TypeScript projects without errors/warnings.

Of course, it provides no type checking but that can be rolled in gradually.

### DIFF
--- a/ajv.d.ts
+++ b/ajv.d.ts
@@ -1,0 +1,9 @@
+// Type definitions for ajv
+// Project: https://github.com/epoberezkin/ajv
+// Definitions by: mrchief <https://github.com/mrchief>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare module 'ajv' {
+  function Ajv(opts?: any): any;
+  export default Ajv;
+}


epoberezkin:
Shouldn't it be at least `function Ajv(opts?: object): object;` ?

epoberezkin:
Also it seems like the methods can be defined too...

blakeembrey:
@epoberezkin Don't merge this, it's wrong. It should be an external module to work for TypeScript users. I can work with you to get a complete definition in for you, we've been maintaining one already: https://raw.githubusercontent.com/enriched/typed-ajv/66c35ae67b4e2792f0caf8db70e6ed5d937a8f94/ajv.d.ts.

blakeembrey:
@epoberezkin I can work with you on this. For more information, there's https://github.com/typings/typings/issues/322.

epoberezkin:
Cool. I was actually about to ask you about it as I saw you are involved in typings :) I had a feeling that it's at best a stub and simply masks warnings, which I don't think is a good thing. But you seem to have it all worked out already :)

epoberezkin:
@mrchief thanks for starting it. I was going to write that it needs to be done properly, not just a stub. Thanks again.

epoberezkin:
@blakeembrey so is it better if it is inside ajv package (as you write in typings/typings#322) or if it is a separate thing? I can see a couple of corrections that are needed in the type definitions https://github.com/enriched/typed-ajv/blob/master/ajv.d.ts. - some options should use union-type `string | boolean`.

blakeembrey:
@epoberezkin Definitely more valuable to have it here, TypeScript users will be able to use it immediately. Let's open a new PR to collaborate on those fixes and get it supported :+1:

mrchief:
@epoberezkin It was meant to be a starting point, not the final thing. Unfortunately, I didn't have more time to put into this but I'm glad @blakeembrey has this going. In light of his work, I would say don't bother merging this.

Would love to see the typings getting added soon though!

blakeembrey:
@mrchief Just for reference, type definition are meant to be external modules and not ambient definitions. That's the biggest issue you had here, since making it ambient doesn't actually make it any easier to someone using TypeScript to consume it.

Edit: I'll try to do a PR tomorrow with the suggested changes.

epoberezkin:
Thank you @blakeembrey @mrchief

epoberezkin:
@blakeembrey, sorry to chase you :), can I help with this one somehow maybe?

PR #150, opened 2016-03-22
Rename exclusiveFormat to formatExclusive (v5 draft)

lyngaas:
Unless there exists a more recent v5 proposal for these keywords, the names should be:
- formatExclusiveMinimum
- formatExclusiveMaximum

https://github.com/json-schema/json-schema/wiki/formatMinimum-(v5-proposal)

### DIFF
--- a/KEYWORDS.md
+++ b/KEYWORDS.md
@@ -16,7 +16,7 @@ The keywords and their values define what rules the data should satisfy to be va
     - [maxLength/minLength](#maxlength--minlength)
     - [pattern](#pattern)
     - [format](#format)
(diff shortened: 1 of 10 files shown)


epoberezkin:
Oh my... I guess I liked my version more :) It never occurred to me they are different.
It probably should be a major version change...
Thanks!

epoberezkin:
btw, given that it's not implemented anywhere yet (as far as I know) maybe we could simply edit the proposal...

@geraintluff what do you think?

epoberezkin:
Thank you!

PR #157, opened 2016-03-31
Added more NPM keywords

garrows:
When I was looking on NPM for json schema validators I never came across this one. Looking at the keywords I can see why. Thought I might not be the only one.


epoberezkin:
Thanks! So more people will find it :)

PR #165, opened 2016-04-17
TypeScript type definitions - corrections

epoberezkin:


epoberezkin:
@blakeembrey , could you please review these changes?

There are a couple of things I don't understand.
1. Why return values of the methods are sometimes separated with ":" (e.g. [this line](https://github.com/epoberezkin/ajv/pull/165/files#diff-89794cf8c2ab6af000576e99672b9b2cL5) and many other places) and sometimes with "=>" (e.g. [this line](https://github.com/epoberezkin/ajv/pull/165/files#diff-89794cf8c2ab6af000576e99672b9b2cL39))?
2. [ErrorParameters](https://github.com/epoberezkin/ajv/blob/master/lib/ajv.d.ts#L68) interface looks as if it is an object with many optional properties. In reality it is one of the definitions below (MinMaxParam | AdditionalPropertyParam | ...) depending on the keyword property in ErrorObject. Is it possible to express it or should alteration ("|") be used in this case?
3. Why Dependency and PatternGroup in ErrorParameters have "[]" after them? Probably I should read the manual more attentively...

epoberezkin:
Also, would that be a correct syntax?

```
interface FormatValidator {
  string | RegExp | (data: string) => boolean
}
```

blakeembrey:
@epoberezkin Quick TypeScript primer. You have primitive types: `number`, `string`, `void`, `any`, `boolean`. Includes `any[]` (array syntax) and `() => any` (function syntax). On top of that, you can write interfaces:

``` ts
interface Foo {
  (): any // `Foo` is a function.
  method (): any // Has a `method` method.
  prop: () => any // Equivalent to `method` method, but using our primitives.
  x: string // Another prop.
  [key: string]: string // Dictionary.
}
```

So, interfaces can kind of describe anything. On top of that, you have type expression (E.g. unions and intersections, etc.) which can be used in any type position _or_ with the `type` keyword.

``` ts
type Bar = () => any
```

Now let's use an expression to demonstrate where expressions can be valid.

``` ts
function foobar (x: string, y: { [key: string]: Foo }): Bar
```

blakeembrey:
> Also, would that be a correct syntax?

No, you want:

``` ts
type FormatValidator = string | RegExp | (data: string) => boolean;
```

> Why return values of the methods are sometimes separated with ":"

Either works, as described above.

> Is it possible to express it or should alteration ("|") be used in this case?

Yes, it's possible to express. TypeScript has string literal types that could be used here, but it'll probably get verbose.

``` ts
interface ErrorObject1 {
  keyword: 'foo'
  params: {
    x: number
    y: number
  }
}

type ErrorObject = ErrorObject1 | ...
```

> Why Dependency and PatternGroup in ErrorParameters have "[]" after them?

Array syntax.

blakeembrey:
I think everything in this PR looked valid (we could hook up a test using `tsc` on the `.d.ts` file), but you may want to replace some of those `Object` usages with index signatures (`{ [key: string]: Foo }`).

epoberezkin:
@blakeembrey thank you very much!

When I run `./node_modules/.bin/tsc lib/ajv.d.ts` on [this file](https://github.com/epoberezkin/ajv/blob/typescript/lib/ajv.d.ts) I get these errors:

```
lib/ajv.d.ts(44,31): error TS1005: ')' expected.
lib/ajv.d.ts(44,33): error TS1131: Property or signature expected.
lib/ajv.d.ts(44,39): error TS1005: ';' expected.
lib/ajv.d.ts(44,41): error TS1128: Declaration or statement expected.
lib/ajv.d.ts(45,10): error TS1109: Expression expected.
lib/ajv.d.ts(46,20): error TS1109: Expression expected.
lib/ajv.d.ts(47,19): error TS1109: Expression expected.
lib/ajv.d.ts(48,16): error TS1109: Expression expected.
lib/ajv.d.ts(49,17): error TS1109: Expression expected.
lib/ajv.d.ts(50,18): error TS1109: Expression expected.
lib/ajv.d.ts(51,25): error TS1109: Expression expected.
lib/ajv.d.ts(52,19): error TS1109: Expression expected.
lib/ajv.d.ts(53,14): error TS1109: Expression expected.
lib/ajv.d.ts(54,14): error TS1109: Expression expected.
lib/ajv.d.ts(55,11): error TS1109: Expression expected.
lib/ajv.d.ts(58,49): error TS1005: ')' expected.
lib/ajv.d.ts(58,57): error TS1005: ';' expected.
lib/ajv.d.ts(58,59): error TS1128: Declaration or statement expected.
lib/ajv.d.ts(185,1): error TS1128: Declaration or statement expected.
```

I installed with `npm install typescrpt`
It seems like it doesn't like anonymous functions in parameters...
What am I missing?

Thank you again :)

blakeembrey:
Odd, worked for me from `master`. Can you make sure it wasn't your PR you tested?

Edit: Sorry, just saw we are checking the PR branch.

blakeembrey:
@epoberezkin I forgot to mention, for `Promise` support, you'll want to add some promise typings locally for it to work properly. TypeScript < ES6 does not have a promise type, and the promise type also requires a generic parameter.

blakeembrey:
^ As does `Array`. It's a generic, should be `Array<any>`, for example.

blakeembrey:
Finally, for the union with an anonymous function use-case, wrap it in some extra parens for parsing:

``` ts
transpile?: string | ((code: string) => string);
```

epoberezkin:
@blakeembrey thanks a lot, all works now! I just added `type Promise = Object;` and I was wondering whether it should import this instead: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/es6-promise/es6-promise.d.ts ?

Thank you again :)

epoberezkin:
@blakeembrey Ok, I've figured re Promise. Should've read your comments more attentively :)

blakeembrey:
@epoberezkin Where are you using promises, I want to check if we can improve this. Otherwise, non-ES6 TypeScript clients will get confused because `Promise<>` does not exist for them. Is it just you accept something that returns promises or do you actually return promises yourself? If it's the former, we can fix this and inline a type definition that's compatible for you (TypeScript is structural, it'll be clearer that way).

epoberezkin:
@blakeembrey Promises are used for [asynchronous validation](https://github.com/epoberezkin/ajv#asynchronous-validation). Ajv supports both custom keywords and formats that return promises and in this case the schema itself becomes asynchronous. Promise-compatible inline definition is absolutely fine, I actually started with it: `type Promise = Object;`. Or should it be something more elaborate than that?

blakeembrey:
Should be more elaborate, unless you accept pretty much anything? In TypeScript, `Object` is pretty equivalent to `any`. E.g. http://www.typescriptlang.org/play/#src=const%20test%3A%20Object%20%3D%2010

PR #171, opened 2016-04-22
Distribute dist folder to Bower

zetaj:
Currently if using Bower to install this ajv, it doesn't have dist folder which contains ajv.min.js.

This PR will export dist folder which contains ajv.bundle.js, ajv.min.js, ajv.min.js.map, nodent.min.js and regenerator.min.js to Bower. Also, it will clean up unused files, like configuration file that Bower doesn't need.

It's version number should match the version in package.json and should create a tag for it after got merged. Then Bower can get this latest version of ajv. (e.g. 4.0.3)


epoberezkin:
Please see #132

zetaj:
Thanks for pointing me to that discussion. Unfortunately, "**compiled ajv bundle on cdnjs**" will cause some effort to it keep up-to-data. Our Ember project has it's own asset compilation way and it depends on Bower to keep front-end dependencies up-to-date.(Import assets within bower_components) If ajv can provide a similar way like bootstrap does - https://github.com/twbs/bootstrap, it will benefit user a lot.

BTW, I have a short-term solution for our Ember project. If this bower issue can be solved, I am happy to use it.

I will close this PR and looking forward to seeing the solution.

epoberezkin:
Not sure what is your solution, what is possible though I think is to create another repo that will only contain compiled bundles (travis could push all tags there or every time npm version is changed e.g.) and publish to bower from there

zetaj:
It would be great if you can support this, or I am willing to help you on that.

PR #184, opened 2016-05-10
Copy jsdoc comments to typescript definition file

HolgerJeromin:
VisualStudio uses the jsdoc of this file for tooltip.

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -28,8 +28,6 @@ declare namespace ajv {
     * Adds schema to the instance.
     * @param {Object|Array} schema schema or array of schemas. If array is passed, `key` and other parameters will be ignored.
     * @param {String} key Optional schema key. Can be passed to `validate` method instead of schema object or id/ref. One schema per instance can have empty `id` and `key`.
-    * @param {Boolean} _skipValidation true to skip schema validation. Used internally, option validateSchema should be used instead.
-    * @param {Boolean} _meta true if schema is a meta-schema. Used internally, addMetaSchema should be used instead.
     */
     addSchema(schema: Array<Object> | Object, key?: string): void;
     /**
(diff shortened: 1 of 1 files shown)


epoberezkin:
@HolgerJeromin thank you. Looks mostly good, although there are cases when jsdoc is not consistent with typescript (@this Ajv, private parameters).

@blakeembrey can you please have a look? Thank you.

HolgerJeromin:
Good catch. I can adjust TS if it is desirable.
Adding definition of the schema itself (for example [this](https://gist.github.com/enriched/c84a2a99f886654149908091a3183e15) from @enriched) would be a benefit, too.

epoberezkin:
You meant "adjust jsdoc" I think? These are not public parameters.
Not sure about the need to include meta-schema in typescript format...

HolgerJeromin:
I meant ts. But adjusting jsdoc seems to be better, yes. 
People handling json schemas wants to have a meta schema for ts. Tv4 have one included. 
But including is not really needed.

blakeembrey:
FWIW, I have a JSON schema definition in Typings already: https://github.com/typed-typings/lib-json-schema/blob/master/v4/index.d.ts.

The PR looks good, subjective to @epoberezkin for formatting/whitespace/etc.

epoberezkin:
@blakeembrey thank you. The problem with existing JSON-schema definitions is that Ajv also supports v5 proposals. So I would leave it as Object for now.

@HolgerJeromin will you update those jsdocs then?

HolgerJeromin:
`skipValidation` in `addMetaSchema` is not in the README but not marked as internal Parameter in jsdoc. Perhaps this needs adjustment. I removed it. Same with throwOrLogError in validateSchema.

I added `errors?: Array<ErrorObject>;` on the instance, which was missing.

epoberezkin:
Thank you!

PR #196, opened 2016-05-25
Handle overlap between props and patternProps

BryanJacobs:
When a particular attribute matches both an explicit "properties" entry and a
"patternProperties" pattern, we should not fail validation if the attribute's
value is appropriate for the "properties" schema but not for
"patternProperties".

### DIFF
--- a/lib/dot/properties.jst
+++ b/lib/dot/properties.jst
@@ -133,6 +133,7 @@ var valid{{=$it.level}} = true;
 {{?}}
 
 {{ var $useDefaults = it.opts.useDefaults && !it.compositeRule; }}
+var matchedProps = {};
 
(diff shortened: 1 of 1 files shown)


BryanJacobs:
An example of this happening is if you have a schema object with "properties": { "foo": { "type": "integer" } }, and then "patternProperties": {"^.*$": { "type": "string" } }. Data with "foo": 3 will fail validation because it tries to match both the integer and string schemas, where it should stop when the integer constraint is satisfied.

BryanJacobs:
Looked into the failing test in the json schema. Apparently the standard specifies tha both patternProperties and properties must be matched. Never mind!

epoberezkin:
You can use `anyOf` to require that either one is valid, not both.

Issue #197, opened 2016-05-25
only iterate over own properties


mbroadst:
Please excuse me if I'm missing some crucial option, but it seems like the generated code for the validator iterates over enumerable types (`for (var d in data) { ... }`) whereas I would like to restrict that to only own types (the same as `Object.keys`). Is it possible to tweak the generated validator to accommodate this?

epoberezkin:
@mbroadst it is still validating prototype properties against schemas in properties keyword (additionalProperties and patternProperties are resolved with #198, but it doesn't cover properties themselves).

That's the failing test: https://runkit.com/esp/585d958bc5386f00131c3093

epoberezkin:
@mbroadst the option is also not used in "required" keyword validation

epoberezkin:
TODO:
- [x] "properties"
- [x] "required" with "properties"
- [x] "required" for few properties
- [x] "required" for many properties
- [x] "required" with $data
- [x] "dependencies" with property list
- [x] "dependencies" with schema

PR #198, opened 2016-05-26
feat(own-properties): support optionally checking own properties

mbroadst:
Traditionally ajv validates additionalProperties by checking all
enumerable properties of an object. This patch allows the user to
specify that they only want own properties of an object to be
considered in this validation

### DIFF
--- a/README.md
+++ b/README.md
@@ -842,6 +842,7 @@ Defaults:
   format:           'fast',
   formats:          {},
   schemas:          {},
+  ownProperties:    false,
   // referenced schema options:
   missingRefs:      true,
   loadSchema:       undefined, // function(uri, cb) { /* ... */ cb(err, schema); },
@@ -878,7 +879,7 @@ Defaults:
 - _format_: formats validation mode ('fast' by default). Pass 'full' for more correct and slow validation or `false` not to validate formats at all. E.g., 25:00:00 and 2015/14/33 will be invalid time and date in 'full' mode but it will be valid in 'fast' mode.
 - _formats_: an object with custom formats. Keys and values will be passed to `addFormat` method.
 - _schemas_: an array or object of schemas that will be added to the instance. If the order is important, pass array. In this case schemas must have IDs in them. Otherwise the object can be passed - `addSchema(value, key)` will be called for each schema in this object.
-
+- _ownProperties_: indicates that when iterating properties of data to be validated, only properties found directly upon the object are used (rather than all enumerable properties)
 
 ##### Referenced schema options
 
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -104,6 +104,7 @@ declare namespace ajv {
     format?: string;
     formats?: Object;
     schemas?: Array<Object> | Object;
+    ownProperties?: boolean;
     missingRefs?: boolean | string;
     loadSchema?: (uri: string, cb: (err: Error, schema: Object) => any) => any;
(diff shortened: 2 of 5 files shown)


mbroadst:
resolves #197

epoberezkin:
Cool, thank you. There may be a couple of other places where all properties are iterated that should also be affected by this option. + docs update is needed.

mbroadst:
@epoberezkin gotcha, just wanted to see if you'd be interested in the feature, updates incoming

epoberezkin:
The feature definitely makes sense for the cases when there are class instances in your data when class extends prototype without making these methods enumerable. In your code you can control it, but most libraries (including ajv :) don't bother defining instance methods as non-enumerable. And it also makes sense to make it an option as otherwise it would affect performance of the main use case - validating objects parsed from JSON strings where it is not needed.

A more performant approach could be to get the list of own properties in array and then iterating it with "for" loop (one function call to Object.getOwnPropertyNames) instead of checking each property (N function calls)

mbroadst:
@epoberezkin agreed. In fact, I'm trying to retrofit an ODM, replacing its schema validation with ajv/json-schema. It's a common case in ORM/ODMs to have instance objects which represent the data but are also otherwise extended (event emitters, etc), so I was running into all sorts of pain there.

I agree with your performance observations, but I figure I can get the implementation done first and optimize later :)

On another note, I've noticed you can probably speed up your code in a number of places by using the following pattern:

```
for (var i = 0; ii = <something>.length; i < ii; ++i)
```

in most cases you're making a call to check `length` on every iteration it looks like. I can put together a PR for that as well, time permiting

epoberezkin:
On the second issue there is this interesting "essay": http://mrale.ph/blog/2014/12/24/array-length-caching.html

mbroadst:
@epoberezkin ah, very interesting! thanks for the link

mbroadst:
@epoberezkin the only other place I see this being relevant is here: https://github.com/epoberezkin/ajv/blob/master/lib/dot/v5/patternRequired.jst#L14

I modified the code, but I'm not sure how to add it to the v5 tests. I'm sure I could explicitly add it to the options tests - would that suffice?

Also, can you think of anywhere else this might be relevant?

epoberezkin:
Also:
https://github.com/epoberezkin/ajv/blob/master/lib/dot/properties.jst#L214
and
https://github.com/epoberezkin/ajv/blob/master/lib/dot/properties.jst#L253

unless I am missing something.

I was thinking about some other cases, but it's schema that iterated there, that is highly unlikely to be some class instance, so no need to change anything there...

Yes, in options tests is perfect.

Thank you very much

mbroadst:
@epoberezkin okay that should be it. It looks like vscode autocorrected (or at least modified) the smiley face UTF16 character there, but I'm not quite sure how to revert it tbh

mbroadst:
@epoberezkin any idea what sort of timeline there might be for a patch/minor release including this feature? Definitely do not intend to be pushy, but I'm heavily dependent on it for this other work I'm doing and the nature of the build process leaves me manually linking my local builds to my `node_modules` folder - not an ideal process!

epoberezkin:
No worries, today or tomorrow, sorry for the delay :)

epoberezkin:
in [4.1.0](https://github.com/epoberezkin/ajv/releases/tag/4.1.0)

mbroadst:
rad, much appreciated!

PR #219, opened 2016-06-26
Add `schema` to interface returned by `getSchema`

dancasey:
`getSchema` returns "schema validating function (with property `schema`).", but property `schema` is not included in interface `ValidateFunction`.

This fix seems to work for me, since VS Code no longer complains when I do something like `let mySchema = ajv.getSchema(id).schema;`, but there may be a better way to annotate this.


epoberezkin:
Thank you, seems ok. And it should be optional indeed as the same definition is used for custom keyword functions (although maybe it's worth using different type definitions?).

PR #239, opened 2016-07-19
fix errors in CUSTOM doc

jonathanGB:
Title says it all... errors I've spotted while reading the CUSTOM doc


epoberezkin:
Thank you!

Issue #158, opened 2016-03-31
Allow coerce to array of {integer,number,string,boolean,null}


willfarrell:
I have a use case when I'm validating query parameters. I have `/api?filter=A` => `{ filter:'A' }` and `/api?filter=A&filter=B` =>  `{ filter:['A','B'] }` being passed in. Currently I'm iterating through the object and schema to find this case and coerce before validation, but it would be faster and cleaner to have it built-in.

Can we extend the coerce to allow casting to an array when it's items type is non-array/object?

**example**
partial schema:

``` javascript
{ "type":"object", "properties":{
"filter":{
   "type":"array",
  "items":{ "type":"integer" } // possible types: integer,number,string,boolean,null
}
}}
```

input

``` json
{
  "filter":"42"
}
```

expected output:

``` json
{
  "filter":[42]
}
```

epoberezkin:
It may be an interesting idea, please let me think about it.
The immediate workaround is to either use anyOf/oneOf keywords or to define some custom keyword.

willfarrell:
Not a problem, I have a working work around in place for now. Just looking for a more performant solution.
I was using anyOf/oneOf keywords before moving to the solution I have in place now. I found this approach to be cleaner.

epoberezkin:
My concerns are:
1. unless you expect this coercion it can be surprising and unexpected.
2. it may be coerced twice (first to array, then potentially to the type of array item, although maybe it's fine).
3. if it coerces to array it should also be able to coerce array of a single item to a scalar type, but it means that it would coerce twice in one go (first to a scalar than to the expected type)...

So if all these things seem ok, then at least it should be a different option value (`coerceTypes: 'array'`) that would coerce to/from arrays in addition to scalar coercion...

willfarrell:
All three points are bang on for what I was thinking. `coerceTypes: 'array'` would be perfect.

epoberezkin:
in [4.3.0](https://github.com/epoberezkin/ajv/releases/tag/4.3.0)

PR #246, opened 2016-07-23
Add in ability to coerce an array

willfarrell:
As per discussion at #158

### DIFF
--- a/COERCION.md
+++ b/COERCION.md
@@ -1,6 +1,6 @@
 # Ajv type coercion rules
 
-To enable type coercion pass option `coerceTypes` to Ajv (it is `false` by default). See [example](https://github.com/epoberezkin/ajv#coercing-data-types).
+To enable type coercion pass option `coerceTypes` to Ajv with `true` or `array` (it is `false` by default). See [example](https://github.com/epoberezkin/ajv#coercing-data-types).
 
 The coercion rules are different from JavaScript:
 - to validate user input as expected
@@ -10,6 +10,8 @@ The coercion rules are different from JavaScript:
 Type coercion only happens if there is `type` keyword and if without coercion the validation would have failed. If coercion to the required type succeeds then the validation continues to other keywords, otherwise the validation fails.
 
 If there are multiple types allowed in `type` keyword the coercion will only happen if none of the types match the data and some of the scalar types are present (coercion to/from `object`/`array` is not possible). In this case the validating function will try coercing the data to each type in order until some of them succeeds.
+
+If `coerceTypes` is set to `array`
 
 Possible type coercions:
 
@@ -98,3 +100,9 @@ Unlike JavaScript, only these numbers can be coerced to `boolean`:
 #### To boolean type
 
 `null` coerces to `false`
+
+## Coersion from array
+
+Wraps and non-array input in an array.
+
+- `"foo"` -> `[ "foo" ]`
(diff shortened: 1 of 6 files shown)


epoberezkin:
This is a partial solution. It coerces to array but doesn't coerce from array. Coercions should be reversible, otherwise it would fail with coercions in cases it would pass without (when anyOf/etc are used).

All rules from #158 need to be implemented before it can be merged...

| schema | data | result | pass |
| :-- | :-: | :-: | :-: |
| {type:'number'} | [5] | 5 | no |
| {type:'string'} | [5] | '5' | no |
| {type:'array',items':{type:'number'}} | 5 | [5] | yes |
| {type:'array',items':{type:'string'}} | 5 | ['5'] | yes |

epoberezkin:
Also, please avoid unnecessary white-space changes. It's spaces, not tabs, 2 space indentations everywhere.

epoberezkin:
Also, there should be tests of standard coercions with coerceTypes:'array' option to make sure they are not broken, and tests of array coercions with coerceTypes:true to make sure it doesn't coerce to/from array.

willfarrell:
Sorry got distracted by my own use case. Array to other types added, more tests added, and I hope I caught all the white space changes too. Thanks for reviewing the PR.

epoberezkin:
Thank you, I will review/merge this week.

epoberezkin:
Getting there... :) Could you please also remove unnecessary changes of `==` to `===`? In my "book", I always use `==`/`!=` unless `===`/`!==` must be used, for several reasons :)

epoberezkin:
By the way, this object <-> array coercions seems a bit over the top - I'd limit to scalars...

epoberezkin:
Also it seems to be only one way at the moment, so removing it is probably better.

epoberezkin:
Thank you. Will publish today evening.

epoberezkin:
in 4.3.0

jfstgermain:
It's unfortunate that non scalar coercion was dropped.  It's very useful when having to convert XML payloads to JSON.  In XML, Arrays of one object are converted to a simple object since there's no way of telling the intent of the markup eg: 
`<users>
      <name>Joe</name>
   </users>`

In our swagger file, `users` is an array but this gets transformed into:
`{users:{name:Joe}}` instead of `{users:[{name:Joe}]}`

From what I see it doesn't seem to be the case, but is there a way to add custom rules or override them?  

Thanks

epoberezkin:
Ajv supports custom keywords, any data transformation can be defined using them. If you need to run transformation before validation you can do it with allOf keyword (in the same schema custom keywords are executed after).

willfarrell:
Array coercion was dropped? If I write another PR, can we get it back? As @jfstgermain mentioned super useful. I use it for API query params.
FYI The docs still mention support for it: `{ coerceTypes: 'array' }`.

epoberezkin:
@willfarrell it was not dropped. I think @jfstgermain refers to the ability to define custom coercions.

jfstgermain:
I meant Array coercion for non scalar types...

epoberezkin:
I see. It's quite easy to define as a custom keyword.

jfstgermain:
Looking into how to achieve this.  I'm really impressed by the extensibility of the lib!  Great job @epoberezkin  :)

willfarrell:
Take a look at the PR itself, maybe a PR for extending it may be in order.

PR #251, opened 2016-07-27
[typescript] Use Class in type's definition to allow use 'new' operator

dalcib:
After struggle a lot with the message `[ts] Only a void function can be called with the 'new' keyword`, every time that I try to use the code above, I got to resolve the issue  by using export class insted a namespace.

```
import Ajv from 'ajv';
...
const ajv = new Ajv(ajvOptions);
```

This change allows to type cheking the options, outside of the parameter like this:

```
import Ajv, {Options} from 'ajv';
const ajvOptions: Options = {...}
const ajv = new Ajv(ajvOptions);
```

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -1,7 +1,9 @@
-declare function ajv (options?: ajv.Options): ajv.Ajv;
-
-declare namespace ajv {
-  interface Ajv {
+  export default class Ajv {
+    /**
+     * Constructor
+     * @param  {Object} Options
+     */
+    constructor(options?)
     /**
     * Validate data using schema
     * Schema will be compiled and cached (using serialized JSON as key. [json-stable-stringify](https://github.com/substack/json-stable-stringify) is used to serialize.
@@ -77,7 +79,7 @@ declare namespace ajv {
     * @return {String} human readable string with all errors descriptions
     */
     errorsText(errors?: Array<ErrorObject>, options?: ErrorsTextOptions): string;
-    errors?: Array<ErrorObject>;
+    errors: Array<ErrorObject>;
   }
 
   interface Thenable <R> {
@@ -95,7 +97,7 @@ declare namespace ajv {
     schema?: Object;
   }
 
-  interface Options {
+  export interface Options {
     v5?: boolean;
     allErrors?: boolean;
     verbose?: boolean;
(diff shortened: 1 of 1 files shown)


epoberezkin:
Looks ok to me, @blakeembrey I'd appreciate if you could look as well...

blakeembrey:
This shouldn't be merged. It's not correct for TypeScript and work in the runtime. You can read the difference between ES6 `export default` and `export =` here: https://www.typescriptlang.org/docs/handbook/modules.html.

lib/ajv.d.ts
```diff
@@ -1,7 +1,9 @@
-declare function ajv (options?: ajv.Options): ajv.Ajv;
-
-declare namespace ajv {
-  interface Ajv {
+  export default class Ajv {
+    /**
+     * Constructor
+     * @param  {Object} Options
+     */
+    constructor(options?)
```
blakeembrey, line 6: This breaks with `noImplicitAny`.

lib/ajv.d.ts
```diff
@@ -1,7 +1,9 @@
-declare function ajv (options?: ajv.Options): ajv.Ajv;
-
-declare namespace ajv {
-  interface Ajv {
+  export default class Ajv {
```
blakeembrey, line 1: This breaks because this library is not an `export default`. You can read about what `export default` and `export =` means https://www.typescriptlang.org/docs/handbook/modules.html

blakeembrey:
Also, changing it to a class instead of a function would completely break backward compatibility for people who do use it as a function currently. You should instead enable support for both ways.

blakeembrey:
You should change it to this instead:

```
declare var ajv: { 
  (options?: ajv.Options): ajv.Ajv;
  new (options?: ajv.Options): ajv.Ajv;
}
```

epoberezkin:
@blakeembrey thank you, I was thinking about it not being backwards compatible...

dalcib:
Ok. I understand. I going to make another PR fixing this.

blakeembrey:
@dalcib Just FYI, but you would usually update the PR instead. It helps maintainers keep track of the discussion and changes, and generally avoids unneeded churn. To update the PR, you can just keep working on the branch and the maintainers and anyone else subscribed to the PR would see updates.

Issue #261, opened 2016-08-04
date-time full format validation error on date-only strings


kroky:
Hello and thanks for the great lib!

I found an issue validating a date-time string format against date-only strings like '2016-08-04'. The error is:

`TypeError: Cannot read property 'match' of undefined in lib/compile/formats.js:93`

It seems the date_time function splits by the time separator and passes the second part to time function which calls match on the string but string is undefined in the above use-case. Will submit a PR with a possible fix...

epoberezkin:
Good catch, thank you. It should still fail validation though, there is "date" format for date-only strings.

PR #262, opened 2016-08-04
fix: date-time string validation to fail validating for date-only string

kroky:
This should resolve issue #261. Added specs for the failing use-case specifically. Let me know if you see any issues.

### DIFF
--- a/lib/compile/formats.js
+++ b/lib/compile/formats.js
@@ -90,6 +90,8 @@ function date(str) {
 
 
 function time(str, full) {
+  if (str === undefined) return false;
+
   var matches = str.match(TIME);
   if (!matches) return false;
 
--- a/lib/compile/formats.js
+++ b/lib/compile/formats.js
@@ -90,8 +90,6 @@ function date(str) {
 
 
 function time(str, full) {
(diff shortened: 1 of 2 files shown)


lib/compile/formats.js
```diff
@@ -90,6 +90,8 @@ function date(str) {
 
 
 function time(str, full) {
+  if (str === undefined) return false;
```
epoberezkin: The guards should be on the calling side, in date_time function. This function is also used for "time" format and it will never be called for undefined, so this condition is only necessary if called from date_time

spec/tests/rules/format.json
```diff
@@ -101,6 +101,22 @@
     ]
   },
   {
+    "description": "validation of date-time strings",
```
epoberezkin, line 104: Strange there were no tests for this format at all. There is a mistake in date format tests above when in description it's called "date-time" - maybe that's why... Thank you

kroky:
Moved the guard to date_time function. If you think these commits gets messy, feel free to make the changes yourself, I don't mind.

epoberezkin:
Thank you

PR #263, opened 2016-08-05
[typescript] Allow use 'new' operator

dalcib:


dalcib:
@blakeembrey, with `export =`I can export only one thing. How could I to export another thing like 'Options'?

blakeembrey:
@dalcib You would use a namespace, it's already used here.

blakeembrey:
LGTM 👍

Issue #274, opened 2016-08-14
Add assert method


felixfbecker:
Should have the same signature as `validate()`, but instead of returning a boolean, throw an error when the validation failed. This could be used both in APIs (with a custom error handler) and in API tests and would reduce boilerplate code. The error class should be accessible for `instanceof` checks and should have the `errors` property to get details about the failure. The `message` should be set to the return of `errorsText()`.

epoberezkin:
You can extend it like this:

``` javascript
Ajv.prototype.assert = function(schema, data) {
  if (!this.validate(schema, data)) throw new Ajv.ValidationError(this.errors);
};
```

I will think whether it should be a part of the library

epoberezkin:
For tests there is [chai-ajv-json-schema](https://www.npmjs.com/package/chai-ajv-json-schema)

broofa:
Worth revisiting?  Every time I approach this library, it is with the intention of using it in much the same manner as Node's built in `assert()` method.  It would be nice to be able to `compile` a validator that throws rather than returning.  This would be useful, for example, in creating custom sequelize validator functions ...
```
sequelize.define('mymodel', {
  someColumn: {
    ...
    validate: {
      customValidator: ajv.compileAssert({... schema definition ...})
    }
  }
});
```

Also, the suggested workaround (extending `AJV.prototype`) kind of breaks down if you're requiring AJV from multiple files.  You end up with potential race conditions where the file extending AJV has to be required before all other files.

epoberezkin:
@broofa https://github.com/epoberezkin/ajv/blob/master/FAQ.md#can-we-change--extend-api-to-add-a-method-that-would-return-errors-rather-than-assign-them-to-errors-property

Given that it’s convenient to have a wrapper for the library in the application, this would be the place to introduce this API.

PR #275, opened 2016-08-14
Assert

felixfbecker:
#274

I added an `assert` method that works like `validate` but throws a `ValidationError` if the validation failed. The error also has `actual` and `errorsText` set. To make it easier to use `errorsText`, I added support for omitting `errors` argument but providing `options`. I also added the `ValidationError` because it was missing.

I wasn't sure how to handle async validation. A method should better not return sometimes a promise, sometimes throw an error. If we want to cover async validation, it would be better to add another method `assertAsync()`. But I think it's fine as-is because people expect an `assert` function to throw sync.

### DIFF
--- a/README.md
+++ b/README.md
@@ -747,6 +747,15 @@ __Please note__: every time this method is called the errors are overwritten so 
 If the schema is asynchronous (has `$async` keyword on the top level) this method returns a Promise. See [Asynchronous validation](#asynchronous-validation).
 
 
+##### .assert(Object schema|String key|String ref, data)
+
+Just like `validate`, but throws a [`ValidationError`](#validation-errors) if the validation fails.
+
+The error object has the `errors` property and the `errorsText` method.
+
+This method does not support async validation.
+
+
 ##### .addSchema(Array&lt;Object&gt;|Object schema [, String key])
 
 Add schema(s) to validator instance. From version 1.0.0 this method does not compile schemas (but it still validates them). Because of that change, dependencies can be added in any order and circular dependencies are supported. It also prevents unnecessary compilation of schemas that are containers for other schemas but not used as a whole.
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -1,6 +1,7 @@
-declare var ajv: { 
+declare var ajv: {
   (options?: ajv.Options): ajv.Ajv;
   new (options?: ajv.Options): ajv.Ajv;
+  ValidationError: ajv.ValidationErrorConstructor;
 }
 
 declare namespace ajv {
@@ -257,6 +258,25 @@ declare namespace ajv {
   }
 
   interface NoParams {}
+
+  interface ValidationErrorConstructor {
+    prototype: ValidationError;
+    new (errors: ErrorObject[]): ValidationError;
+  }
+
+  interface ValidationError extends Error {
+
+    /** "validation failed" */
+    message: string;
+
+    /** true */
+    avj: boolean;
+
+    /** true */
+    validation: boolean;
+
+    errors: ErrorObject[];
+  }
 }
 
 export = ajv;
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -79,7 +79,8 @@ declare namespace ajv {
     * @param  {Object} options optional options with properties `separator` and `dataVar`.
     * @return {String} human readable string with all errors descriptions
     */
-    errorsText(errors?: Array<ErrorObject>, options?: ErrorsTextOptions): string;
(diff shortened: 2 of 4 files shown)


epoberezkin:
@felixfbecker thanks for the PR. I will think about it, no promises though. See [FAQ: can we change/extend API](https://github.com/epoberezkin/ajv/blob/master/FAQ.md#can-we-change--extend-api-to-add-a-method-that-would-return-errors-rather-than-assign-them-to-errors-property).

My objection is that assert method is not substantially different from validate - they both do validation, just return results in a different way, which is a matter of convenience, rather than additional functionality. I think that convenience API wrappers belong in the application code, not in the library.

To have this method in the application you need 1 line function (https://github.com/epoberezkin/ajv/issues/274#issuecomment-239668695). In the library it requires 100+ lines - test, docs, etc.

felixfbecker:
I see your point. But I think an `assert` method is something very basic every validation library should have. Joi's validation API is very slim to, but it has `assert`, see https://github.com/hapijs/joi/blob/v9.0.4/API.md#assertvalue-schema-message. There definitely is a great use case, it's more than a nice-to-have but rather an essential interface for validation libraries.

You also have to admit that assigning on `errors` can be very error-prone.

> To have this method in the application you need 1 line function (#274 (comment)). In the library it requires 100+ lines - test, docs, etc.

Well first patching the prototype is a bad idea if a library may ever decide to implement the same function and use it internally. So a wrapper it would be. But I see it the other way around: To have this method, thousands of users have to all write their own wrapper, and write their own unit tests for it, update it etc. If it was in the library, a user can rely on it that it is well-tested.

Anyway, sleep a night over it, let me know how you decided. This library is great 🙂

epoberezkin:
Will think indeed... 

I think Joi is one of the very few that has `assert`, if not the only one. I've never seen it before in other validators to be honest.

gajus:
@felixfbecker

> But I think an assert method is something very basic every validation library should have.

But none of the competing libraries do have it:
- https://github.com/geraintluff/tv4
- https://github.com/mafintosh/is-my-json-valid
- https://github.com/bugventure/jsen
- https://github.com/AlexeyGrishin/schemasaurus
- https://github.com/zaggino/z-schema
- https://github.com/pandastrike/jsck
- https://github.com/tdegrunt/jsonschema

API familiarity is a valid argument. However, please provide more than one example when making a case.

> To have this method, thousands of users have to all write their own wrapper, and write their own unit tests for it, update it etc. If it was in the library, a user can rely on it that it is well-tested.

Just release an NPM package `ajv-assert` that provides a wrapper for the check.

epoberezkin:
@blakeembrey what do you think?

blakeembrey:
No strong arguments either way. Personally, I've built `assert` into my own validation library but it's used for type assertion (not validation - E.g. does the shape match, not the validations match).

It does look like the code is relying on the `typeof` check for the return to know whether it's async (I'm unfamiliar with the async API here to be honest, is that a promise return?). Is there a better way to do that? I realise most people should fix that immediately, but having side effects that might still be running while the function has thrown is iffy.

epoberezkin:
@blakeembrey thank you. async is a concern indeed for this api (although this particular issue could be solved - it is known in advance whether the schema is async).

@gajus I actually like the idea of a separate module - different assertions could have been implemented there over time, rather than just one. You could assert for failure, for a presence/absence of a particular error (or particular error only, e.g. in order to pass the data, but generate a warning in case of additional properties validation errors), for being valid against multiple schemas, etc.

@felixfbecker what do you think? There are several ajv features in separate packages already and they are all mentioned in docs - [ajv-i18n](https://github.com/epoberezkin/ajv-i18n), [ajv-cli](https://github.com/jessedc/ajv-cli) (@jessedc started it), [ajv-merge-patch](https://github.com/epoberezkin/ajv-merge-patch), [ajv-keywords](https://github.com/epoberezkin/ajv-keywords). Given that these features are only useful for some users, it seemed wrong to add them to the main ajv package. I am actually considering splitting both v5 proposals (more likely) and async schemas support (less likely) to separate packages in the next (or the following) major version to reduce the size of the ajv core and to have a smaller browser bundle for the majority of users.

If you like to do it this way and the quality of ajv-assert is on the same level (docs, tests, visible test coverage) I would be happy to refer to it in ajv docs in similar way my packages are mentioned.

felixfbecker:
I'm convinced that having this in core is not desirable, and async schemas are indeed a problem. I don't think I will publish (and maintain, test etc) a seperate NPM package for this, since I only need the very basic three-line assert method mentioned here. Thank you for your time.

Issue #285, opened 2016-08-21
Add support for absolute JSON Pointers in the $data reference


HotelDon:
**What version of Ajv you are you using?**
4.5.0
**What problem do you want to solve?**
The $data reference only supports relative JSON pointers, which limits code reuse for data that is always in one location but $data referenced in multiple places. It also creates problems for more complex schemas; for example, a recursive schema can't reference data in the root of the document because it's relative distance to the root changes depending on how many levels deep you've gone. 
**What do you think is the correct solution to problem?**
Add in support for regular (absolute) JSON pointers to the $data reference. I have an issue open on the JSON-Schema Github for adding this to the v5 proposal (https://github.com/json-schema-org/json-schema-spec/issues/40) that explains the issue in a bit more depth. The short version is, regular JSON pointers start with a / instead of a number, and the pointer is always anchored to the root of the document, which solves the problem. Additionally, since relative pointers ALWAYS start with a number, there's no ambiguity as to which type of pointer you're trying to use. 
**Will you be able to implement it?**
I couldn't find where the code for handling the relative pointers is located, but point me in the right direction and I might be able to. However, the $data reference looks like some of the more complicated parts of the library, so I can't say for sure.

epoberezkin:
Yep, good idea.

You can just edit the wiki  re $data proposal btw.

It's here btw: https://github.com/epoberezkin/ajv/blob/master/lib/compile/util.js#L218

HotelDon:
The proposals are all on the old repo and haven't been moved over, so I was a bit confused if anyone would have even noticed if I had changed it. Didn't want to ruffle too many feathers. 

I'll try to have a pull request/admission of failure ready by tomorrow.

epoberezkin:
In 4.6.0

PR #289, opened 2016-08-25
Add JSON pointer support to $data reference.

HotelDon:
Issue #285
I added support for regular JSON pointers to the $data reference
Make sure I didn't accidentally create the regex from hell when I used and modified the existing relative-json-pointer regex.

### DIFF
--- a/lib/compile/util.js
+++ b/lib/compile/util.js
@@ -214,20 +214,33 @@ function getPath(currentPath, prop, jsonPointers) {
 }
 
 
+var JSON_POINTER = /^#|\/(?:[^~]|~0|~1)*?$/;
 var RELATIVE_JSON_POINTER = /^([0-9]+)(#|\/(?:[^~]|~0|~1)*)?$/;
 function getData($data, lvl, paths) {
-  var matches = $data.match(RELATIVE_JSON_POINTER);
-  if (!matches) throw new Error('Invalid relative JSON-pointer: ' + $data);
-  var up = +matches[1];
-  var jsonPointer = matches[2];
-  if (jsonPointer == '#') {
-    if (up >= lvl) throw new Error('Cannot access property/index ' + up + ' levels up, current level is ' + lvl);
-    return paths[lvl - up];
+  var up;
+  var jsonPointer;
+  var data;
+  var matches;
+  if ($data.charAt(0) == '/') {
+    matches = $data.match(JSON_POINTER);
+    if (!matches) throw new Error('Invalid JSON-pointer: ' + $data);
+    jsonPointer = matches[0];
+    data = 'data';
   }
+  else {
+    matches = $data.match(RELATIVE_JSON_POINTER);
+    if (!matches) throw new Error('Invalid JSON-pointer: ' + $data);
+    up = +matches[1];
+    jsonPointer = matches[2];
+    if (jsonPointer == '#') {
+      if (up >= lvl) throw new Error('Cannot access property/index ' + up + ' levels up, current level is ' + lvl);
+      return paths[lvl - up];
+    }
 
-  if (up > lvl) throw new Error('Cannot access data ' + up + ' levels up, current level is ' + lvl);
-  var data = 'data' + ((lvl - up) || '');
-  if (!jsonPointer) return data;
+    if (up > lvl) throw new Error('Cannot access data ' + up + ' levels up, current level is ' + lvl);
+    data = 'data' + ((lvl - up) || '');
+    if (!jsonPointer) return data;
+  }
 
   var expr = data;
   var segments = jsonPointer.split('/');
--- a/lib/refs/json-schema-v5.json
+++ b/lib/refs/json-schema-v5.json
@@ -30,7 +30,10 @@
             "properties": {
                 "$data": {
                     "type": "string",
-                    "format": "relative-json-pointer"
+                    "anyOf": [
+                        { "format": "relative-json-pointer" }, 
+                        { "format": "json-pointer" }
+                    ]
                 }
             },
             "additionalProperties": false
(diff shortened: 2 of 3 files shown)


HotelDon:
Not sure where the test code for $data is located.

lib/refs/json-schema-v5.json
```diff
@@ -30,7 +30,9 @@
             "properties": {
                 "$data": {
                     "type": "string",
-                    "format": "relative-json-pointer"
+                    "format": {
+                        "anyOf": ["relative-json-pointer", "json-pointer"]
+                    }
                 }
```
epoberezkin, line 37: This is not valid - should be `anyOf` with two `format` keywords inside.

epoberezkin:
Thank you for the PR!
Could you please:
- determine whether it's a relative or an absolute pointer by checking the first character `$data[0] == '/'`,  matching regexp twice is wasteful.
- fix meta schema (see comment in code)
- squash existing commits (with the change above) in your branch
- add tests [here](https://github.com/epoberezkin/ajv/tree/master/spec/v5/%24data) (as a separate commit). I suggest creating a separate file `absolute_ref.json` to test several different keywords (constant and a couple of others). There should also be tests of how it works inside $ref (to make sure it can go to a higher level than the root of $ref, because relative refs can't do it on purpose (I remember so at least :), but absolute ones should be able to come to a higher level. If at the moment they can't, please skip those tests, I will think how to improve it (you can use `skip: true` in json test file, both on a suite level and a test level).

HotelDon:
I can say definitively from my own manual testing, absolute pointers cannot go higher than the root of a $ref (`$data: "/"` will always resolve to the root of the $ref, not the entire document), which was an issue I was going to bring up once I got this commit fixed up. It looks like this limitation is handled on the $ref side of the code, not the $data side (or at least, not in any of the code I've touched so far). At least I know this was a side effect of something done intentionally, I was going a little crazy wondering some of my test schemas weren't working the way they should have. 

As for everything else, I'll get those fixes and tests written up in a day or two. Thanks!

HotelDon:
Alright, I finished putting together the tests. The last two tests in the file I added are set to `skip: true`, as they are testing the interaction between $ref and $data, which currently fails. The rest of the tests were copied over from other files and modified to use regular JSON pointers, and removing the tests that don't apply (the # sign has a completely different use in non-relative JSON pointers). If there's anything else I need to fix, let me know!

epoberezkin:
Thank you! I will try to fix this issue - I hoped that even if it doesn't go outside of $ref at least the value would be correct based on the absolute reference. From what you say it's even worse - it just points to the wrong value (in case the root for the ref is different from the data root).

Docs need to be updated too, I will do it.

epoberezkin:
@HotelDon I understand how to resolve the issue - the validation function should be passed an additional 5th argument - `rootData`, then the code generated by getData function can use this variable instead of `data` (that is the top level data at the current $ref'd schema).

I don't think it will be a breaking change for anyone - It would only be breaking if somebody were passing additional arg(s) to compiled validating function (they could only be used in inline custom keywords).

lib/compile/util.js
```diff
@@ -214,20 +214,33 @@ function getPath(currentPath, prop, jsonPointers) {
 }
 
 
+var JSON_POINTER = /^#|\/(?:[^~]|~0|~1)*?$/;
```
epoberezkin, line 217: Why do you need `"#"` in the pattern? If you mean it as the pointer to the root it should be `""` (an empty string) instead. `"/"` is the pointer to the empty property at the top level. `"#"` in relative json pointer is a special thing resolving to the name of the property/index rather than to data. Or am I missing something?

epoberezkin:
In 4.6.0

lib/compile/util.js
HotelDon: Oops. I forgot to take it out. The # sign at the start of a pointer DOES have a specific meaning for regular JSON pointers, but it's one that has no relevance to JSON schema, and I should have taken it out of the regex.

epoberezkin: Not really: https://tools.ietf.org/html/rfc6901#section-3
It's when JSON pointer is in URI (in $ref)

HotelDon: Oh jeeze, I had never actually noticed that $refs are also JSON pointers. Now that you've pointed that out, I have noticed something. A $data reference formatted like a URI (`"$data": "#/foo"`) will be considered valid in the schema, but the data won't validate. Formatting $ref as a non-URI pointer has a similar effct: `"$ref": "/definitions/foo"` won't throw an error in the schema, but the data won't validate. 

According to the JSON-schema spec, a $ref has to be formatted as a URI, and the $data proposal doesn't mention URI support at all (while I can think of some uses for it, it sounds like a potential security nightmare).

epoberezkin: `/defenitions/foo` is a valid relative URI (definitions is a folder and foo is a file), schema compilation fails because it cannot find schema with such uri.

But `#/foo` is not a valid JSON pointer, it passes json-pointer format validation because I allowed (maybe mistakenly? Not sure, it's not precisely specified) URI encoded JSON-pointers in this format.

I don't think it's such a big deal to worry about, as long as you are aware that $data only takes normal JSON pointers it's fine. Allowing uri encoded JSON pointers would be a bit more effort that just stripping #, you'd need to decode as well.

HotelDon: Cool, I understand. 

Thanks for all your help, by the way. This pull request solved a lot of headaches I was having. If you're ever near San Jose, I'll buy you a beer.

epoberezkin: Ha :) Thanks!

Issue #156, opened 2016-03-31
Serializable validation function


ReactiveRaven:
Is there a way to get a serializable validation function that can be used in another context?

For example, CouchDB is a document-based database with a 'validate_doc_update' function which is serialised as a string on a special class of JSON documents called Design Documents. It would be very useful to be able to store the validate function to a string so it could be used to validate documents as they're being updated in the document-database, but currently there are dependencies outside the validation function's context (eg the `equals` function being missing is the current error message I'm getting).

Short of serialising the entire `bundle.js` file in from npm, eval-ing it and generating a validation function inside the inner context, is there a way to generate a serializable validation function?

epoberezkin:
Not at the moment, validation function is a closure as you pointed out.

I was planning to implement it in [ajv-cli](https://github.com/jessedc/ajv-cli/issues/1) (and it would require splitting some run-time dependency from ajv).

Maybe the better way to do it is to have the api for creating a serialisable validation function exposed in ajv and have ajv-cli simply use this api.

epoberezkin:
@ReactiveRaven I was looking at [couchdb docs](http://guide.couchdb.org/draft/validation.html), it seems like this function should have everything inside, so it won't be able to use any dependencies or globals, correct? Are there any other limitations?

gabrielmancini:
ajv still dont work on couchdb

epoberezkin:
@gabrielmancini this is a very old issue - since then you can create serialisable self-contained validation functions with ajv-pack (with some limitations).

mugwhump:
@gabrielmancini I can confirm that it does, in case anyone else is wondering. The output generated by `ajv-cli compile` can be plunked into a design doc as a commonJS lib and used by other design functions. :)

PR #293, opened 2016-08-31
Add option to not retain validator source code

rf:
Programs with many schemas eat a lot of memory in the form of validator source strings.

Added an option to retain the `sourceCode` property and changed the `eval` to a `new Function` which is indirect and therefore does not retain references to the current context. This required explicitly passing in all variables that the compiled functions depend on. The option is called `retainSourceCode` and it defaults to true.

This has the added benefit of explicitly showing what variables are being used by the validator functions.

@MadanThangavelu @Matt-Esch

### DIFF
--- a/lib/compile/index.js
+++ b/lib/compile/index.js
@@ -32,7 +32,8 @@ function compile(schema, root, localRefs, baseId) {
     , patternsHash = {}
     , defaults = []
     , defaultsHash = {}
-    , customRules = [];
+    , customRules = []
+    , retainSourceCode = this._opts.retainSourceCode !== false;
 
   root = root || { schema: schema, refVal: refVal, refs: refs };
 
@@ -54,7 +55,7 @@ function compile(schema, root, localRefs, baseId) {
       cv.refVal = v.refVal;
       cv.root = v.root;
       cv.$async = v.$async;
-      cv.sourceCode = v.sourceCode;
+      if (retainSourceCode) cv.sourceCode = v.sourceCode;
     }
     return v;
   } finally {
@@ -100,7 +101,7 @@ function compile(schema, root, localRefs, baseId) {
 
     sourceCode = vars(refVal, refValCode) + vars(patterns, patternCode)
                    + vars(defaults, defaultCode) + vars(customRules, customRuleCode)
-                   + sourceCode;
+                   + sourceCode + 'return validate;';
 
     if (opts.beautify) {
       /* istanbul ignore else */
@@ -114,7 +115,37 @@ function compile(schema, root, localRefs, baseId) {
       validateCode = $async && transpile
                       ? transpile(sourceCode)
                       : sourceCode;
-      eval(validateCode);
+
+      var makeValidate = new Function(
+        'refVal',
+        'co',
+        'equal',
+        'ucs2length',
+        'root',
+        'formats',
+        'customRules',
+        'ValidationError',
+        'RULES',
+        'defaults',
+        'self',
+        'validate',
+        validateCode
+      );
+
+      validate = makeValidate(
+        refVal,
+        co,
+        equal,
+        ucs2length,
+        root,
+        formats,
+        customRules,
+        ValidationError,
+        RULES,
+        defaults,
+        self
+      );
+
       refVal[0] = validate;
     } catch(e) {
       console.error('Error compiling schema, function code:', validateCode);
@@ -127,7 +158,7 @@ function compile(schema, root, localRefs, baseId) {
     validate.refVal = refVal;
     validate.root = isRoot ? validate : _root;
     if ($async) validate.$async = true;
-    validate.sourceCode = sourceCode;
+    if (retainSourceCode) validate.sourceCode = sourceCode;
 
     return validate;
   }
--- a/lib/compile/index.js
+++ b/lib/compile/index.js
@@ -9,6 +9,17 @@ var beautify = (function() { try { return require('' + 'js-beautify').js_beautif
 
 var validateGenerator = require('../dotjs/validate');
 
+/**
+ * Functions below are used inside compiled validations function
+ */
+
+var co = require('co');
+var ucs2length = util.ucs2length;
+var equal = require('./equal');
+
+// this error is thrown by async schemas to return validation errors via exception
+var ValidationError = require('./validation_error');
+
 module.exports = compile;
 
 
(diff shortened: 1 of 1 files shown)


epoberezkin:
Thanks for the PR!
This essentially contains two [independent] changes:
1. optionally not retaining the source code (I'd like to see that it reduces the memory usage - you can get the source code of the function only by calling `.toString()` method, or this property adds additional overhead? If that's the case this option can become the default in 5.0.0).
2. different way of creating the closure (rather than scoped to the context, it is scoped to the wrapping function)

Not sure I understand why the 2nd is needed... All the shared objects that the closures are using will be shared anyway, all the closure-specific objects will be different anyway...

The inventory of these variables is useful in any case, as I keep thinking about some "easy" way to solve #156

The build fails because of ESLint

lib/compile/index.js
```diff
@@ -100,7 +101,7 @@ function compile(schema, root, localRefs, baseId) {
 
     sourceCode = vars(refVal, refValCode) + vars(patterns, patternCode)
                    + vars(defaults, defaultCode) + vars(customRules, customRuleCode)
-                   + sourceCode;
+                   + sourceCode + 'return validate;';
 
     if (opts.beautify) {
       /* istanbul ignore else */
```
epoberezkin, line 118: Would this still affect `.toString()`code?

epoberezkin:
Also I think in case the second part is needed (new Function instead of eval) [this line](https://github.com/epoberezkin/ajv/blob/master/lib/dot/validate.jst#L37) should be `var validate = ...`.

rf:
With a direct call to `eval`, the `validate` function you get back retains a reference to the `sourceCode` in its closure. That looks like this:

screen shot 2016-08-31 at 4 44 22 pm

The blue edges are "context" references, which is what happens when a variable is closed over. Once you remove the explicit property, it looks like this:

screen shot 2016-08-31 at 4 46 40 pm

Only blue context edges left. Once you remove those, by using an indirect eval:

screen shot 2016-08-31 at 4 47 15 pm

The string is no longer retained.

These are heap dumps obtained with the [heapdump](http://npmjs.com/package/heapdump) module and viewed in Chrome's inspector, by the way.

rf:
The `validate` variable is declared because I passed it into the `new Function()` constructor on [line 131](https://github.com/epoberezkin/ajv/pull/293/files#diff-c2431ee2ac14bc35ae194a4341f9d217R131) which makes it an argument to the function being created.

rf:
So it turns out that the strings will still be retained by the code object to support `.toString()` on the function.

However, it may make sense to keep the `new Function()` change as it may make it easier to minify the generated javascript, which could potentially reduce the memory usage of the source strings. I'm going to investigate that tomorrow.

epoberezkin:
> The validate variable is declared because I passed it into the new Function() constructor on line 131 which makes it an argument to the function being created.

I see. It's a bit unusual to pass an undefined value simply to declare the local variable.

epoberezkin:
As far as I understand it, only the variables used inside closure are retained on the heap. So from the memory utilisation point of view it shouldn't make any difference whether you're using eval or function constructor...

Am I wrong?

It may or may not have difference on the time it takes to create this closure and also to execute it. But 1) closure creation time is not very important, 2) the difference can be negative - we will need to benchmark it to see if there is a positive difference.

epoberezkin:
> So it turns out that the strings will still be retained by the code object to support .toString() on the function.

It's a different string though, so probably keeping it in the property as well doubles the size

Re eval vs new Function I am not sure it has any difference.

rf:
It is unusual to define the variable this way, but I wanted to make the change in the least intrusive way possible.

So in fact, there's one context created for the outer function `compile` and any functions inside it will refer to this context. So even though the generated `validate` function doesn't use the `sourceCode` variable, it will still retain it. This is precisely what Vyacheslav Egorov describes in [this blog post](http://mrale.ph/blog/2012/09/23/grokking-v8-closures-for-fun.html).

I'm working on my own heap dump analysis tool; before my changes here, we see this:

```
 validate =  (functi(28735) type:string size:2176 deepsize:-1 ownership:
   retained as source type:internal in (code,28733)
   retained as sourceCode type:context variable in (closure,32415)
   retained as validateCode type:context variable in (closure,32415)
   retained as sourceCode type:property in (closure,32415)
   retained as (Compilation cache) type:hidden in system / Context(hidden,32517)
   retained as (Builtins) type:hidden in system / Context(hidden,32517)
   retained as (Extensions) type:hidden in (object properties)(array,32567)
```

There's a `source` edge, of type `internal`, from a `code` object. There's also a `property` edge from a `closure`, as well as `context` edges from that closure. So the string retained for `toString()` functionality is the same one that is used by `eval()` to compile the code.

It is most likely desirable to not create this closure at all. This indirect eval approach will likely save some amount of memory, though unfortunately it won't cause those strings to not be retained at all. After this change, I found this in the heap:

```
(function(refVal,co,(29693) type:string size:2304 deepsize:-1 ownership:
   retained as source type:internal in (code,29691)
```

This is the `new Function()` generated code; and there's still an `internal` edge from a `code` object. But at least no closure was created.

For information on `new Function()` vs `eval()` and why this matters, see [this blog post](http://www.2ality.com/2014/01/eval.html). Essentially, any "indirect" call to `eval()` will cause the code to be executed in the global scope instead of in the current scope. This means you can't access any local variables (which is why I had to pass them in) and it also won't close over any variables in the current scope.

I think it's beneficial to keep this change; it will create one less closure, and it also makes it easier to potentially minify the generated JS to save memory or create portable validation functions as was requested in #156.

epoberezkin:
Thank you, it's very interesting.
I am going to check that capturing the whole context with eval usage is still the case in node and modern browsers (Egorov's post is 4yo). That's a bit surprising - I thought v8 would be a bit more clever and only capture what's used :)

epoberezkin:
It does indeed capture everything with eval :)

Could you please:
- rename the property to `sourceCode` (it will be false by default in the next major version)
- remove validate from parameters, remove `+ 'return validate;'` from [line 103](https://github.com/epoberezkin/ajv/pull/293/files#diff-c2431ee2ac14bc35ae194a4341f9d217L103) and instead replace `validate =` with `return` in [validate.dot](https://github.com/epoberezkin/ajv/blob/master/lib/dot/validate.jst#L37). That would make it cleaner.
- add test to https://github.com/epoberezkin/ajv/blob/master/spec/options.spec.js
- add docs to https://github.com/epoberezkin/ajv#options

Then it's good to go.
Thank you again!

epoberezkin:
going to merge/complete...

rf:
Oh hey, I actually just finished addressing your comments, haha. But if you've done it already that's fine.

epoberezkin:
in [4.7.0](https://github.com/epoberezkin/ajv/releases/tag/4.7.0) :)

epoberezkin:
Thank you!

epoberezkin:
@rf https://medium.com/mail-online/eval-is-evil-but-not-why-you-may-think-25961f9b01bb

rf:
Nice article! Thanks for the shout out! 💯

PR #301, opened 2016-09-16
Add missing rootData parameter from ValidateFunction typescript interface

darcyparker:

This resolves a typescript validation error of `ajv.compile(someJSONSchemaObject).toString();`

I did some minor refactoring of the function like this (adding the function's interface):

``` typescript
import * as ajv from 'ajv';
export var validate: ajv.ValidateFunction = function validate (data, dataPath, parentData, parentDataProperty, rootData) {
  //...the body of compiled function
}
```

And I get this typescript error:

```
 Type '(data: any, dataPath: string, parentData: Object | any[], parentDataProperty: string | number, ro...' is not assignable to type 'ValidateFunction'.
```

> The problem is that `rootData` argument is not listed in the interface type definition.

I added the missing argument in the typescript interface type definition.

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -92,7 +92,8 @@ declare namespace ajv {
       data: any,
       dataPath?: string,
       parentData?: Object | Array<any>,
-      parentDataProperty?: string | number
+      parentDataProperty?: string | number,
+      rootData?: any
     ): boolean | Thenable<boolean>;
     errors?: Array<ErrorObject>;
     schema?: Object;
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -93,7 +93,7 @@ declare namespace ajv {
       dataPath?: string,
       parentData?: Object | Array<any>,
       parentDataProperty?: string | number,
-      rootData?: any
+      rootData?: Object | Array<any>
     ): boolean | Thenable<boolean>;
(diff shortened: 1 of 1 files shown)


lib/ajv.d.ts
```diff
@@ -92,7 +92,8 @@ declare namespace ajv {
       data: any,
       dataPath?: string,
       parentData?: Object | Array<any>,
-      parentDataProperty?: string | number
+      parentDataProperty?: string | number,
+      rootData?: any
```
epoberezkin: rootData should be: Object | Array<any>
Same as parentData

epoberezkin (changes requested):

lib/ajv.d.ts
darcyparker: Thanks @epoberezkin - I made that change.  You can squash the 2 commits if you want before you merge. I was guessing it had the same type as `data` because I saw https://github.com/epoberezkin/ajv/blob/master/lib/dot/validate.jst#L51 ...

Should `data` be updated so it is `Object | Array<any>` too?

epoberezkin: Thank you.
No, data can be anything, but parentData and rootData can only be array or object (otherwise you would never get to children).

epoberezkin (approved):

epoberezkin:
in 4.7.1

PR #327, opened 2016-10-18
fix #326 by lazy-requiring js-beautify

### LINKED ISSUE #326: js-beautify is loaded synchronously via try-catch, which is problematic when trying to get BI about require errors
3.4.0

We use AJV at Wix.com (thank you!) and we configured our requirej.onError to notify us whenever a script failed to load in require. This helps us monitor issues with loading scripts for our users.

js-beautify is loaded synchronously via try-catch during the bootstrapping of the lib, and assumes that beautify will only exist if you want to use it (and you pass the responsibility of the consumer of AJV to load js-beautify before AJV is loaded).

If we don't want to use the beautify option, which we don't, then we have a require error trying to load this script.

TLDR; require js-beautify lazily (but still synchronously), only if beautify option was specified.

Since you pass the responsibility of loading js-beautify to the consumer of AJV (and it must be loaded before AJV), I think that you should either:
- enable building a different (custom) bundle of AJV with options (beautify: false) so that it does not even try to require or use beautify (complex solution) 
- change API so consumer of AJV passes AJV the beautify function (probably a decent solution, but a bigger change)
- attempt to require js-beautify only if beautify:(true|options) was specified in the config, which is the only time beautify is used (much simpler). 

Yes. I'll attach a PR here.

danyshaanan:
https://github.com/epoberezkin/ajv/issues/326

"require js-beautify lazily (but still synchronously), only if beautify option was specified" (@EtaiG)

Not that I can think of.

---

`npm test` passed successfully.

If anything additional is required, please specify so that we could follow up on this.

Thanks.

### DIFF
--- a/lib/compile/index.js
+++ b/lib/compile/index.js
@@ -1,20 +1,11 @@
 'use strict';
 
 var resolve = require('./resolve')
-  , util = require('./util')
-  , stableStringify = require('json-stable-stringify')
-  , async = require('../async');
+    , util = require('./util')
+    , stableStringify = require('json-stable-stringify')
+    , async = require('../async');
 
-var beautify
-  , didWeTryToLoadBeautify = false;
-
-function loadBeautifyIfNeeded(){
-  if (!didWeTryToLoadBeautify) {
-    try { beautify = require('' + 'js-beautify').js_beautify; } catch(e) {/*empty*/}
-    didWeTryToLoadBeautify = true;
-  }
-}
-
+var beautify;
 var validateGenerator = require('../dotjs/validate');
 
 /**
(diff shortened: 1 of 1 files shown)


lib/compile/index.js
```diff
@@ -115,6 +123,7 @@ function compile(schema, root, localRefs, baseId) {
                    + sourceCode + 'return validate;';
 
     if (opts.beautify) {
+      loadBeautifyIfNeeded();
```
epoberezkin: Could we simplify a bit? Something like:

``` javascript
if (beautify === undefined) {
  (function() {
    try { beautify = require('' + 'js-beautify').js_beautify; }
    catch(e) { beautify = false; }
  })();
}
```

No need to create two new names really...

epoberezkin (changes requested):

lib/compile/index.js
EtaiG: sure, on it.

epoberezkin:
I merged it (with some corrections - removed whitespace changes and made it function back again...)
try catch should be in a separate function (that's why it was in closure in my sample), v8 doest use optimising compiler for the whole function with try/catch (see https://bugs.chromium.org/p/v8/issues/detail?id=1065).

EtaiG:
Cool, interesting to know 👍 

I didn't know that code inside an IIFE is not optimized as part of the function either (meaning the optimization won't take a hit for unoptimized code there)

EtaiG:
Also, can you let me know when the package.json is updated with a version I can use for this? 

Thanks for the quick turnaround!

epoberezkin:
In 4.8.1
Thank you, it was the right thing to do.

Issue #340, opened 2016-11-13
fast-stable-stringify


kapouer:
https://github.com/nickyout/fast-stable-stringify
might be a better maintained alternative to the current module.
It is GPL3 though.

epoberezkin:
@kapouer thanks for the suggestion, I will think about it. Maybe make a PR so that we can see that all tests are passing?

kapouer:
Thinking of which, i'm surprised you don't use the schema id to build a cache key (if it exists of course).
This would work unless the schema is mutable, and would avoid the cost of building that key.

epoberezkin:
id is optional... They also can be subschemas (when you use $refs).

alexkuz:
@epoberezkin is it possible to use schema itself as a cache key and move `stableStringify` in the default `Cache` object? That way one could replace `Cache` with `WeakMapCache` for example, that would use WeakMap with raw (immutable) schemas as keys.

epoberezkin:
@alexkuz I think it can be done in 5.0.0, need to think about it.
The downside is that if the same fragments exist in different places they will be compiled twice. Maybe it's an edge case.

Why do you want it, I wonder?

epoberezkin:
Actually, [sacjs](https://github.com/epoberezkin/sacjs) that can be used as an alternative already has json-stable-stringify inside...

alexkuz:
I'm using `ajv` with huge schemas and it goes really slow - mostly because of `json-stable-stringify` (I compared it with `JSON.stringify` on simple test - the latter goes 5 times faster). Maybe it will be better after I refactor my own code, it's quite unoptimized right now, but I suspect this would be still a bottleneck, which I could avoid using WeakMap.

epoberezkin:
I see. I suspected something like this... What is huge? I guess you retrieve by id, but it still takes some time to cache them... Also it can consume a bit of memory.

Feel free to make a PR (move stringify to cache and update docs) to 5.0.0 branch, I am going to release the next beta soon (although there will be some migration effort needed).

Is it maybe a good idea to always use weakmap if they are available? It'll just have to take into account that schemas can be Boolean, although it's an edge case when they'll be cached separately.

epoberezkin:
Probably better to make it a separate package

alexkuz:
Ok, I'll try to implement new caching later.

Here's the simple test I used:

```js
'use strict';

var getAjvInstances = require('./spec/ajv_instances');


var instances = getAjvInstances({
[...]

PR #341, opened 2016-11-14
Use fast-stable-stringify

kapouer:
That module claims to be faster, but is under GPL-3.
See also https://github.com/epoberezkin/ajv/issues/340


epoberezkin:
@kapouer it appears that GPL cannot be used inside MIT as GPL is more restrictive, according to this: http://stackoverflow.com/questions/3902754/mit-vs-gpl-license

@nickyout would you consider changing it to MIT?

btw, the build in fast-stable-stringify fails...

Issue #291, opened 2016-08-26
Extend format keyword to support all data types


jeremykentbgross:
The readme currently indicates that custom format validation is only applied to string properties. However my understanding from reading the draft 4 specifications is that any property could have a format attribute, it just so happens that all the predefined formats only apply to strings.

I have not (yet) checked if your implementation does indeed only support custom formats for strings and not other types, but that is what is currently stated in the readme. As noted in another issue, the readme currently broken/out of date in some other places, so perhaps this information is incorrect, however assuming it is correct, ideally when defining custom format validation functions you should be allowed to specifying the type they apply to.

There are many use cases for applying formats to other property types, some examples may be found even in software linked/recommend by the main json schema website.

I have several possible use cases for custom formats on other types in upcoming projects. For example on objects, the custom format matcher "isA:\w_" could trigger a custom format validation function to verify that the object represents data serialized from a class derived from the class matched by "\w_". While in theory this could be achieved with features such as "oneOf", that would assume the schema must know all possible descendant classes at the time it is authored, or be dynamically updated and recompiled whenever a new class factory is registered. This method on the other hand allows traversing inheritance tree information that may not be available to the schema, and even custom recursive validation by another schema found in said tree.

Many more examples abound, but the important thing is that the draft 4 specification implies custom formats can be applied to any type, unless i misunderstand their meaning. Ideally ajv stays in line with the standard on this point, and/or updates it's documentation to indicate it's existing compliance.

jeremykentbgross:
The \w's was followed by a star, representing regex for matching a word, but it's not visible when posted.

epoberezkin:
The draft 4 spec describes `format` keyword as optional, it only says that if it is implemented, some predefined attributes should be implemented too - see [7.2](http://json-schema.org/latest/json-schema-validation.html#anchor106). It does say that it "generally" applies to a limited set of data types, so applying it only to strings is not against the standard, although one may imply that other types should be possible too:

> A format attribute can generally only validate a given set of instance types. If the type of the instance to validate is not in this set, validation for this format attribute and instance SHOULD succeed.

At the same time it says:

> The "format" keyword is defined to allow interoperable semantic validation for a fixed subset of values which are accurately described by authoritative resources, be they RFCs or other external specifications.

What would be the example of such external spec for non-strings? Also to me it seems semantically wrong using it for anything but strings, so Ajv indeed only supports formats for strings.

[...]

PR #350, opened 2016-11-27
WIP: add format support to numeric types

nimerritt:
PR: for discussion
See https://github.com/epoberezkin/ajv/issues/291 for context.

### DIFF
--- a/lib/dot/definitions.def
+++ b/lib/dot/definitions.def
@@ -2,6 +2,7 @@
   {{
     var $lvl = it.level;
     var $dataLvl = it.dataLevel;
+    var $type = it.schema.type || '';
     var $schema = it.schema[$keyword];
     var $schemaPath = it.schemaPath + '.' + $keyword;
     var $errSchemaPath = it.errSchemaPath + '/' + $keyword;
--- a/lib/dot/format.jst
+++ b/lib/dot/format.jst
@@ -81,11 +81,15 @@
     var $isObject = typeof $format == 'object'
                     && !($format instanceof RegExp)
                     && $format.validate;
+    var $types = $isObject && $format.types || ['string'];
     if ($isObject) {
       var $async = $format.async === true;
       $format = $format.validate;
     }
   }}
+  {{? $types.indexOf($type) === -1 }}
+    {{# def.skipFormat }}
+  {{?}}
   {{? $async }}
     {{
       if (!it.async) throw new Error('async format in sync schema');
--- a/spec/ajv.spec.js
+++ b/spec/ajv.spec.js
@@ -377,26 +377,70 @@ describe('Ajv', function () {
 
 
   describe('addFormat method', function() {
-    it('should add format as regular expression', function() {
-      ajv.addFormat('identifier', /^[a-z_$][a-z0-9_$]*$/i);
-      testFormat();
+    describe.skip('without any type restrictions specified', function() {
+      it('should add format as regular expression', function() {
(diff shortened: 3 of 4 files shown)


lib/dot/definitions.def
```diff
@@ -2,6 +2,7 @@
   {{
     var $lvl = it.level;
     var $dataLvl = it.dataLevel;
+    var $type = it.schema.type || '';
```
nimerritt, line 5: This could be an array according to JSON schema spec?

spec/ajv.spec.js
```diff
@@ -377,26 +377,70 @@ describe('Ajv', function () {
 
 
   describe('addFormat method', function() {
-    it('should add format as regular expression', function() {
-      ajv.addFormat('identifier', /^[a-z_$][a-z0-9_$]*$/i);
-      testFormat();
-    });
+    describe.skip('without any type restrictions specified', function() {
```
nimerritt, line 380: I think a reasonable approach to this is: If format is not explicitly declared in the schema then custom formatters should not be applied?

lib/dot/definitions.def
epoberezkin: yes, can be any of basic types or integer

spec/ajv.spec.js
epoberezkin: I don't understand. If you mean that "if type is not specified", then it is not correct. Format should be applied string and number after this change, depending on format keyword value, but even if the type is not explicitly specified. So this schema `{ "format": "int32" }` would be valid for numbers only if they are int32 but all other types will be valid too. You need to think about keywords as about restrictions. Type keyword is also a restriction, but it does not affect format validation, data type does.

epoberezkin:
> @epoberezkin Perhaps it is okay to only apply the format if the type is explicitly declared in the schema?

see comment above

> I think it makes sense to restrict it to just string/numeric types explicitly declared in the schema

I think there should be a way to define format for any types. Type that the format applies to should be a part of format definition with string being the default (if type is not present in the definition).

EDIT: actually it's ok, only for string and number is fine, there is no use case really to use format for arrays/objects/boolean/null, other keywords should be used for them.

lib/dot/format.jst
```diff
@@ -81,11 +81,15 @@
     var $isObject = typeof $format == 'object'
                     && !($format instanceof RegExp)
                     && $format.validate;
+    var $types = $isObject && $format.types || ['string'];
     if ($isObject) {
```
epoberezkin, line 85: Let's not support formats that apply to multiple types, there is no real use case for it and it would make code simpler.

spec/ajv.spec.js
epoberezkin: By the way, rules are called based on the data type, not on the presence of type keyword. So the way you did it (format keyword added to two types) there can be some code added for both string and integer data type and the code in format.jst should skip adding code if type is incorrect (I think you're doing it already, and I suggested to simplify not to support arrays of types). So I don't think that type keyword should be explicitly specified for it to work.

epoberezkin:
Thank you, it's getting there it seems :) Some more tests and docs and it's done. Also, I am going to merge it into 5.0.0 branch - I will release the next beta version once it's finished.

epoberezkin:
@nimerritt finished and merged in 5.0.0

nimerritt:
Thanks for finishing this up for me :) Actually needed it again today, so 👍

PR #354, opened 2016-11-30
Update types for options from https://github.com/DefinitelyTyped/Defi…

domoritz:
`extendRefs` is current missing here. Why are there two typings anyway? I think https://github.com/DefinitelyTyped/DefinitelyTyped/blob/types-2.0/ajv/index.d.ts should be merged into this and then deleted. 

cc @plantain-00


plantain-00:
When I used ajv, I didn't aware of there is a typings in the library.
I'm OK with deleting the one in the DefinitelyTyped if this PR is merged.

domoritz:
In this PR, I only merged the options properties. There may be more things missing from these typings.

plantain-00:
One thing I'm worried about is that libraries' version is rarely updated just because it's typings is updated, so sometimes you want typings in the library updated, then you have to wait the library's next release.
Maybe the one in DefnitelyTyped is still needed, just another option.

There is a library, it bundles typings, and the auther is still updating the typings in DefinitelyTyped(https://github.com/DefinitelyTyped/DefinitelyTyped/pull/12974 ).
The vuejs bundled typings and DefinitelyTyped has one too.

epoberezkin:
@domoritz thank you, I will review/merge

> When I used ajv, I didn't aware of there is a typings in the library.

@plantain-00 it was added after, @blakeembrey was saying the preference is to have it here.

> One thing I'm worried about is that libraries' version is rarely updated just because it's typings is updated

I don't mind adding version on typings update.

domoritz:
Awesome. I think having the typings with the code and in only one place is better.

blakeembrey:
Ideally it would only exist in the libraries. I commented on @plantain-00's original PR, hope you don't mind 😄  Having it in two places is likely to cause confusion, and if releasing a new version is the concern you can actually just copy it locally and change `typeRoots` with TypeScript.

One thing to note about this PR - it breaks compatibility with TypeScript 1.x and should be released as a major version. As for removing the definition from DefinitelyTyped, the instructions are in https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package.

epoberezkin:
> One thing to note about this PR - it breaks compatibility with TypeScript 1.x

@blakeembrey thank you. I will merge it in 5.0.0 then.

epoberezkin:
It's merged and released as [5.0.0-beta.1](https://github.com/epoberezkin/ajv/releases), you can `npm install ajv@^5.0.0-beta`

plantain-00:
FYI: I started a PR to remove the types in DefinitelyTyped. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13028

Issue #389, opened 2017-01-03
Remove (or reduce) `addKeyword()`'s requirement that `keyword` is a valid JS identifier


farrago:
**What version of Ajv you are you using?**
4.3.10

**What problem do you want to solve?**
[Swagger](http://swagger.io/) API definitions use JSON Schema to define input and output values.  However, the swagger [spec for vendor extensions](http://swagger.io/specification/#vendorExtensions) (i.e. custom keywords) says they MUST start with `x-`.  

This is incompatible with `addKeyword()`'s requirement that keywords must be a valid JS identifier.  So as it stands, ajv can't be used to validate any custom keywords / vendor extentions in conjunction with Swagger.

**What do you think is the correct solution to problem?**

From looking at `keyword.js` I can't see why this test is in place as the name doesn't seem to be used as a JS identifier directly (i.e. all property accessors use bracket notation rather than dot notation, so any string is ok).

I also tested just commenting out the test, and was able to successfully add and validate against a keyword called `x-test-keyword`.

I presume I am missing something as it was deliberately added in commit 694e81684fcbc0e29ef787764779f42c2c4f5196.  I'm assuming it's something to do with macro function keywords (which I haven't used)?

Based on the above, my proposal is to entirely remove the constraint, or potentially only check it if a `macro` function is being specified.  Does either sound reasonable?

**Will you be able to implement it?**
Happy to, if the proposed solution seems reasonable?

epoberezkin:
I think it is used as property access somewhere in .jst files, probably in custom.jst, errors.def, maybe in compile/index.js too (the part that generates code?)

What is the limitation you propose instead? If it is identifier with optional hyphens inside (not starting or trailing) I don't mind.

The starting point it to write a couple of additional tests in custom.spec.js (or maybe just change camel case to dash case in some examples) to see where it fails.

farrago:
Thanks for the quick response!

You're right, it appears to be failing on compiled code.  If I comment out the `if` that checks for valid JS identifiers, then change `custom.spec.js`: `testEvenKeyword()` to use `x-foo` as the keyword instead of `even`, then all 6 tests that use testEvenKeyword() fail with errors that look like:

```
1) Custom keywords custom rules rule with "interpreted" keyword validation should add and validate rule:
     ReferenceError: foo is not defined
      at validate (eval at localCompile (lib\compile\index.js:139:26), <anonymous>:3:397)
      at shouldBeValid (spec\custom.spec.js:863:5)
[...]

PR #391, opened 2017-01-05
Allow custom keywords that are not valid JS identifiers (fixes issue #389)

farrago:
 
Custom keywords were required to be valid JS identifiers due to the way they were being used in the generated code.  This prevented the use of keywords with the `x-` prefix that is required by the [swagger](http://swagger.io) API definition spec.

See issue #389 for discussion.

This change fixes that by updating the only place that was using them directly as an identifier to use _brackets_ notation rather than _dot_ notation.  This brings the definition of `$schemaPath` in `definitions.def` in line with the way the keyword definition is accessed from `RULES.custom` at the top of `custom.jst`.

The validation of the keyword name has been changed to allow any name in this change, but this should be updated to either just remove the test or update it to a similar but more relaxed rule to the previous one 
before this change is included in the project.

**Tests**

This updates a number of the tests in `custom.spec.js` to use non-valid identifiers for the names of some of the custom keywords being tested (primarily by adding `x-` to the front of the name).

Test Plan:
- Run `npm test` and ensure all tests run and pass

### DIFF
--- a/lib/dot/definitions.def
+++ b/lib/dot/definitions.def
@@ -3,7 +3,7 @@
     var $lvl = it.level;
     var $dataLvl = it.dataLevel;
     var $schema = it.schema[$keyword];
-    var $schemaPath = it.schemaPath + '.' + $keyword;
+    var $schemaPath = it.schemaPath + '[\'' + $keyword + '\']';
     var $errSchemaPath = it.errSchemaPath + '/' + $keyword;
     var $breakOnError = !it.opts.allErrors;
     var $errorKeyword;
--- a/lib/dot/definitions.def
+++ b/lib/dot/definitions.def
@@ -3,7 +3,7 @@
     var $lvl = it.level;
     var $dataLvl = it.dataLevel;
     var $schema = it.schema[$keyword];
-    var $schemaPath = it.schemaPath + '[\'' + $keyword + '\']';
+    var $schemaPath = it.schemaPath + it.util.getProperty($keyword);
     var $errSchemaPath = it.errSchemaPath + '/' + $keyword;
     var $breakOnError = !it.opts.allErrors;
     var $errorKeyword;
--- a/lib/keyword.js
+++ b/lib/keyword.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var IDENTIFIER = /^.*$/i;
+var IDENTIFIER = /^[a-z_$][a-z0-9_$\-]*$/i;
 var customRuleCode = require('./dotjs/custom');
 
 module.exports = {
@@ -12,7 +12,7 @@ module.exports = {
 /**
  * Define custom keyword
  * @this  Ajv
- * @param {String} keyword custom keyword, should be a valid identifier, should be different from all standard, custom and macro keywords.
+ * @param {String} keyword custom keyword, should be unique (including different from all standard, custom and macro keywords).
  * @param {Object} definition keyword definition object with properties `type` (type(s) which the keyword applies to), `validate` or `compile`.
  */
 function addKeyword(keyword, definition) {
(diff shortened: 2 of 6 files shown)


lib/keyword.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var IDENTIFIER = /^[a-z_$][a-z0-9_$]*$/i;
+var IDENTIFIER = /^.*$/i;
 var customRuleCode = require('./dotjs/custom');
```
epoberezkin, line 4: `/^[a-z_$][a-z0-9_$\-]*$/i`, as you've suggested

lib/dot/definitions.def
```diff
@@ -3,7 +3,7 @@
     var $lvl = it.level;
     var $dataLvl = it.dataLevel;
     var $schema = it.schema[$keyword];
-    var $schemaPath = it.schemaPath + '.' + $keyword;
```
epoberezkin, line 6: it should be: `it.schemaPath + it.util.getProperty($keyword);`

This way it would still have direct property access when possible (which is in most cases and for all standard keywords)

epoberezkin (changes requested):
I will also need to double check that there are no other places where it can have direct property access.

epoberezkin:
Also docs needs to be updated here: https://github.com/epoberezkin/ajv#addkeywordstring-keyword-object-definition

I would even say that "it is recommended to use hyphens in application-specific custom keyword to prevent possible name collisions and to differentiate them"? Not sure.

lib/dot/definitions.def
farrago: I did try this, but it gets copied directly to the generate code rather than being interpreted first.
i.e. you get a line in `lib\custom.js` that looks like:
```javascript
var $schemaPath = it.schemaPath + it.utils.getProperty($keyword);
```

when what we wanted is:
```javascript
var $schemaPath = it.schemaPath.foo; // or it.schemaPath['x-foo'] as appropriate
```

I couldn't work out how to make it insert the result of `getProperty()` rather than the literal value. So I just used hardcoded brackets notation directly as done in `custom.jst`:

```
23:   var {{=$definition}} = RULES.custom['{{=$keyword}}'].definition;
```

Can you suggest how to make this work (I'm not very familiar with dotjs)?

As an aside, is there a reason to favour one over the other?  I did do a quick search for performance comparisons, and while there was nothing conclusive it seemed that for fixed strings, brackets and dots are very similar in terms of performance.  Using variables with brackets notation is obviously slower as it can't optimise the lookup, but we are using a hardcoded string so it should be ok.

epoberezkin: > i.e. you get a line in lib\custom.js that looks like:
> var $schemaPath = it.schemaPath + it.utils.getProperty($keyword);

If you mean the file lib\dotjs\custom.js then that's what it should be.
It's not a generated code of validation function, it's a code that generates code of validation function :)

epoberezkin: The only question is if it works or not.

epoberezkin: > I'm not very familiar with dotjs

It's a normal (and very simple) templating engine that allows JavaScript in it. Whatever is inside {{  }} is just code, so it is copied as is, while things in {{= }} get into the output.

It's a bit easier to manage template then it would be to manage string concatenation to generate code, that's it, but it's the same really.

farrago: Ahh, I was trying to use it as a template parameter (i.e. wrap it in {{}}) which was making all sorts of wrong results.  But just copying it in directly does the right thing!  Well that was an hour wasted to my confusion!

I'll update with this and the other comments shortly.

epoberezkin (approved):

epoberezkin:
Thank you!

epoberezkin:
in 4.10.4

PR #397, opened 2017-01-16
Allow _ in the domain of an email address

### LINKED ISSUE #396: Allow underscores in email domains
We're using ajv to validate customer data.
We have quite a few records which have an underscore in the domain part of the email address: `example@my_domain.com`

I have done a bit of research and it seems like ajv uses the regex mentioned in the HTML 5 specification. According to this, underscores are not allowed in the domain.
However in practise these email addresses are used and they do work (because underscores are allowed in dns entries).

Given my use case, the appropriate action is to be lenient and accept these email addresses.

I am mainly creating this issue to see whether it is worth opening a PR with the changed regex.

wdullaer:
This pull request updates the email regex to allow _
It includes some tests to verify the behaviour.
Fixes #396


lib/compile/formats.js
```diff
@@ -33,7 +33,8 @@ formats.fast = {
   // email (sources from jsen validator):
   // http://stackoverflow.com/questions/201323/using-a-regular-expression-to-validate-an-email-address#answer-8829363
   // http://www.w3.org/TR/html5/forms.html#valid-e-mail-address (search for 'willful violation')
```
epoberezkin, line 35: Please update these references to match the regex (the part(s) of RFC(s) that say that underscores are fine)

lib/compile/formats.js
```diff
@@ -33,7 +33,8 @@ formats.fast = {
...
-  email: /^[a-z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?(?:\.[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)*$/i,
+  // Modified to allow '_' in the domain
+  email: /^[a-z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-z0-9_](?:[a-z0-9-_]{0,61}[a-z0-9_])?(?:\.[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)*$/i,
   hostname: HOSTNAME,
```
epoberezkin, line 38: I suspect that the requirement to have hyphens only in the middle is also unnecessary for domains (we need reference as well to RFC), so regex can be substantially simplified to have fewer matching ranges. Can't it be email@-domain.com ? Or it is still invalid?

epoberezkin (changes requested):

lib/compile/formats.js
wdullaer: So I've done some more research here.
Turns out that hostnames can only contain letters/digits/hyphens.
However domain names can have mare characters
See this stackoverflow post with some references to RFCs: http://stackoverflow.com/a/2183140

So if I'm understanding this correctly, this would be valid: "foo@my_sub.domain.com"
But this wouldn't be: "foo@my_domain.com"

The email RFC says that the domain part is either a hostname or a dns domain name or an IP in square brackets (it basically refers to those RFCs)

epoberezkin: I think both are valid, strictly speaking... You won't be able to use my_domain.com as a host-name (as a part of URL) but you should be able to use it as a domain name. But practically given that everybody uses domains as hostnames maybe we shouldn't allow underscores in the second level and only allow it in the third etc.

epoberezkin: Having read the docs, I think that it should be extended even more rather than just adding underscores...

wdullaer: The more I read into this, the more confused I get.
"I am an email"@email.com should be valid apparently, and you an IP in the domain using square brackets should also be allowed me@[127.0.0.1] (although wikipedia references some article that mentions this will trigger quite a lot of spam filters, so in practice people probably shouldn't be doing this)

Regarding underscores: most registrars don't allow you to register something with an underscore in the root domain I've found, but eg AWS Route53 does allow you to put underscores in A records and CNAME records for subdomains (and applications can work with these without an issue)

(I'll probably be a bit slower over the next days to reply, things are rather busy at the moment)

lib/compile/formats.js
wdullaer: I think RFC 5321 is the current version of the SMTP standard.
Section 4.1.2 describes the structure of email addresses (among other things).
Here's a relevant section

```
   To promote interoperability and consistent with long-standing
   guidance about conservative use of the DNS in naming and applications
   (e.g., see Section 2.3.1 of the base DNS document, RFC 1035 [2]),
   characters outside the set of alphabetic characters, digits, and
   hyphen MUST NOT appear in domain name labels for SMTP clients or
   servers.  In particular, the underscore character is not permitted.
   SMTP servers that receive a command in which invalid character codes
   have been employed, and for which there are no other reasons for
   rejection, MUST reject that command with a 501 response (this rule,
   like others, could be overridden by appropriate SMTP extensions).
```
So they explicitly prohibit the use of underscore's even in cases where DNS would actually accept them.

Also if you look at how domain names are defined:
```
   Domain         = sub-domain *("." sub-domain)
   sub-domain     = Let-dig [Ldh-str]
   Let-dig        = ALPHA / DIGIT
   Ldh-str        = *( ALPHA / DIGIT / "-" ) Let-dig
```
A domain name cannot begin with a hypen, only a letter or digit

I think the current regex is actually a good representation of the RFC.

I have seen email addresses with underscores in subdomains which seem to work, but if the RFC explicitly rules them out, I don't think it's worth putting it in the regex. Servers that accept them must be considered faulty, and users that have these kind of email addresses will bump into problems sooner rather than later.

In short, after doing more research, I think it's probably best to close my PR and leave the regex as is.

epoberezkin: @wdullaer Ok, thank you for your time spent to clarify it. In any case, if somebody needs to support emails with underscores they can easily update the format to allow it.

PR #426, opened 2017-02-23
Cross platform "bundle"

### LINKED ISSUE #206: Support creating ajv bundle in Windows
Hi, 

While install ajv throug npm I'm getting error. Attached is the error log "npm-debug.txt"
[npm-debug.txt](https://github.com/epoberezkin/ajv/files/302571/npm-debug.txt)

I was also not able to create separate UMD bundle for browser using npm run bundle script. Attached in the error log "npm-run-bundle-debug.txt" Is it required that I install ajv before creating the bundle? 
[npm-run-bundle-debug.txt](https://github.com/epoberezkin/ajv/files/302577/npm-run-bundle-debug.txt)

Let me know what could be the problem here.

Thanks.

boenrobot:
Replaced the "bundle" script with a cross platform Node script called "bundle.js" (modeled after the original).

Added "del-cli" as a dev dependency, as a cross platform replacement of "rm".

Changed the quotes in the "if-node-version" call to doubles, to prevent dummy file being created on Windows.

Fixes #206.

EDIT: I don't think the test failures are related to this... are they?


epoberezkin:
@boenrobot Sorry, just noticed your comment. Yes, the test failures are definitely related to the change. The error message "generators, nodent and regenerator are not available" means that neither nodent bundle nor generator bundle doesn't allow to require them.

Compare the output after the change: https://travis-ci.org/epoberezkin/ajv/jobs/204781630#L2329
and before the change: https://travis-ci.org/epoberezkin/ajv/jobs/207894092#L2950

The new output is practically empty, which means that the bundle file is most likely empty before it gets to uglify (that generates all this "unused variables" output)

epoberezkin:
I see. Thank you, I will review/merge it later today or tomorrow.

epoberezkin:
in 4.11.5

epoberezkin:
@boenrobot now the problem is to make it work with uglify-js 3.x.

PR #493, opened 2017-05-18
Tweaked bundle.js to support UglifyJS3.

boenrobot:
As requested in #426 😉 .


epoberezkin:
That was quick :) Thank you!

PR #544, opened 2017-08-06
fix: use uri-js to resovle uri's

### LINKED ISSUE #423: Fails to find and validate schemas that are defined as a $ref by URN
## The code is available on runkit:
https://runkit.com/sondrele/58a88642b5e12e0013c5bf15

4.11.2 (also tested on 5.0.2-beta.0)

**Ajv options object (see https://github.com/epoberezkin/ajv#options):**

```javascript
var Ajv = require('ajv@4.11.2');
ajv = new Ajv({
    v5: true,
    allErrors: true,
    errorDataPath: 'property',
    verbose: true,
    missingRefs: "fail",
    unknownFormats: 'ignore',
    multipleOfPrecision: 12,
    extendRefs: true,
    passContext: true,
    i18n: false
});

```

**JSON Schema (please make it as small as possible to reproduce the issue):**

```json
var schema = {
    "type": "object",
    "properties": {
        "ip1": {
            "id": "urn:some:ip:prop",
            "type": "string",
            "format": "ipv4"
        },
        "ip2": {
            "$ref": "urn:some:ip:prop"
        }
    },
    "required": [ 
        "ip1", 
        "ip2"
    ]
};

```

**Data (please make it as small as posssible to reproduce the issue):**

```json

var data = {
    "ip1": "0.0.0.0",
    "ip2": "0.0.0.0"
};


```

**Your code (please use `options`, `schema` and `data` as variables):**

```javascript
console.log(ajv.validate(schema, data));

```

**Validation result, data AFTER validation, error messages:**

```
false

```

I had expected the data to be successfully validated, instead the validation ends with the error message `"can't resolve reference urn:some:ip:prop from id #"`.

It seams like there might be a problem with the `id` being encoded as URN's, because the validation will succeed if we change the `"id": "urn:some:ip:prop"` to e.g. `"id": "urn:some/ip:prop"` (notice that I've only replaced the second `:` with a `/`).

When I'm looking at the object that keeps references to the internal schemas, I can see that the key for the `ip1` schema gets translated to `urn:some/ip:prop`, so that when ajv tries to validate `ip2` it fails to find the schema by it's original id.
It looks like the id's of the schemas might be treated like URL's instead of URN's, but I can't say this for sure.

sondrele:
This is a PR to resolve #423 

In the issue you mentioned that it should maybe be part of the JSON-schema-test-suite, but juding from the last comment on the issue, it might not be necessary.

Replaced usage of the node `url` package with `uri-js` for resolving url's and schema id's.

This changes a pretty central part of url resolving, but it seems like it should be covered by the tests.

Let me know if you'd like to do some other changes, if it need more tests, or implement it some other way.

### DIFF
--- a/lib/compile/resolve.js
+++ b/lib/compile/resolve.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var url = require('url')
+var URI = require('uri-js')
   , equal = require('fast-deep-equal')
   , util = require('./util')
   , SchemaObject = require('./schema_obj')
@@ -67,7 +67,7 @@ function resolve(compile, root, ref) {
  */
 function resolveSchema(root, ref) {
   /* jshint validthis: true */
-  var p = url.parse(ref, false, true)
+  var p = URI.parse(ref)
     , refPath = _getFullPath(p)
     , baseId = getFullPath(this._getId(root.schema));
   if (refPath !== baseId) {
@@ -115,9 +115,9 @@ var PREVENT_SCOPE_CHANGE = util.toHash(['properties', 'patternProperties', 'enum
 /* @this Ajv */
 function getJsonPointer(parsedRef, baseId, schema, root) {
   /* jshint validthis: true */
-  parsedRef.hash = parsedRef.hash || '';
-  if (parsedRef.hash.slice(0,2) != '#/') return;
-  var parts = parsedRef.hash.split('/');
+  parsedRef.fragment = parsedRef.fragment || '';
+  if (parsedRef.fragment.slice(0,1) != '/') return;
+  var parts = parsedRef.fragment.split('/');
 
   for (var i = 1; i < parts.length; i++) {
     var part = parts[i];
@@ -206,14 +206,13 @@ function countKeys(schema) {
 
 function getFullPath(id, normalize) {
   if (normalize !== false) id = normalizeId(id);
-  var p = url.parse(id, false, true);
+  var p = URI.parse(id);
   return _getFullPath(p);
 }
 
 
 function _getFullPath(p) {
-  var protocolSeparator = p.protocol || p.href.slice(0,2) == '//' ? '//' : '';
-  return (p.protocol||'') + protocolSeparator + (p.host||'') + (p.path||'')  + '#';
+  return URI.serialize(p).split('#')[0] + '#';
 }
 
 
@@ -225,7 +224,7 @@ function normalizeId(id) {
 
 function resolveUrl(baseId, id) {
   id = normalizeId(id);
-  return url.resolve(baseId, id);
+  return URI.resolve(baseId, id);
 }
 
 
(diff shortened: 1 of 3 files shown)


epoberezkin:
I'll have a look a bit later

epoberezkin:
@sondrele I'm a bit reluctant to switch from internal url to uri-js for several reasons:

1. the use case when URIs are used in $ref is not common.
2. it may have some differences in behaviour compared to `url` that is difficult to identify/test.
3. the bundle becomes 4kb bigger.

I think a better approach could be to allow passing uri-js (or any other library) as an option and specify in docs what methods it should support. What do you think?

sondrele:
@epoberezkin I understand your concerns, both about size, and about a possibly different behavior (although that could maybe be solved with a version bump). It's a little unfortunate that not all urn's are supported, but I guess it's fine as long as it's documented as a limitation.

As a curiosity (and side-note), have you considered looking into replacing the legacy `url.parse()` and `url.resolve()` functions from the `node-url` package, with the `whatwg-url` API instead? That should help decrease the bundle size even further (and maybe offer improved, native performance, for all I know), if url's are going to be the main concern for Ajv.

An option to configure the `id` resolving would work for my case above, and it might also allow for some other use-cases where opaque identifiers are all that's needed. What's your stance on the API?

epoberezkin:
@sondrele 

> have you considered looking into replacing the legacy functions from the node-url package, with the whatwg-url API instead?

That would make it more difficult to allow supporting all URIs, no?

> What's your stance on the API?

I thought it would be just an option "uri" that should be an object compatible with the current "legacy" url package API (i.e. supply methods parse and resolve, where parse returns an object with properties (at least) protocol, href, host, path). If it is not supplied, "url" will be used.

sondrele:
> That would make it more difficult to allow supporting all URIs, no?

Probably, but I thought I'd just throw it in as a different issue, since URL id's are going to be the main concern for Ajv. :)
The "legacy" part comes from [this issue](https://github.com/nodejs/node/issues/12682) and [comment](https://github.com/nodejs/node/issues/12168#issuecomment-291548561) on the node.js issue tracker. They seam to talk about deprecation of `url.parse()`, but not for the foreseeable future, judging from the comments.

 I'll get around looking into a `uri` option when time allows :)

sirugh:
Using ajv for schema compilation in the browser is proving difficult because of the use of the node `url` module. We're currently wrestling with rollup/webpack trying to figure out a way around it but I personally would love to see this issue/PR handled. This library's tag-line is "The fastest JSON Schema validator for Node.js and browser" but it can't run natively in the browser, at least not when used as a dependency of our own bundle.

epoberezkin:
@sirugh your comment is unrelated to the problem this PR addresses (support for full URI spec in refs rather than just URLs). Please submit a specific example that doesn’t work in the browser for you together with how you load ajv. At the very least I expect it to work ok with the provided bundle, so maybe your issue can be addressed with the changes in how you build it?

sirugh:
You're right. I'm just at wits end here trying to figure it out. I'll try to come up with a reasonably simple use case that displays the error I'm talking about.

Issue #129, opened 2016-02-22
removeAdditional doesn't work well with anyOf, oneOf or allOf


gtrindade:
Hi, first of all, awesome work with this validator! Very handy!

Now here is the problem I'm facing, I'm not sure if this is the expected behavior, but if it is, it's not very intuitive... my scenario is a bit more complex, but I'll simplify in the following example.

I'm initializing ajv with:

``` javascript
const ajv = Ajv({removeAdditional: true})
```

This is the sample schema:

``` javascript
const schema = {
    type: `object`,
    additionalProperties: {
      anyOf: [{
        type: `object`,
        properties: {
          a: {
            type: `string`
          }
        },
        required: [`a`],
        additionalProperties: false
      }, {
        type: `object`,
        properties: {
          b: {
            type: `string`
          }
        },
        required: [`b`],
        additionalProperties: false
      }]
    }
  }
```

It should accept an object with as many objects as we want, and each of these objects should have at least one property called `a`, or one called `b`. If `a` or `b` are not present, it should fail. It should also discard any other property that is not `a` or `b`.

Now, with this data:

``` javascript
  const data1 = {
    obj1: {
      a: `a`
    },
    obj2: {
      b: `b`
    },
    obj3: {
      c: `c`
    }
  }
```

It should fail only on `obj3`, but that's only the case when `removeAdditional` is `false`. If `removeAdditional` is `true`, it seems like as soon as it evaluates the first option of the `anyOf`, it removes the additional properties of the object before trying the second option.

the removeAdditional is very useful, but it needs to wait until all possibilities of oneOf to be evaluated before removing anything.

This is the error I get by running the code above:

``` javascript
     Error: the array [
  {
    "dataPath": "['obj2']"
    "keyword": "required"
    "message": "should have required property 'a'"
    "params": {
      "missingProperty": "a"
    }
    "schemaPath": "#/additionalProperties/anyOf/0/required"
  }
  {
    "dataPath": "['obj2']"
    "keyword": "required"
    "message": "should have required property 'b'"
    "params": {
      "missingProperty": "b"
    }
    "schemaPath": "#/additionalProperties/anyOf/1/required"
  }
  {
    "dataPath": "['obj2']"
    "keyword": "anyOf"
    "message": "should match some schema in anyOf"
    "params": {}
    "schemaPath": "#/additionalProperties/anyOf"
  }
  {
    "dataPath": "['obj3']"
    "keyword": "required"
    "message": "should have required property 'a'"
[...]

PR #553, opened 2017-08-31
keep properties of invalid schemas

sberan:

https://github.com/epoberezkin/ajv/issues/129#issuecomment-325155599

Only remove additional properties when the schema is valid.

### DIFF
--- a/lib/dot/properties.jst
+++ b/lib/dot/properties.jst
@@ -57,7 +57,9 @@ var {{=$nextValid}} = true;
 {{? $ownProperties }}
   var {{=$dataProperties}} = undefined;
 {{?}}
-
+{{? $removeAdditional }}
+  var additionalToRemove{{=$lvl}} = [];
+{{?}}
 {{? $checkAdditional }}
   {{# def.iterateProperties }}
     {{? $someProperties }}
@@ -86,7 +88,7 @@ var {{=$nextValid}} = true;
       if (isAdditional{{=$lvl}}) {
     {{?}}
     {{? $removeAdditional == 'all' }}
-      delete {{=$data}}[{{=$key}}];
+      additionalToRemove{{=$lvl}}.push({{=$key}});
     {{??}}
       {{
         var $currentErrorPath = it.errorPath;
@@ -97,7 +99,7 @@ var {{=$nextValid}} = true;
       }}
       {{? $noAdditional }}
         {{? $removeAdditional }}
-          delete {{=$data}}[{{=$key}}];
+          additionalToRemove{{=$lvl}}.push({{=$key}});
         {{??}}
           {{=$nextValid}} = false;
           {{
@@ -121,7 +123,7 @@ var {{=$nextValid}} = true;
               if (errors) validate.errors.length = errors;
               else validate.errors = null;
             }
-            delete {{=$data}}[{{=$key}}];
+            additionalToRemove{{=$lvl}}.push({{=$key}});
           }
 
           {{# def.resetCompositeRule }}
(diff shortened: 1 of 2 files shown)


epoberezkin (changes requested):
Thank you for the PR!

1. The change of behaviour is only possible for a specific value of removeAdditional option, as you proposed in https://github.com/epoberezkin/ajv/issues/129#issuecomment-325134083, it cannot be done for all option values. Not sure if "afterValidation" is clear enough, I will think about it, but the main point is that it should be something specific. Current values should keep the current behaviour.
2. In addition to the test you added (where the option value should be changed), there should be two more tests 
    - for the option `removeAdditional: true` the property is removed even though the validation failed (the current behaviour)
    - for the new option value the property is removed when validation passed

An open question is how would be handled the situation when additionalProperties is a schema? There three possibilities
1. just preserve old behaviour
2. we need two new option values (new behaviour for current options `false` and `"fail"`)
3. create a completely new option, e.g. `removeAdditionalOnSuccess`: true

I need to sleep on it, but the last choice seems better.
For choices 2 or 3 there will be needed additional 3 tests (same as above) for cases when additionalProperties is a schema.

What do you think?

spec/options.spec.js
```diff
@@ -43,43 +52,39 @@ describe('Ajv Options', function () {
...
+        fizz: 'buzz'
       };
 
-      ajv.validate('//test/fooBar2', object).should.equal(true);
-      object.should.have.property('foo');
-      object.should.have.property('bar');
-      object.should.have.property('baz');
-      object.should.not.have.property('fizz');
+      console.log(ajv.getSchema('//test/fooBar2').toString());
+      ajv.validate('//test/fooBar2', failingObject).should.equal(false);
```
sberan: @epoberezkin I think I found a bug in master. When `additionalProperties` is a schema, properties are not validated (and this test fails at this location 😬 .

sberan: (it's only a problem if you supply an invalid additional property with an additional properties schema, and `removeAdditional: 'failing'` )

sberan: @epoberezkin I figured out the issue. `$nextValid` was not being unset when the additional property was removed, which caused the function to exit with no errors.

sberan:
@epoberezkin I can go with `removeAdditionalOnSuccess` - I've implemented it in this PR and added tests for all the cases noted. What do you think?

sberan:
@epoberezkin note that the diff is much easier to read if you ignore whitespace (`?w=1`) 😄

epoberezkin:
@sberan I will review, thank you

lib/dot/properties.jst
```diff
@@ -121,7 +131,8 @@ var {{=$nextValid}} = true;
               if (errors) validate.errors.length = errors;
               else validate.errors = null;
             }
-            delete {{=$data}}[{{=$key}}];
+            {{=$nextValid}} = true;
```
sberan, line 134: setting this field to true fixed the bug I mentioned.

epoberezkin:
@sberan Thank you!

I don't think it is going to work as expected (that is to only remove additional properties when otherwise current schema succeeds). In your code additional properties are removed when all other properties succeed, but it doesn't check that other keywords succeed, e.g. `required` or `minProperties` etc.

Also the changes in tests are difficult to follow. I suggest not touching the existing tests and instead add other tests that validate that without this new option even if schema fails additional props are removed anyway and with this option they don't.

Regarding the fix for 'failing' option - I still don't understand the issue. Maybe it's better to make it a separate PR or at least make a separate test case that demonstrates the issue.

epoberezkin:
Closing, it is old.

PR #592, opened 2017-10-15
Fix type definitions

epoberezkin:
@blakeembrey can you have a look please, particularly at https://github.com/epoberezkin/ajv/pull/592/commits/28386786fd9f6229652829a673745a1563e097c7 ?

If that's ok, then is it maybe ok to replace `declare var ajv` with `declare class Ajv` (and drop function form without new) in the next major version?

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -110,8 +110,13 @@ declare namespace ajv {
       parentDataProperty?: string | number,
       rootData?: Object | Array<any>
     ): boolean | Thenable<any>;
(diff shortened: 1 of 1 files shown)


blakeembrey:
Haven't tested it, but it looks good. If the next major version promotes using it only as a class, go for it! The `var ajv` form was mainly there to support the two different styles of creating the object (based on the docs a while back IIRC).

epoberezkin:
@Techniv I'm not sure it's always the case.

delagen:
Ready #606

PR #621, opened 2017-11-14
Update draft-06 meta-schema

stuartpb:
See json-schema-org/json-schema-spec#481 - the previously-published meta-schema was missing a definition for the "examples" array.

### DIFF
--- a/lib/refs/json-schema-draft-06.json
+++ b/lib/refs/json-schema-draft-06.json
@@ -57,6 +57,10 @@
             "type": "string"
         },
         "default": {},
+        "examples": {
+            "type": "array",
+            "items": {}
+        },
         "multipleOf": {
             "type": "number",
             "exclusiveMinimum": 0


epoberezkin:
I thought it was added in draft-07...

stuartpb:
Nope, it was in draft-06, that's why @handrews merged this upstream: https://tools.ietf.org/html/draft-wright-json-schema-validation-01#section-7.4

handrews:
@epoberezkin [definitely draft-06](https://tools.ietf.org/html/draft-wright-json-schema-validation-01#section-7.4) (https://tools.ietf.org/html/draft-wright-json-schema-validation-01#appendix-B).  No idea why it didn't make it to the meta-schema, although we kind of re-arranged how the meta-schema got published to the web site by removing the duplicate copies, and it's possible the duplicates were not actually duplicates.  But `examples` definitely belongs here.

Issue #618, opened 2017-11-08
Configuration for console logging


meirotstein:
**What version of Ajv you are you using?**
5.1.3

**What problem do you want to solve?**
Ajv uses console.warn to report some warnings to the console.
I had several issues with some consumers of my package (which uses Ajv) that on there site they have a customized version of the console object and they don't interested with the Ajv warnings, in some cases the customized version did not implemented the `warn` function what leads to an Error.

This force these kind of sites to change the implementation of the console object in order to use packages that are depended in Ajv - which is not so easy when it comes to a large scale sites of big firms.
 
**What do you think is the correct solution to problem?**
I think that there should be a new option: `consoleLog`, default true, that when set to false there will be no reports to the console object whatsoever. 

**Will you be able to implement it?**
Yes

epoberezkin:
@meirotstein that's good, thank you.

I suggest calling the option `logger` and support:
- `false` to suppress all logging (but not exceptions)
- a custom logger object with methods `log`, `warn`, `error`

By default Ajv would simply assign `console` to this option.
There should be tests for all cases for all existing calls to console.

I don't think there is any logging during validation apart from $comment keyword but it is off by default and it has its own option that enables it.

meirotstein:
HI @epoberezkin 

I like your suggestion - sounds better than just a simple boolean option.

As for the current logging - you can find a call for console.warn() [here](https://github.com/epoberezkin/ajv/blob/c73ff4435e0457a336209773c5aecf924ef3d74c/lib/ajv.js#L169), [here](https://github.com/epoberezkin/ajv/blob/c73ff4435e0457a336209773c5aecf924ef3d74c/lib/ajv.js#L383) and [here](https://github.com/epoberezkin/ajv/blob/c73ff4435e0457a336209773c5aecf924ef3d74c/lib/ajv.js#L389)

epoberezkin:
in 5.4.0

PR #624, opened 2017-11-19
Adding logger option

meirotstein:
This change it related to issue #618 

Added logger option which can get either custom logger object or `false` for no logging at all


epoberezkin (changes requested):
@meirotstein Thank you for the PR!

The problem with this implementation is that even though `logger` is the option of an instance, setting it in one instance will set it for all instances.

The implementation should allow to have ajv instances with a different value of this option.

lib/compile/logger.js
```diff
@@ -0,0 +1,41 @@
...
+ * @this Logger
+ * @param {String} level log level - log, warn or error
+ * @param {Array} args log argumets
+ * @private
+ */
+function _logFunc(level, args) {
+  if(this.logger && this.logger[level]) this.logger[level].apply(this, args);
+}
+
+function noop() {}
```
epoberezkin: This whole Logger class seems like an unnecessary indirection, to be honest.
It can be much simpler without it completely:
1. if undefined is passed - use console (in Ajv constructor)
2. if false is passed assign a noop object: `{log: noop, warn: noop, error: noop}`
3. otherwise just use what is passed

meirotstein: It seems to me that it is more safe to use the Logger class because:
1. it is more testable - I was able to write a standalone tests for all the logger scenarios (logger.spec.js), if I move the logic into Ajv ctor it will be much harder as I will have to find specific scenarios with Ajv that leads to every console call (log, warn, error) - and anyhow it might change in the future
2. partial custom logger implementation can leads to an error - this code checks for the log function existence before calling it

epoberezkin: 1. You won't have to test it if it doesn't exist. You only really need to test 3 scenarios for the option: logs to console/log to passed logger/doesn't log.
2. It essentially hides the missing method in passed logger. Throwing an exception is better. If a user want's to hide say warnings but not errors they just need to pass logger with a noop for a method.

meirotstein: 1. I see what you mean - still we have to keep in mind that the test is a bit implied as we test a scenario that we knows that currently happen to write to the log - the test does not control it so it is less "unit" then we would want.

2. a real life scenario: my package it used by a site that in a certain page the global `console` object is overridden by oracle portal (they use their own implementation of console, apparently..), the oracle version does implement the `warn` method - nor I or the site owners were aware of this and it ends up with an exception from Ajv that prevents some capabilities to work.
IMHO it is better to secure our code from such scenarios - if our code throws an exception due to bad input it is still our responsibility (unless it is part of the expected documented scenarios, of course..).

Anyhow - sorry for my insistence - at the end it is your call :)  if I still wasn't able to convince I will change the implementation as you suggested.

epoberezkin: Thank you. To address the problem above, whether with the passed object or with the default global console, rather than hiding the problem it is better to check that the object has all 3 methods during the instance initialisation (in the constructor) and throw an exception if it doesn't:

```
if (!(logger.log && logger.warn && logger.error)) throw new Error('logger is invalid');
```

meirotstein: this would not help with the scenario I described as at the initialization phase it was still the global console object, that was later changed by the oracle portal page.
My way to deal with it will be to disable the log once this change will be rolled out.

epoberezkin: It would, because once console is assigned to the instance the fact that oracle overridden a global wouldn't have affected it. It would still continue using the old one.

meirotstein: correct :)
anyhow, it still throws an exception.. think of a package that uses a package that uses Ajv in an oracle page - quite a mess...

epoberezkin: The difference is that it would throw an exception early and always rather than late and sometimes (in case console is replaced before ajv initialisation; if it happens after as you wrote, it would simply work). I prefer this behaviour to simply not having warnings.

meirotstein: Let's agree to disagree then :)
I will push a change soon.

epoberezkin (approved):

meirotstein:
Thanks!
When do you expect it to be merged?

epoberezkin:
in 5.4.0

Issue #625, opened 2017-11-21
Make ajv chainable


pithu:
Thank you for providing this great and awesome library.

I would like to ask for the following improvement. 

What do you think about making some ajv method returning `this`, that would allow to write more compact code by chaining methods. For example the following code would be possible:

```javascript
const validator = new Ajv()
    .addSchema(someSchema)
    .addMetaSchema(someMetaSchema)
    .addFormat('customeType', customValidator)
    .getSchema('customUri')

if (!validator(someJsonObject)) { // fail }
```

As far as i can see candidates for chaining would be `addSchema`, `addMetaSchema`, `addFormat`, `addKeyword`.

epoberezkin:
Looks good. And drop “add”? :)

epoberezkin:
Also removeSchema and removeKeyword can be chained, although they aren’t used too often.
@pithu if you do the PR please update .d.ts file as well

epoberezkin:
I am now thinking whether renaming all methods (maybe leaving old names as aliases) is a good idea:

addSchema -> schema
addMetaSchema -> metaSchema
addFormat -> format
addKeyword -> keyword
getSchema -> get
removeSchema -> remove

pithu:
I am fine with the current naming prefix `add*` and `remove*`, because imo it's more explicit and better readable.

epoberezkin:
in 5.5.0

pithu:
cool, thx for merging

isTravis:
Forgive me if I've misunderstood the implementation of this, but it seems that addSchema is not chainable if an array of schemas is submitted.

I made a quick pull-request: #640.

pithu:
You are right, i forgot the `this` at that place

PR #629, opened 2017-11-22
Add method chaining to all add* and remove* methods

pithu:
This pull request add method chaining to all add* and remove* methods.

According to issue #625 this enables more compact code like

```
new Ajv().addSchema(mySchema).validate(schema, data)
```
I am still thinking about, how to get rid of the `add` ;)

I add `return this` to 6 methods: `addSchema`, `addMetaSchema`,  `removeSchema`, `addFormat`, `addKeyword`, `removeKeyword`.
updated jsdoc and the typescript definitions. 

I added tests for the methods with an extra method test, except for `addMetaSchema`, because i couldn't find a good place for the test.

I hope i didn't oversee anything, if so let me know, would be great if it get merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,17 @@ if (!valid) console.log(ajv.errorsText());
 // ...
 ```
 
+or with method chaining
+
+```javascript
+// ...
+var validate = new Ajv()
+    .addSchema(cardSchema)
+    .getSchema('/card#/definitions/organisation');
+var valid = validate(data);
+// ...
+```
+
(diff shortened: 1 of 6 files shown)


epoberezkin:
Thank you. API section in readme needs to be updated as well and there are also several examples that could use chaining: https://github.com/epoberezkin/ajv#api

pithu:
Ok thx for your feedback, i will have a look and try to add something reasonable to the API doc.

epoberezkin (approved):

Issue #320, opened 2016-10-11
Missing dist folder for releases


dfederm:
For all releases, the dist folder is missing.

Take a look at how jQuery manages releases for example: https://github.com/jquery/jquery/commit/ee2e377494a882f043e6d8abc67ac6370ee83d9c

They fork off a branch, tag it, and commit the built files.

The reasoning for this is so that people can pull the package in via bower. Currently bower.json points to dist/ajv.min.js, but that file doesn't actually exist when pulling the package in using bower.

epoberezkin:
Yes, you have to build it so it's there... I will look at jquery approach, not sure I like the idea of release branches, at the moment they are just tags in master branch. I definitely don't like the idea of having dist in master.

It's been raised on several occasions: #132, #171, #186

I am not sure I am interested to support bower, npm seems to be more convenient these days.

dfederm:
Most people see npm as server-side and bower as client side. In my scenario I'm hoping to use this in the browser, so we'd need to consume this package through bower.

epoberezkin:
I don't think it's "most" nowadays any more. I'd say "some". Most people I know, work with, read, etc. switched to using npm for both client and server side, particularly given that there is appearing more and more isomorphic libraries. I think it's time to forget about bower - it seems to have outlived its usefulness...

epoberezkin:
I looked at jquery. They are not keeping version branches, their releases are just tags in the master branch. It results in distorted statistics that doesn't reflect contributions and instead shows the releases: https://github.com/jquery/jquery/graphs/code-frequency . Creating and keeping version branches would make branches difficult to manage.

Also, jquery is primarily browser library. Although you can use it in node, very few people do. Still, there is not a word of using it with bower in readme...

Ajv, on the other hand is isomorphic, so you still need to npm-install it if you use node.js (and validating only in the browser is a bad idea, you should validate in the server as well).

For some schemas you already can precompile them to importable modules (with ajv-cli) and include them in your application bundle to be used in the browser without ajv.

So I will not include browser bundles in the repo, instead one of these should be used:
- npm
- cdnjs
- github or bower with an additional step that builds bundles

ghost:
[...]

PR #669, opened 2018-01-09
Publish built versions to epoberezkin/ajv-dist

ghost:
As per discussion in #320 we decided to publish built version of ajv to a separate repo.

@epoberezkin So it's up to you to create both target repo and bower package :)


epoberezkin:
@dcbrwn I invited you to https://github.com/epoberezkin/ajv-dist/invitations

ghost:
Yep, it works for this repo too.

epoberezkin:
@dcbrwn could you please push some tag in the branch so we both can see it working? And after that force-push its removal so it's not merged?

Also, shouldn't bower.json be updated to point to a new repo?

epoberezkin:
Actually I see something got published there... Or should it be just me publishing it to bower?

ghost:
@epoberezkin I've pushed tag v6.0.0-publish-test-1 into my fork of ajv and it got published into the mirror. As for the repo, probably bower.json should point to the source, i. e. epoberezkin/ajv.

I intentionally unregistered my ajv-dist package after I tested that all works smoothly, since all in all you're the owner of ajv :)

_Update:_

"repository" property in bower.json doesn't matter when one registers package. According to [this](https://github.com/bower/spec/blob/master/json.md#repository) it indeed should point to source code. List of requirements for bower packages from [official docs](https://bower.io/docs/creating-packages/#register), doesn't state that it's needed either.

ghost:
@epoberezkin you think some additional testing or changes required? Let me know if so.

epoberezkin:
Thank you. Merged. Let's see how it works :)

epoberezkin:
@dcbrwn it seems like it doesn't remove files that should not be there...

ghost:
I thought that whole dist folder can be useful? or we should leave only ajv.min.js?

epoberezkin:
The current [dist](https://cdnjs.com/libraries/ajv) doesn't have regenerator/nodent, and ajv-dist does.

I think the dist folder should be removed after cloning from ajv-dist before copying the build result over, as otherwise the files removed in build remain there.

epoberezkin:
Probably just needs `rm -rf ../ajv-dist/dist && mkdir ../ajv-dist/dist` instead of `mkdir -p ../ajv-dist/dist`

Issue #671, opened 2018-01-11
errorDataPath: 'property' with additionalProperties: false results in faulty dataPath values with square brackets


lehni:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**

The error happens on the latest versions: 6.0.0 & 6.0.1. 

**Ajv options object**

```js
{
  errorDataPath: 'property'
}
```

**JSON Schema**
```js
{
  type: 'object',
  required: ['a'],
  additionalProperties: false,
  properties: {
    a: { type: 'string' }
  }
}
```

**Sample data**

```js
{
  a: 'test',
  additional: 1
}
```
**Your code**

```javascript
const Ajv = require('ajv')

const schema = {
  type: 'object',
  required: ['a'],
  additionalProperties: false,
  properties: {
    a: { type: 'string' }
  }
}

const ajv = new Ajv({ errorDataPath: 'property' });

const data = {
  a: 'test',
  additional: 1
}

const validate = ajv.compile(schema)
const valid = validate(data)
if (!valid) {
  console.log(validate.errors[0])
}
```

**Validation result, data AFTER validation, error messages**

```js
{
  dataPath: '[\'additional\']',
  keyword: 'additionalProperties',
  message: 'should NOT have additional properties',
  params: { additionalProperty: 'additional' },
  schemaPath: '#/additionalProperties'
}
```

**What results did you expect?**

The dataPath should be `'.additional'` without the square brackets, as this is not an array index.

**Are you going to resolve the issue?**

I don't know enough about Ajv to be able to.

Here's a runkit test-case for easy reproduction:

https://runkit.com/lehni/ajv-additionalproperties

epoberezkin:
That is the intended behaviour.

As [documented](https://github.com/epoberezkin/ajv#error-objects), by default ajv uses JavaScript property access notation. It includes dot-notation and square brackets notation; the latter is used in this case. Dot-notation is only used for properties that are both valid identifiers/reserved keywords and are defined in the schema. As this property is not defined in the schema it is reported as `"['additional']"` and not as `".additional"`, to avoid the cost of validation-time check of its validity for dot-notation.

If you need to parse property names in dataPath, it is better to use the option `jsonPointers: true`, because JS property access notation is difficult to parse, in general case.

lehni:
[...]

PR #674, opened 2018-01-12
Fix additionalProperties error with errorDataPath

lehni:
As discussed with @epoberezkin in https://github.com/epoberezkin/ajv/issues/671#issuecomment-357173944, here now the PR to add correct error messages for `additionalProperties` validation errors when `errorDataPath` is set to `'property'`

### DIFF
--- a/lib/dot/errors.def
+++ b/lib/dot/errors.def
@@ -94,7 +94,7 @@
   'false schema':  "'boolean schema is false'",
   $ref:            "'can\\\'t resolve reference {{=it.util.escapeQuotes($schema)}}'",
   additionalItems: "'should NOT have more than {{=$schema.length}} items'",
-  additionalProperties: "'should NOT have additional properties'",
+  additionalProperties: "'{{? it.opts._errorDataPathProperty }}is an invalid additional property{{??}}should NOT have additional properties{{?}}'",
   anyOf:           "'should match some schema in anyOf'",
   const:           "'should be equal to constant'",
   contains:        "'should contain a valid item'",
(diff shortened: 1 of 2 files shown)


lehni:
This probably should need a test for it too, but I couldn't figure out where to put it, and couldn't find a similar one for the same behavior for `required` [here](https://github.com/lehni/ajv/blob/42fcfa6693433c27bd7826b46e26d63e27a1b59e/lib/dot/errors.def#L115).

epoberezkin:
@lehni well spotted the test was missing for the other change. The tests should be somewhere here https://github.com/epoberezkin/ajv/blob/797dfc8c2b0f51aaa405342916cccb5962dd5f21/spec/errors.spec.js#L45 and here https://github.com/epoberezkin/ajv/blob/797dfc8c2b0f51aaa405342916cccb5962dd5f21/spec/errors.spec.js#L145 (for "required" errors).

At the moment it tests only dataPaths, should also check messages.

lehni:
@epoberezkin thanks for the pointer. Here now my added tests:

https://github.com/epoberezkin/ajv/pull/674/commits/4a12ca3b9214d6df9caa7019d80a043b44c6c9b8#diff-831dd6f83c36a271e002b7770f3dbac9R67

I hope these suffice?

Also, please let me know if `'is an invalid additional property'` sounds like a good message to you. The wording can change. Alternatives to `invalid` are `illegal` or `unsupported`

epoberezkin:
@lehni thank you. I've merged. Maybe just "is an additional property"?

lehni:
I think the message should somehow state that additional properties aren't allowed. `should NOT have additional properties` implied that. But how to phrase it for the message that is directly attached to the property?

PR #678, opened 2018-01-18
Add ESLint to "Some packages using Ajv" section

platinumazure:

None

Added ESLint as a project/package that uses Ajv. (We use Ajv for validating lint rule options against option schemas, which are configured as JSON schemas.)

I don't think so.


epoberezkin:
Thank you!

Issue #668, opened 2018-01-09
schemaId type definition does not include "auto"


mazswojejzony:
Currently `schemaId` option is defined as follows:
```javascript
schemaId?: '$id' | 'id';
```
It should be updated to include the recently added "auto" mode:
```javascript
schemaId?: '$id' | 'id' | 'auto';
```

PR #689, opened 2018-01-29
Added 'auto' into schemaId type definition #668

keyten:
I just added 'auto' into ajv.d.ts, nothing more.

Issue #668


epoberezkin:
closes #668

PR #716, opened 2018-02-22
[Doc] Make .compile API clearer

wla80:


epoberezkin:
@wla80 thank you for the PR.
This change makes it incorrect for both paragraphs, I am afraid.
`null` is the value of `errors` property, not a property.
Also “unless” is “if not”, not “if”...

wla80:
Hello @epoberezkin , thanks for your feedback. By reading the paragraph, I thought `null` is one of the property. Sorry about this. Please let me know if my newest revision is inappropriate.

epoberezkin:
Your first revision just highlights that my wording is not clear, I'll read the last version.
Thank you!

PR #732, opened 2018-03-09
Add FoalTS/foal to 'Some packages using Ajv'

LoicPoullain:

None.

Added FoalTS as a project/package that uses Ajv. FoalTS is a Node.JS framework to quickly build web apps in TypeScript. Ajv is used to control the data received by the server.

No


epoberezkin:
Sorry, npm provides a better list of packages that depend on ajv. Only de-facto standard libraries or ajv plugins need to be added to the docs.

Issue #687, opened 2018-01-25
ES6 module export


aubergene:
I'm using [Rollup](https://rollupjs.org/) to build my project. It would be good if AJV had ES6 module exports so that it could make full use of Rollup.

I could have a go at implementing this if you like

epoberezkin:
@aubergene It'll be just a wrapper, correct?

epoberezkin:
It's been migrated to typescript, so should be an issue of changing the compilation target probably. Please check v7-alpha and comment if something is still needed - closing

PR #748, opened 2018-03-30
feat: ES module default export interop

bradleyayers:

#687 

- Added a `default` property on the package top-level module.
- Changed the TypeScript definition to use an ES module default export. This will require TypeScript consumers to change `import Ajv = require('ajv')` to `import Ajx from 'ajv'`.

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -349,4 +349,4 @@ declare class MissingRefError extends Error {
   missingSchema: string;
 }
 
-export = ajv;
+export default ajv;
--- a/lib/ajv.js
+++ b/lib/ajv.js
@@ -11,6 +11,7 @@ var compileSchema = require('./compile')
   , util = require('./compile/util');
 
 module.exports = Ajv;
+module.exports.default = Ajv;
 
 Ajv.prototype.validate = validate;
 Ajv.prototype.compile = compile;


epoberezkin:
> Changed the TypeScript definition to use an ES module default export. This will require TypeScript consumers to change import Ajv = require('ajv') to import Ajx from 'ajv'.

It should be in the next major version then.

bradleyayers:
> It should be in the next major version then.

I've amended the comment to include `BREAKING CHANGE`.

thomasgwatson:
Yay, thanks for this

SelfMade1101 (approved):

CrispusDH:
any updates?

moderndeveloperllc:
This would be so useful. Not only for TS folks, but for those of us writing web components which are ES6. In my use case, I would like to do some browser-side validation before I send off an API call. I have a web component for all the API work that this would fit quite nicely in if I could import it properly.

nicferrier:
hello! is there  another solution to this? or is it just stalled?

giacomo-cosimato:
Is this going to be merged anytime soon?

marcospassos:
Super useful!

kuldeepdaftary:
Any updates on this please @epoberezkin?

epoberezkin:
As I wrote above, it should be merged in the next major versions as it requires the update of the code that uses it:  https://github.com/epoberezkin/ajv/pull/748#issuecomment-377728201

alexs-mparticle:
Love this change and I'm eager to use it in my project. Any idea when it'll be ready to go?

asap:
Love this change and it'll help in my current project. Any idea when this will be ready to go?

epoberezkin:
It’s switched to ESM/typescript modules in v7 - see Getting started.

Issue #760, opened 2018-04-18
[typescript] Format validators need to take a string


jwalton:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**

Latest

**Your code**

```ts
ajv.addFormat({
    type: 'number',
    validate: (x: number) => x < 10
});
```

**What results did you expect?**

This should compile, but it doesn't, because [FormatValidator](https://github.com/epoberezkin/ajv/blob/master/lib/ajv.d.ts#L162-L168) needs to take a string as a parameter, and because FormatDefinition is not allowed to have a "type" field.

**Are you going to resolve the issue?**

Yes.  PR incoming.  :)

PR #761, opened 2018-04-18
Fix typescript defs to allow numeric validators.

jwalton:

https://github.com/epoberezkin/ajv/issues/760

I fixed it so `FormatDefinition` is now split into `StringFormatDefinition` and `NumberFormatDefinition`.  `NumberFormatDefinition` requires the "type" field, and requires it to be "number", whereas in StringFormatDefinition "type" is optional, but if there it has to be "string".

The upshot of this is, if you don't specify "type" or you specify it as "string", then you must supply a `validate` function that takes a string as the first parameter.  If you specify "type" as "number", then you must specify a `validate` function that takes a number.  Same for `compare`.

No.

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -1,4 +1,4 @@
-declare var ajv: { 
+declare var ajv: {
   (options?: ajv.Options): ajv.Ajv;
   new (options?: ajv.Options): ajv.Ajv;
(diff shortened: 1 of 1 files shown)


jwalton:
(Tested on my local project via `npm link`)

epoberezkin:
Thank you!

PR #764, opened 2018-04-20
support RegExp flag

thinkoliao:
support RegExp flag for "pattern" keyword at JSON Schema Validation.


epoberezkin:
Flags are not part of JSON schema spec, so supporting them in “pattern” keywords is not correct - the whole string should be treated as a pattern. You can use custom regexp keyword from ajv-keywords.

PR #771, opened 2018-04-27
passContext in recursive $ref

### LINKED ISSUE #768: passContext is not used with "$ref" referenced schema

6.4.0, yes

**Ajv options object**

```javascript
{ passContext: true }
```

**JSON Schema**

```json
{
	"$id": "https://www.example.com/schemas/resource.json",
	"type": "object",
	"properties": {
		"leaf": {
			"type": "string",
			"leaf": true
		},
		"resources": {
			"$ref": "resource.json"
		}
	}
}
```

**Sample data**

```json
{
	"leaf": "first",
	"resources": {
		"leaf": "second"
	}
}
```

**Your code**

```javascript
ajv.addKeyword('leaf', {
	type: 'string',
	validate: function(schema, data) {
	// for second leaf is false
        console.log(this && this.context === 'my-context')
        return true
    }
})

validate.call({ context: 'my-context' }, {
    leaf: 'one',
    resources: {
        leaf: 'second'
    }
})

```

`this` from `validate.call(this, data)` === `this` from custom keyword function
in test is not true for schema referenced with `$ref`

My guess is, in this line https://github.com/epoberezkin/ajv/blob/e88e5f53d71e11f94898fb9f7fe8bd0b0bd43692/lib/compile/index.js#L74

must be `this`, `var result = validate.apply(this, arguments);`

cvlab:
Fix #768 

`.apply(null, arguments)` -> `.apply(this, arguments)` in 2 `callValidate` functions to pass custom context.
I'm not sure which are all cases when `callValidate`s are triggered. I added 2 tests. First triggers `callValidate` from `/lib/ajv.js` and second from `/lib/compile/index.js` and in both this tests $ref is recursive.

I made changes in `ref.jst`. My goal is, now and in future, `callValidate` always to be called with `this` as `undefined` when `passContext` is `false`. (Always `callValidate()` but not `someObj.callValidate()`)

Straight ahead way is to change 
```javascript
{{## def._validateRef:_v:
  {{? it.opts.passContext }}
    {{=_v}}.call(this,
  {{??}}
    {{=_v}}(
  {{?}}
    {{=$data}}, {{# def.dataPath }}{{# def.passParentData }}, rootData)
#}}
```
to
```javascript
{{## def._validateRef:_v:
  {{? it.opts.passContext }}
    {{=_v}}.call(this,
  {{??}}
    {{=_v}}.call(undefined,
  {{?}}
    {{=$data}}, {{# def.dataPath }}{{# def.passParentData }}, rootData)
#}}
```

but `fn.call(undefined, ...)` is slower than `fn()`.
Faster method that I implemented is to create aditional var
```javascript
var $refCodeFn = {{= $refCode }};
...
$refCodeFn();
```
in case of `$refCode` = `fn` or `$refCode` = `someObj.fn` always will be passed `this` as `undefined`, but not `someObj`.

On the other hand `{{=_v}}.call(undefined,` is more readable and easier to understand.

Thank You!

### DIFF
--- a/lib/ajv.js
+++ b/lib/ajv.js
@@ -363,9 +363,11 @@ function _compile(schemaObj, root) {
   return v;
 
 
+  /* @this   {*} - custom context, see passContext option */
   function callValidate() {
+    /* jshint validthis: true */
     var _validate = schemaObj.validate;
-    var result = _validate.apply(null, arguments);
+    var result = _validate.apply(this, arguments);
     callValidate.errors = _validate.errors;
     return result;
   }
--- a/lib/compile/index.js
+++ b/lib/compile/index.js
@@ -69,9 +69,11 @@ function compile(schema, root, localRefs, baseId) {
     endCompiling.call(this, schema, root, baseId);
   }
 
+  /* @this   {*} - custom context, see passContext option */
   function callValidate() {
+    /* jshint validthis: true */
     var validate = compilation.validate;
-    var result = validate.apply(null, arguments);
+    var result = validate.apply(this, arguments);
     callValidate.errors = validate.errors;
     return result;
   }
--- a/spec/issues.spec.js
+++ b/spec/issues.spec.js
@@ -703,3 +703,113 @@ describe('property __proto__ should be removed with removeAdditional option, iss
     Object.keys(data.obj) .should.eql(['a', 'b']);
   });
 });
+
+
+describe('issue #768, fix passContext in recursive $ref', function() {
+  var ajv, contexts;
+
+  beforeEach(function() {
+    contexts = [];
+  });
+
+  describe('passContext = true', function() {
+    it('should pass this value as context to custom keyword validation function', function() {
+      var validate = getValidate(true);
+      var self = {};
+      validate.call(self, { bar: 'a', baz: { bar: 'b' } });
+      contexts .should.have.length(2);
+      contexts.forEach(function(ctx) {
+        ctx .should.equal(self);
+      });
+    });
+  });
+
+  describe('passContext = false', function() {
+    it('should pass ajv instance as context to custom keyword validation function', function() {
+      var validate = getValidate(false);
+      validate({ bar: 'a', baz: { bar: 'b' } });
+      contexts .should.have.length(2);
+      contexts.forEach(function(ctx) {
+        ctx .should.equal(ajv);
+      });
+    });
+  });
+
+  describe('ref is fragment and passContext = true', function() {
+    it('should pass this value as context to custom keyword validation function', function() {
+      var validate = getValidateFragments(true);
+      var self = {};
+      validate.call(self, { baz: { corge: 'a', quux: { baz: { corge: 'b' } } } });
+      contexts .should.have.length(2);
+      contexts.forEach(function(ctx) {
+        ctx .should.equal(self);
+      });
+    });
+  });
+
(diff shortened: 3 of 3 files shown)


lib/dot/ref.jst
```diff
@@ -58,8 +58,9 @@
 
 {{? $refCode }}
   {{# def.beginDefOut}}
-    {{# def._validateRef:$refCode }}
+    {{# def._validateRef }}
   {{# def.storeDefOut:__callValidate }}
+  var $refCodeFn = {{= $refCode }};
```
epoberezkin: I don't understand these changes. It looks like `$refCodeFn` is assigned after it is used? Probably not - $refCodeFn is a validation time var, the convention is to use $ for compile time variables. In any case, I don't understand what this re-assignment achieves?

lib/dot/ref.jst
```diff
@@ -2,11 +2,11 @@
 {{# def.errors }}
 {{# def.setupKeyword }}
 
-{{## def._validateRef:_v:
+{{## def._validateRef:
   {{? it.opts.passContext }}
-    {{=_v}}.call(this,
+    $refCodeFn.call(this,
```
epoberezkin: If the idea was to simplify why can't it be `{{= $refCode }}.call(this,` ...

lib/dot/ref.jst
```diff
@@ -2,11 +2,11 @@
...
   {{??}}
-    {{=_v}}(
+    $refCodeFn(
```
epoberezkin: ... and `{{= $refCode }}(` ?

cvlab:
@epoberezkin Changes in `ref.jst` are optional. Idea is that in this code 
```javascript
{{## def._validateRef:_v:
  {{? it.opts.passContext }}
    {{=_v}}.call(this,
  {{??}}
    {{=_v}}(
  {{?}}
    {{=$data}}, {{# def.dataPath }}{{# def.passParentData }}, rootData)
#}}
```
I want to make, that if `passContext` is `false` (`{{=_v}}(` part) to send `undefined` as `this`. This is optional, because, I think, called function anyway do not use `this` var, but it will be good that `this` will be `undefined` when `passContext` is `false`.

Compiled sync function looks like this
```javascript
if (!refVal1(data.resources, (dataPath || '') + '.resources', data, 'resources', rootData)) {
...
}
```
or
```javascript
if (!root.refVal[0](data[i0], (dataPath || '') + '[' + i0 + ']', data, i0, rootData)) {
...
}
```
in second case, to function will be sent `root.refVal` as `this`, but it will be good to be `undefined`

Changes in `ref.jst` copile sync function in this way

```javascript
var $refCodeFn = root.refVal[0];
if (!$refCodeFn(data[i0], (dataPath || '') + '[' + i0 + ']', data, i0, rootData)) {
...
}
```
and now function is called with `undefined` `this`. Anyway this is optional.

epoberezkin:
> I want to make, that if passContext is false ({{=_v}}( part) to send undefined as this

Even though it changes the compiled code, I believe that this change doesn't change the validation behaviour and results. I may be wrong - could you make the test that would fail without this change? I only mean the changes in ref.jst, another thing I understand.

cvlab:
I checked all functions that are called from this line, none are using `this` var when `passContext ` is `false`.
I changed PR, now it's without ref.jst. Thank You!

PR #799, opened 2018-06-05
Update uri-js

dobladov:
By updating *uri-js* we fix the a bug related to IE11 and webpack 4 discussed here garycourt/uri-js#29

Change the version of *uri-js* in the package.json 


epoberezkin:
"^4.2.1" means any version higher than 4.2.1 with the same major version, so no need to change it explicitly.

PR #810, opened 2018-06-21
support resolving internal refs from macro keywords

jstewmon:
As described in epoberezkin/ajv-merge-patch/issues/24, macro keywords cannot resolve internal refs when the schema lacks id.

I added an instanceof check during fragment resolution to find internal refs during compilation (when schema is not a SchemaObject).

Not that I know of... I added a test case to demonstrate what did not work previously.

### DIFF
--- a/lib/compile/resolve.js
+++ b/lib/compile/resolve.js
@@ -70,7 +70,7 @@ function resolveSchema(root, ref) {
   var p = URI.parse(ref)
     , refPath = _getFullPath(p)
     , baseId = getFullPath(this._getId(root.schema));
-  if (refPath !== baseId) {
+  if (Object.keys(root.schema).length === 0 || refPath !== baseId) {
     var id = normalizeId(refPath);
     var refVal = this._refs[id];
     if (typeof refVal == 'string') {
--- a/spec/custom.spec.js
+++ b/spec/custom.spec.js
@@ -209,6 +209,44 @@ describe('Custom keywords', function () {
       testMultipleRangeKeyword({ type: 'number', macro: macroRange }, 2);
     });
 
+    it('should support resolving $ref without id or $id', function () {
+      instances.forEach(function (_ajv) {
+        _ajv.addKeyword('macroRef', {
+          macro: function (schema, parentSchema, it) {
+            it.baseId .should.equal('#');
+            var ref = schema.$ref;
+            var validate = _ajv.getSchema(ref);
+            if (validate) return validate.schema;
+            throw new ajv.constructor.MissingRefError(it.baseId, ref);
+          },
+          metaSchema: {
+            "type": "object",
+            "required": [ "$ref" ],
+            "additionalProperties": false,
+            "properties": {
+              "$ref": {
+                "type": "string"
+              }
+            }
+          }
+        });
+        var schema = {
+          "macroRef": {
+            "$ref": "#/definitions/schema"
+          },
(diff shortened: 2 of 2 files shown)


epoberezkin:
@jstewmon can you please explain how you determined that exactly this change is necessary? The reason I am asking that it changes the execution logic in many test cases, not just in the one you've added. I think there should be somewhat narrower condition, this seems too wide... I didn't have enough time to think about it I should admit, but I've traced the number of extra times the if branch is entered into during the test...

jstewmon:
I'm not sure that this is the exact change necessary :-)

I used a debugger with `ajv-merge-patch` to find that the difference between omitting/providing an `$id` is that the block is entered when an `$id` is given b/c `refPath` is always `#` for an internal ref and `baseId` will be `#` when `$id` is omitted.

I also observed that `root.schema` is empty when `resolveSchema` is eventually called by `getSchema` in the macro.

My first attempt at resolving the issue was to check whether `baseId === '#' || refPath !== baseId`, but that resulted in some failing tests.

So, I debugged some additional tests and observed that `root` may be an instance of `SchemaObject`. That led me to try this fix, which passed all current tests as well as the case where `getSchema` is called in a macro with an internal ref.

I'm not very familiar with the internals, so I'm open to investigating further. Do you have any ideas that I could leverage to look for an alternate fix?

epoberezkin:
I understand. Looks like I will have to debug myself, I am not quite sure yet what is going on here :)

epoberezkin:
> So, I debugged some additional tests and observed that root may be an instance of SchemaObject. That led me to try this fix, which passed all current tests as well as the case where getSchema is called in a macro with an internal ref.

That may be a correct fix, even though it comes at the cost of going inside this branch when it is not necessary... I just want to make sure I understand why and also that we're not creating any side effects here.

Thank you, I will try to get to it quickly.

epoberezkin:
If you want to investigate further empirically, I would look into the properties of root rather whether it is or isn't an instance of SchemaObject, I think it may be a coincidence that the latter worked, and there is some more specific condition that is common for failing test...

jstewmon:
Changing to `Object.keys(root.schema).length === 0 || refPath !== baseId` also fixes the test and only increases the block entrances from `241` to `249` in the fast test suite. The instanceof check raises the block entrances to `4397`.

I'll update the PR to reflect this, and we can go from there.

epoberezkin:
excellent, makes my job easier, thank you :)

jstewmon:
Also, if you think this is an appropriate fix but want to optimize the empty check, I can add an `objectIsEmpty` method to `util` as follows:

```js
function objectIsEmpty(o) {
  for (var key in o) {
    if (o.hasOwnProperty(key)) {
      return false;
    }
  }
  return true;
}
```

That may help preserve Ajv's edge in the benchmarks for larger schemas.

epoberezkin:
I wouldn't worry about compilation performance, validation time isn't affected anyway, so no need to optimise it.

jstewmon:
I went back and reran the block entrance tracing with the new test skipped and found that the number of entrances is unaffected by the addition of the `Object.keys(root.schema).length === 0` check. The increase from `241` to `249` is the result of the `compile` call succeeding in the test with the patch.

epoberezkin:
Thank you

jstewmon:
You’re welcome! Thanks for your attention and for the excellent library! 😃

PR #846, opened 2018-08-23
Fix types

qm3ster:
Mostly just replaced `Boolean` constructor with `boolean` type


forivall:
While working to understand the ajv code, I wrote a lot more type definitions along these lines: https://github.com/forivall/ajv/tree/ts-jsdoc

epoberezkin:
does not apply to v7

PR #860, opened 2018-09-19
Replaced phantomjs dependency with headless chrome by creating custom-launcher

ossdev07:

Replaced phantomjs browser with Headless chrome.
Added custom launcher for HeadlessChrome in karma.conf.js and update karma-chrome-launcher(2.2.0) version in package.json
**Result of test cases with chrome:**
HeadlessChrome 0.0.0 (Ubuntu 0.0.0): Executed 2379 of 2409 (skipped 30) SUCCESS (1 min 51.542 secs / 1 min 45.92 secs)


ossdev07:
I have updated with chrome as phantomjs browser is currently unmaintained. In aarch64 platform, there is no specific phantomjs binary. We need to build it from source, which takes more than 1 hour. Replacing with Headless chrome gives faster output too. Please revert if any queries. Thanks.

PR #867, opened 2018-10-07
Always ignore $comment alongside $ref

### LINKED ISSUE #686: Ignore $comment when used together with $ref
When `$ref` is used, no other keyword shall be used alongside it. That is enforced with the *extendRefs* option set to `"ignore"` or `"fail"`.

However, `$comment`s should be ignored by implementations. They are useful to give other developers additional knowledge about a property without affecting the schema.

Is it possible to change the behavior of the *extendRefs* option so that it will still warn / fail if other keywords except `$comment` are present alongside `$ref` but allow `$comment`?

---

**Example:** Here, the `$ref`s are rather used as a special type to avoid code duplication and the comments adds additional information.

```json
  ...
  "nonEmptySingleLineString": {
    "$id": "nonEmptySingleLineString",
    "type": "string",
    "pattern": "^[^\n]+$"
  },
  ...
  "name": {
    "$comment": "unique in file",
    "$ref": "nonEmptySingleLineString"
  },
  "shortName": {
    "$comment": "unique globally; if not set: name is used",
    "$ref": "nonEmptySingleLineString"
  },
  ...
```

FloEdelmann:

Closes #686.

A new boolean option `allowRefComments`. If it is set to `false` (default), everything works as before. If it is set to true, `$comment` is allowed alongside `$ref` in schemas (and does not cause a log message) when `extendRefs` is set to "ignore" or "fail".

I don't think so.


lib/compile/util.js
```diff
@@ -174,12 +174,22 @@ function finalCleanUpCode(out, async) {
 function schemaHasRules(schema, rules) {
   if (typeof schema == 'boolean') return !schema;
   for (var key in schema) if (rules[key]) return true;
+  return false;
```
epoberezkin: not necessary

README.md
```diff
@@ -1038,6 +1038,7 @@ Defaults:
   schemaId:         '$id',
   missingRefs:      true,
   extendRefs:       'ignore', // recommended 'fail'
+  allowRefComments: false,
```
epoberezkin: I prefer simply always ignoring $comment with $ref, without any extra option.

lib/compile/util.js
```diff
@@ -174,12 +174,22 @@ function finalCleanUpCode(out, async) {
...
+    return !schema && (exceptKeywords.length > 1 || exceptKeywords[0] != 'not');
+
+  for (var key in schema) {
+    var keywordExcluded = false;
+    for (var i = 0; i < exceptKeywords.length; i++)
+      if (key === exceptKeywords[i]) keywordExcluded = true;
+
+    if (!keywordExcluded && rules[key]) return true;
+  }
+  return false;
```
epoberezkin: this change will be much simpler as $comment can be just used here directly

lib/dot/validate.jst
```diff
@@ -16,7 +16,7 @@
 
 {{
   var $async = it.schema.$async === true
-    , $refKeywords = it.util.schemaHasRulesExcept(it.schema, it.RULES.all, '$ref')
+    , $refKeywords = it.util.schemaHasRulesExcept(it.schema, it.RULES.all, (it.opts.allowRefComments ? ['$ref', '$comment'] : ['$ref']))
```
epoberezkin: without option - no change here

epoberezkin (changes requested):
Thank you. I suggest dropping an extra option - it will make the change much simpler.

FloEdelmann:
Removing the option leads to other failing tests:

```
  1) Ajv Options
       $comment
         = true
           should log the text from $comment keyword:

      AssertionError: expected [] to deeply equal [ [ 'property foo' ] ]
      + expected - actual

      -[]
      +[
      +  [
      +    "property foo"
      +  ]
      +]
      
      at test (spec/options.spec.js:1300:30)
      at /home/flo/www/ajv/spec/options.spec.js:1293:11
      at Array.forEach (<anonymous>)
      at Context.<anonymous> (spec/options.spec.js:1289:24)

  2) Ajv Options
       $comment
         function hook
           should pass the text from $comment keyword to the hook:

      AssertionError: expected [] to deeply equal [ Array(1) ]
      + expected - actual

      -[]
      +[
      +  [
      +    "property foo"
      +    "#/properties/foo/$comment"
      +    {
      +      "properties": {
      +        "bar": {
      +          "$comment": "property bar"
      +          "type": "integer"
      +        }
      +        "foo": {
      +          "$comment": "property foo"
      +        }
      +      }
      +    }
      +  ]
      +]
      
      at test (spec/options.spec.js:1341:31)
      at /home/flo/www/ajv/spec/options.spec.js:1330:11
      at Array.forEach (<anonymous>)
      at Context.<anonymous> (spec/options.spec.js:1326:24)
```

epoberezkin:
Well, these tests should pass - they specifically test that the content of $comment itself should be logged when this option $comment is enabled - it is not related to the problem this PR should be addressing.

teoxoy:
Are the tests passing now?

epoberezkin:
Tests pass but the changes are not updated I think.

FloEdelmann:
Sorry this took so long :see_no_evil:

I've updated the code, but the same errors remain. I've short-circuited the precommit hook to show you my current version. Do you have any suggestions what could interfere with the `$comment` option working correctly?

epoberezkin:
No problem with the time - thanks for the contribution.

The change should allow comments and refs in the same schema object, if I understand it correctly. I don’t see how it should affect any of the existing tests (unless you change them).

The additional tests you need should tests that with the comments and ref together Ajv doesn’t throw but it still logs comments (or calls hook).

What am I missing?

FloEdelmann:
I had one more look and sorted out what was still missing :tada: 
So I think this is now ready to review @epoberezkin :)

FloEdelmann:
Any updates about this @epoberezkin?

FloEdelmann:
Still no news about this PR?

lsirivong:
👋 hello! I just ran into this. Any update on when this might be able to ship?

epoberezkin:
this is no longer an issue in v7, closing

Issue #568, opened 2017-09-20
Ajv.ValidationError type is not defined for TypeScript


sundriver:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**
5.2.2, Yes

```javascript

if ( err instanceof Ajv.ValidationError ) { ... }

```
Error I receive: 

Property 'ValidationError' does not exist on type '{ (options?: Options | undefined): Ajv; new (options?: Options | undefined): Ajv; }'.

epoberezkin:
There is also Ajv.MissingRefError

dinvlad:
For some reason, it still reports `Cannot use 'new' with an expression whose type lacks a call or construct signature.` for `ValidationError` when I try `new ValidationError(ajv.errors)`

yume-chan:
@epoberezkin as you declared https://github.com/epoberezkin/ajv/blob/28386786fd9f6229652829a673745a1563e097c7/lib/ajv.d.ts#L1-L7

You are declaring `ValidationError` field as **an instance of** the `ValidationError` class.

You can even do 

``` ts
import ajv from 'ajv';
console.log(ajv.ValidationError.ajv); // the docs say it's `true`, but it's actually `undefined`.
```

----

The correct way should be

``` ts
ValidationError: typeof ValidationError;
```

This allows you to do `new ajv.ValidationError`.

----

But, it still doesn't allow you to declare a variable as `ajv.ValidationError` type, because the type itself is now exported.

``` ts
const error: ajv.ValidationError; // nope
```

To resolve this, I found this strange way:

``` ts
declare namespace Errors {
  class ValidationError extends Error { /* ... */ }

  class MissingRefError extends Error { /* ... */ }
}

declare var ajv: {
  (options?: ajv.Options): ajv.Ajv;
  new(options?: ajv.Options): ajv.Ajv;
  ValidationError: typeof Errors.ValidationError;
  MissingRefError: typeof Errors.MissingRefError;
  $dataMetaSchema: object;
}

declare namespace ajv {
  type ValidationError = Errors.ValidationError;

  type MissingRefError = Errors.MissingRefError;

  /* ... */
}
```

----

Now you can write:

``` ts
import ajv from 'ajv';

const validationError: ajv.ValidationError = new ajv.ValidationError([]);
const missingRefError: ajv.MissingRefError = new ajv.MissingRefError("", "", "");
```

PR #868, opened 2018-10-07
fix: update typescript typings definition file

acheronfail:

There are some issues with the TypeScript typings, especially around the exported classes. Users have had issues with it as well as described in this comment: https://github.com/epoberezkin/ajv/issues/568#issuecomment-405754868

Basically, this issue is that if you try and check if an `error` is an instance of `Ajv.ValidationError` it won't work.

Some steps to reproduce:

```ts
try {
  const validate = ajv.compile(schema); // schema with $async
  await validate(data);
} catch (err) {
  if (!(err instanceof Ajv.ValidationError)) { // <-- TS ERROR HERE
    throw err;
  }
}
```

You get this TypeScript error:

```txt
[ts] The right-hand side of an 'instanceof' expression must be of type 'any' or of a type assignable to the 'Function' interface type.
```

This is because the types don't correctly export the error classes.

This PR updates the TS typings definitions so that this error is resolved.
When making this change I followed the guides found at: https://github.com/typings/typings/blob/master/docs/examples.md#function--utility which are provided by the TypeScript authors.

**Basically, the changes are:**

* export `declare function ajv(options?: ajv.Options): ajv.Ajv;`
* add `export` to all interfaces and types
* move the error classes into the main namespace and export them

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -1,13 +1,9 @@
-declare var ajv: {
-  (options?: ajv.Options): ajv.Ajv;
-  new (options?: ajv.Options): ajv.Ajv;
-  ValidationError: ValidationError;
-  MissingRefError: MissingRefError;
-  $dataMetaSchema: object;
-}
+declare function ajv(options?: ajv.Options): ajv.Ajv;
 
 declare namespace ajv {
-  interface Ajv {
+  export type $dataMetaSchema = object;
+
+  export interface Ajv {
     /**
     * Validate data using schema
     * Schema will be compiled and cached (using serialized JSON as key, [fast-json-stable-stringify](https://github.com/epoberezkin/fast-json-stable-stringify) is used to serialize by default).
@@ -107,7 +103,7 @@ declare namespace ajv {
     errors?: Array<ErrorObject>;
   }
 
-  interface ValidateFunction {
+  export interface ValidateFunction {
     (
       data: any,
       dataPath?: string,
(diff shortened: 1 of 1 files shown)


lib/ajv.d.ts
```diff
@@ -1,13 +1,9 @@
-declare var ajv: {
-  (options?: ajv.Options): ajv.Ajv;
-  new (options?: ajv.Options): ajv.Ajv;
-  ValidationError: ValidationError;
-  MissingRefError: MissingRefError;
-  $dataMetaSchema: object;
-}
+declare function ajv(options?: ajv.Options): ajv.Ajv;
```
epoberezkin, line 1: I think that the usage as constructor may have been lost? Also the issue is to export only ErrorParameters interfaces to use them as type guards - why all interfaces should be exported?

epoberezkin (changes requested):

lib/ajv.d.ts
acheronfail: > I think that the usage as constructor may have been lost?

The `declare function ajv(options?: ajv.Options): ajv.Ajv;` and `export = ajv` in the file ensure that the constructor type isn't lost.

> Also the issue is to export only ErrorParameters interfaces to use them as type guards - why all interfaces should be exported?

Unfortunately, the only way I found to combine the `declare function ajv(...)` type with the `namespace ajv` type was to export all the types inside. Otherwise the types themselves aren't exported as well. 

I tried various ways of fixing the issue, but the only way that worked and didn't break anything was following the typings guide https://github.com/typings/typings/blob/master/docs/examples.md#function--utility here, which says to declare `function` and `namespace` types (where all the types you need from the namespace are exported).

epoberezkin:
replaced by #903

PR #865, opened 2018-10-03
Adding logger to typescript Options declaration.

jsdevel:
**What issue does this pull request resolve?**
Logger wasn't in the `Ajv.Options` declaration, causing my code to break when moving to typescript.

**What changes did you make?**
Updated `lib/ajv.d.ts`

**Is there anything that requires more attention while reviewing?**
Is there a specific interface for the added `CustomLogger` declaration?

PR #872, opened 2018-10-19
Add logger type

sabrehagen:

Missing `logger` option.

Added type definition for logger.

No.


epoberezkin:
Adds the same as #865 but definition is narrow.

PR #891, opened 2018-11-29
Add Typewriter to README

colinking:

Typewriter uses AJV to perform validation in our Javascript-based clients. This PR adds it to the list of packages utilizing AJV. 

https://github.com/segmentio/typewriter

For example: https://github.com/segmentio/typewriter/blob/master/examples/gen-js/js/analytics/generated/index.js


epoberezkin:
Sorry, npm provides a better list of packages that depend on ajv. Only de-facto standard libraries or ajv plugins need to be added to the docs.

PR #899, opened 2018-12-11
fix(issue#897): Wrong results with {removeAdditional: 'failing', allErrors: false} and additionalProperties as failing schema

### LINKED ISSUE #897: Wrong results with {removeAdditional: 'failing', allErrors: false} and additionalProperties as failing schema
I test it with latest AJV V6.6.1. Yes, it is.

**Ajv options object**

```
{
  "removeAdditional": "failing",
  "allErrors": false
}
```

**JSON Schema**

```
{
  "type": "object",
  "properties": {
    "requiredProperty": {
      "type": "boolean"
    }
  },
  "required": [
    "requiredProperty"
  ],
  "additionalProperties": {
    "not": {}
  }
}
```

**Sample data**

```
{
  "foo": 0
}
```

**Your code**

```
var Ajv = require('ajv');
var ajv = new Ajv(options); 
var valid = validate(data); //true - but must be false
console.log(validate.errors); // validate.errors !== null
console.log(data); // {} - foo is removed
```

**Errors**
```
[ 
  {
     keyword: 'not',
     dataPath: '[\'foo\']',
     schemaPath: '#/additionalProperties/not',
     params: {},
     message: 'should NOT be valid' 
  } 
]
```

I expect:
1. `valid === false` (-)
2. `foo` removed (+)
3.  `validate.errors` not null and does not contain error about additionalProperty `foo`, because `foo` is removed from data. (-)

I test it with latest AJV V6.6.1.
I write test and check all other combinations of `removeAdditional` and `allErrors` options and `additionalProperties` schema value. Here is a code (with AJV v4.11.5) http://plnkr.co/edit/ZIPWAn5HB2Ele4YgGVil?p=info
It seems like a bug, or I don't understand behavior.

Faris-Nasybulin:
Fixes [897](https://github.com/epoberezkin/ajv/issues/897)

1. When additional property is removed, then `nextValid` is reset to `true`.
2. Used common `vErrors` variable instead of `validate.errors`.
3. Added in documentation, that is not guaranteed removing additional properties for invalid data with `allErrors: false`.
4. Added some tests.

I work with the source code of the project for the first time, so I'm not sure that I understood everything. Please check everything carefully, maybe add tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1112,7 +1112,7 @@ Defaults:
   - `false` (default) - not to remove additional properties
   - `"all"` - all additional properties are removed, regardless of `additionalProperties` keyword in schema (and no validation is made for them).
   - `true` - only additional properties with `additionalProperties` keyword equal to `false` are removed.
-  - `"failing"` - additional properties that fail schema validation will be removed (where `additionalProperties` keyword is `false` or schema).
+  - `"failing"` - additional properties that fail schema validation will be removed (where `additionalProperties` keyword is `false` or schema). Note that for invalid data with `allErrors: false` removing additional properties is not guaranteed.
 - _useDefaults_: replace missing properties and items with the values from corresponding `default` keywords. Default behaviour is to ignore `default` keywords. This option is not used if schema is added with `addMetaSchema` method. See examples in [Assigning defaults](#assigning-defaults). Option values:
   - `false` (default) - do not use defaults
   - `true` - insert defaults by value (safer and slower, object literal is used).
--- a/lib/dot/properties.jst
+++ b/lib/dot/properties.jst
@@ -107,11 +107,12 @@ var {{=$nextValid}} = true;
 
           if (!{{=$nextValid}}) {
             errors = {{=$errs}};
-            if (validate.errors !== null) {
-              if (errors) validate.errors.length = errors;
-              else validate.errors = null;
+            if (vErrors !== null) {
+              if (errors) vErrors.length = errors;
+              else vErrors = null;
             }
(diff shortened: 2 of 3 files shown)


epoberezkin:
@farisn thank you for the change and sorry it took long time to review.

The problem with reviewing it is that there are 4 different scenarios with 2 options (allErrors = true/false) that are inside a single test, it would be easier to understand if they were split.

Some of these 8 scenarios pass without the change, some require the change (I didn't check all). The lack of descriptive names for these tests also complicates reviewing.

Also this assertion: https://github.com/epoberezkin/ajv/pull/899/files#diff-fe04da3b92b1e5d8609c211a4b069473R193 is too complex, but it will get simpler if the tests are split...

Would you be able to split it to 8+ different tests in 3 separate commits:
- commit 1: 8 tests that pass without the change (they are there to ensure correct behavior remains and skipped tests that fail without the change.
- commit 2: change and un-skipping the tests that now pass
- commit 3: additional skipped tests that should pass, but don't pass because, as you say, with allErrors: false it doesn't work - I would try to fix this scenario, I don't like the idea of the option that occasionally doesn't work - it is better than it it is now, but it doesn't resolve the issue.

Please let me know if you could restructure it in this way.

Thank you.

epoberezkin:
closing it - it needs to be revisited.

Faris-Nasybulin:
1. Is this PR is relevant and I can change it, or will you refactor project and it will become irrelevant?
2. What code style should I follow? Each test should be in a separated `it` call or can I keep current style?

PR #903, opened 2018-12-13
Properly fix type definition for ValidationError and MissingRefError (#568)

yume-chan:

#568

https://github.com/epoberezkin/ajv/issues/568#issuecomment-446890570

replaces #868

Added a generic argument for `validate` and `ValidateFunction`, and some formatting changes.

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -1,12 +1,36 @@
 declare var ajv: {
   (options?: ajv.Options): ajv.Ajv;
-  new (options?: ajv.Options): ajv.Ajv;
-  ValidationError: ValidationError;
-  MissingRefError: MissingRefError;
+  new(options?: ajv.Options): ajv.Ajv;
+  ValidationError: typeof AjvErrors.ValidationError;
+  MissingRefError: typeof AjvErrors.MissingRefError;
   $dataMetaSchema: object;
 }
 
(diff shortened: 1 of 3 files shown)


vogre (approved):
Good stuff, this change compiles correctly when tested in our project!

epoberezkin:
I removed generic argument:
1. Promise result type is different from data type
2. Function should be able to validate any data type, it's not limited to a specific data type

yume-chan:
@epoberezkin 

> Promise result type is different from data type

The README reads:

> Validation result will be a promise that resolves with validated data or rejects with an exception `Ajv.ValidationError` that contains the array of validation errors in `errors` property.

Doesn't it mean the Promise result will be the input data itself?

----

The FAQ reads:

> Ajv also supports asynchronous validation (with custom asynchronous formats and keywords) that returns a promise that either resolves to `true` or rejects with an error.

Is it wrong?

-----

> Function should be able to validate any data type, it's not limited to a specific data type

Writing `validate<T>(data: T): Promise<T>` doesn't require the user to specify `T`, TypeScript will infer `T` from the input `data`. See my tests:

https://github.com/epoberezkin/ajv/blob/95edb4958dd3a93b5dd2179aa71e517a386c7d12/spec/typescript/index.ts#L18-L32

PR #916, opened 2019-01-03
useDefaults: "empty"

### LINKED ISSUE #912: New option to enhance default functionality
6.6.1

When submitting HTML forms, the url-encoded data will send empty strings for fields that are empty.  The joi library has functionality to default empty strings, and I would like to see this same functionality implemented in ajv.

> However, if you want to specify a default value in case of empty string you have to use a different pattern: Joi.string().empty('').default('default value').

Maybe a new enum for the useDefaults option like "emptyValues".  The default logic needs to be modified to be executed for all properties regardless of the property's absence or current value if a default annotation is present in the schema for that property.  The default logic needs to mutate the data if the property is missing or has an "emptyValue" before being validated against the property format.

Probably not, but I'd love to learn.

*Question*
Is it worth also enhancing defaults for null, empty array [], empty object {}, zero 0?

*Examples*
*Defaulting an empty string to a numeric value.*
```
{
   Product: "10mm Wood Screw",
   Quantity: ""
}
//Schema has default: 0 for Quantity
{
   Product: "10mm Wood Screw",
   Quantity: 0
}
```

*Defaulting a zero value to a different value*
```
{
   Model: "2018 Tesla Model S",
   BatteryCapacity: 0
}
//Default in schema is 85 kilowatt hours
{
   Model: "2018 Tesla"
   BatteryCapacity: 85
}
```

epoberezkin:
closes #912

### DIFF
--- a/README.md
+++ b/README.md
@@ -699,13 +699,11 @@ The schema above is also more efficient - it will compile into a faster function
 
 With [option `useDefaults`](#options) Ajv will assign values from `default` keyword in the schemas of `properties` and `items` (when it is the array of schemas) to the missing properties and items.
 
+With the option value `"empty"` properties and items equal to `null` or `""` (empty string) will be considered missing and assigned defaults.
+
 This option modifies original data.
 
-__Please note__: by default the default value is inserted in the generated validation code as a literal (starting from v4.0), so the value inserted in the data will be the deep clone of the default in the schema.
-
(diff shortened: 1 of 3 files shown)


B-Teague:
Thank you for adding this functionality!  Will this automatically work for dynamic defaults as well in the ajv-keywords repo?
Does npm automatically update, or do you submit a new version?

epoberezkin:
released

Issue #941, opened 2019-02-09
NPM: 6.9.0 AJV Commit causing failure to execute node_modules (multiple dependencies affected i.e. ESLint and WebPack Loaders)


caldempsey:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**

Latest version (6.9.0 release I believe)

**Ajv options object**
```
node_modules/ajv/lib/keyword.js:65
      throw new Error('custom keyword definition is invalid: '  + this.errorsText(validateDefinition.errors));
      ^

Error: custom keyword definition is invalid: data.errors should be boolean
    at Ajv.addKeyword (/home/mmacheerpuppy/git/fsa/react-application/node_modules/ajv/lib/keyword.js:65:13)
    at module.exports (/home/mmacheerpuppy/git/fsa/react-application/node_modules/ajv-errors/index.js:10:7)
    at Object.<anonymous> (/home/mmacheerpuppy/git/fsa/react-application/node_modules/schema-utils/src/validateOptions.js:22:1)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
```

**What results did you expect?**

Expected to be able to build node_modules but experiencing above stacktrace.

**Are you going to resolve the issue?**
Current workaround is to remove ESLint dependencies for NPM users, or if using YARN to resolve dependencies to forced version. It's also possible (with the addition of _another_ dependency) to use NPM resolutions (see conversation).

**Reproduction**

Try using the following packages declaration. npm post install will fail to start. ESLint relies on AJV.

```
{
  "name": "food-standards",
  "version": "0.1.0",
  "private": true,
  "dependencies": {
    
    "axios": "^0.18.0",
    "bootstrap": "^4.2.1",
    "react": "^16.8.1",
    "react-dom": "^16.8.1",
    "react-scripts": "2.1.3",
    "reactstrap": "^7.1.0"
  },
  "scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test",
    "eject": "react-scripts eject"
  },
  "eslintConfig": {
    "extends": "react-app"
  },
  "browserslist": [
    ">0.2%",
    "not dead",
    "not ie <= 11",
    "not op_mini all"
  ],
[...]

PR #943, opened 2019-02-10
fix: log instead of throw when definition validation fails

fkleuver:

https://github.com/epoberezkin/ajv/issues/941

When definition validation fails in keyword.js, log the error instead of throwing it so that it doesn't break dependent packages that were passing invalid definitions to it, but still notifies users in some manner.

As this is a breaking change, I believe the throwing should be coupled with a major version bump, rather than a minor one.

I am not sure if there may be other ways in which the latest changes could be breaking dependent packages, but this is the part that surfaced.

To make the tests pass again, I stubbed the error logger to throw so that this change can be rolled out again with a major version bump with minimal effort to fix the tests.

### DIFF
--- a/lib/keyword.js
+++ b/lib/keyword.js
@@ -62,7 +62,7 @@ function addKeyword(keyword, definition) {
     validateDefinition = validateDefinition || this.compile(definitionSchema);
 
     if (!validateDefinition(definition))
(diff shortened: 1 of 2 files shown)


epoberezkin:
It actually was throwing before, it just was a bit less restrictive... I've released the fix - let's see if there is anybody still affected.

fkleuver:
Alright, I'll close this then :)

Issue #950, opened 2019-02-18
Request to make console optional for default construction


shamasis:
**The version of Ajv you are using**
latest

**The environment you have the problem with**
NodeJS vm.js

**Your code (please make it as small as possible to reproduce the issue)**
```
var Ajv = require('ajv'),
    ajv = new Ajv(); // note {logger: console} not provided
```

**Issue**
Ajv assumes console to be present in the environment. It is an acceptable for all normal use cases. But we use AJV within Node VM module and that is sandboxed to not have console.

The following line https://github.com/epoberezkin/ajv/blob/master/lib/ajv.js#L488 assumes console to be present. This results in issues such as https://github.com/postmanlabs/postman-app-support/issues/3199 and requires to have a longer boilerplate code.

If it is okay, we can send a PR that checks if console is absent even in global scope and injects a stub.

epoberezkin:
I am not convinced that making Ajv resilient to the absence of global console is the right approach, given that
1) 2 main JS environments have it (and probably most main environments have it)
2) there are simple solutions that don’t require library changes:
  a) pass console object to Ajv instance - there is an option
  b) define global console if for whatever reason you cannot pass it (in case you don’t control instance options).

Why does it have to fixed in the library?

epoberezkin:
Having thought about it, I believe that even for the environments that do not have "console" object, I believe it is better to fail, to make users aware that the library does not log in cases it is supposed to, rather than to do nothing silently.

Once users understand that either the global console or "logger" option is required, they will have a choice to either supply another way to log or to supply a logger that does nothing.

shamasis:
@epoberezkin - I am with you on the thought process. Maybe as a work around this could be a module level setting. Say...
```
var Ajv = require('ajv');
Ajv.debug({
    console: specialConsoleFunctionForAllSubsequentConstruction
});
```

AJV is now getting distributed to 7+ million users via Postman Sandbox and I really want to make it easier (not quirky) for them. If you think the above idea is good then someone from our team would be glad to send a PR.

epoberezkin:
@shamasis - how do they log? Why there is no console object?

So you suggest to provide a constructor api to provide default options for future instances?

If so, something like Ajv.options() probably makes more sense. Need to think about it.

igorpupkinable:
@epoberezkin logger setter would help a lot since logger instance is constructed on every invocation in some cases/environments.
[...]

PR #951, opened 2019-02-18
Enable running in console less environments

shamasis:
Ajv assumes console to be present in the environment. It is an acceptable for all normal use cases. But we use AJV within Node VM module and that is sandboxed to not have console.

The following line https://github.com/epoberezkin/ajv/blob/master/lib/ajv.js#L488 assumes console to be present. This results in issues such as [postmanlabs/postman-app-support#3199](https://github.com/postmanlabs/postman-app-support/issues/3199) and requires to have a longer boilerplate code.

As reported in: https://github.com/epoberezkin/ajv/issues/950

Inserted a stub console if global console is absent

- Is it okay to inset a stub console or is it wise to check for console before every use? I prefer the former.

### DIFF
--- a/lib/ajv.js
+++ b/lib/ajv.js
@@ -485,7 +485,14 @@ function setLogger(self) {
   if (logger === false) {
     self.logger = {log: noop, warn: noop, error: noop};
   } else {
-    if (logger === undefined) logger = console;
+    if (logger === undefined) {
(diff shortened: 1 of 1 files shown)


epoberezkin:
Thank you for the PR, but it will not be merged. Please see the comment in the issue.

PR #960, opened 2019-03-01
feat: invalidDefaults option to warn when defaults are ignored, fixes #957

### LINKED ISSUE #957: Allow easy detection of ignored `default` keywords in schemas

6.9.1

When using ajv with `useDefaults: true`, the `default` keyword is ignored in some cases, as documented in the readme. If someone is aware of the existence of the `default` keyword but isn't familiar with the details of where it can be used, they might inadvertently use it in a place where it's ignored, resulting in a buggy schema. (This recently caused some bugs in ESLint rules -- also see https://github.com/eslint/eslint/pull/11427.)

We'd like to detect this problem so that schema authors can fix the issue. As far as I can tell, there isn't a simple way for an external tool to figure out whether an ignored `default` keyword exists in a user-provided schema, without reimplementing some of the logic from `ajv`.

It should be possible to opt into a mode where invalid `default` keywords result in a compilation or validation error, rather than being ignored. It seems like this could in principle be added to the meta-schema, although it could make the meta-schema substantially larger. Another possibility could be to add an option to the `Ajv` constructor that causes this case to throw an error.

I'm willing to give it a shot if there is agreement that this is worthwhile.

not-an-aardvark:

#957

This adds an `invalidDefaults` option to implement the suggestion from https://github.com/epoberezkin/ajv/issues/957.

(I chose the name `invalidDefaults` rather than `ignoredDefaults` because the defaults are no longer actually ignored, depending on the value of the option.)

I'm not very familiar with how the `dotjs` templates work, so it might be worth double-checking that I did that correctly.


epoberezkin (commented):
Thank you. Not convinced they should be called “invalid” :) They are valid from JSON Schema point of view, it’s just that Ajv ignores them as there is no easy way to apply them. Will review and merge the code over weekend - ok?

not-an-aardvark:
Sure, no problem. Is there a name you would recommend instead? (Maybe something like `unusedDefaults`?)

epoberezkin:
I think ignoredDefaults was fine, there are two options that allow value “ignore”. I am guessing here - maybe you think it looks too similar to “ignoreDefaults” and can be confusing? We also could “simplify” the whole thing and cram it into useDefaults option - “warn” would log unused defaults and “strict” would throw? Not sure :) 

unusedDefaults is cool too - consistent with useDefaults

epoberezkin:
unusedDefaults it is then :)

not-an-aardvark:
I think my confusion with `ignoredDefaults` is that while the defaults were previously being ignored before adding this option, they are now sometimes logging a warning or reporting an error, which intuitively doesn't seem like they're being ignored

I'm certainly happy to change it to `ignoredDefaults` if you'd prefer -- I'm just mentioning this in case other people would be similarly confused.

epoberezkin:
> I think my confusion with `ignoredDefaults` is that while the defaults were previously being ignored before adding this option, they are now sometimes logging a warning or reporting an error, which intuitively doesn't seem like they're being ignored

I see... so we need some option that would mean to throw if some defaults are not used.

Maybe strictDefaults: true/'log'? The meaning is to only allow defaults that are used, and report other defaults?

From reading the issue you want to throw:

> It should be possible to opt into a mode where invalid `default` keywords result in a compilation or validation error, rather than being ignored.

If so, we could do useDefaults: true/'strict' and leave any possible further extension to the future. Logging is not that useful normally...

Anyway, I am probably thinking too much about it :) Happy with either of the above: strictDefaults or useDefaults extension - your call.

epoberezkin:
Going to merge it with strictDefaults - I want to release it today

not-an-aardvark:
Thanks!

Issue #980, opened 2019-03-29
anyOf validation errors only present when other validations fail


mseeley:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**

6.10.0

**Ajv options object**

```javascript
undefined
```

**JSON Schema**

The schema intends to define a recursive structure. Valid data examples below.

```json
const schema = {
  "$schema": "http://json-schema.org/draft-07/schema",
  "$ref": "#/definitions/stages",
  "definitions": {
    "stages": {
      "type": "array",
      "minItems": 1,
      "items": {
        "anyOf": [
          { "$ref": "#/definitions/stages" },
          { "$ref": "#/definitions/stage" }
        ]
      }
    },
    "stage": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string",
          "minLength": 1
        }
      }
    }
  }
};
```

**Sample data**

Invalid

```json
[{ "name": "" }]
```

Valid (flat)

```
[{ "name": "A" }]
```

Valid (nested)

The JSON is a composite structure. It's an array of object or other arrays containing objects.

```
[
  { "name": "A" },
  [
    { "name": "B" },
    { "name": "C" }
  ]
]
```

**Your code**

Demo: https://runkit.com/mseeley/ajv

```javascript
// See demo
```

**Validation result, data AFTER validation, error messages**

Validation fails due to the empty `name` attribute. Data is unchanged.

```
[
  {
    "keyword": "type",
    "dataPath": "[0]",
    "schemaPath": "#/type",
    "params": {
      "type": "array"
    },
    "message": "should be array"
  },
  {
    "keyword": "minLength",
    "dataPath": "[0].name",
    "schemaPath": "#/definitions/stage/properties/name/minLength",
    "params": {
      "limit": 1
    },
    "message": "should NOT be shorter than 1 characters"
  },
  {
    "keyword": "anyOf",
    "dataPath": "[0]",
    "schemaPath": "#/items/anyOf",
    "params": {},
    "message": "should match some schema in anyOf"
  }
]
```

**What results did you expect?**

I'm expecting an `errors` object containing only the failed `minLength` failure as changing `name` to a string with length > 0 causes the test to pass. This to me suggests the other two errors are not truly errors. I can't filter out `anyOf` errors as an array containing anything besides other array or object should fail `anyOf`.

I _believe_ the non minLength errors are related to the anyOf check where the first member of the array could have been another array instead of an object.

**Are you going to resolve the issue?**

[...]

Issue #828, opened 2018-08-02
Reduced error messages returned from oneOf/anyOf


stoked10:
I'll start off with great tool! I've been converting all of our schema validation from another, much slower, package. The "problem" I have is that we use the error messages returned to display the error to our application's front end. The "issue" I have is that oneOf/anyOf combined with if/then returns all of the failed errors making it impossible to determine why it failed. 

I'm currently using 6.5.2

The follow example is hand typed because I'm on a disconnected network, so please excuse any typos:
```json
{
  "type": "array",
  "additionalProperties": false,
  "items": [
    {
      "type": "object",
      "additionalProperties": false,
      "properties": {
        "type": {
          "type": "string",
          "enum": ["circle","triangle","polygon"]
        },
        "points": {
          "type": "array",
          "items": points,
          "oneOf": [
            {
              "if": {"properties": {"type": {"enum": ["circle"]}}},
              "then": { "minItems": 1, "maxItems": 1 }
            },
            {
              "if": {"properties": {"type": {"enum": ["triangle"]}}},
              "then": {"minItems": 3,"maxItems": 3}
            },
            {
              "if": {"properties": {"type": {"enum": ["polygon"]}}},
              "then": { "minItems": 3 }
            }
          ]
        }
      }
    }
  ]
}
```
points are defined as a latitude and longitude. 
if I pass in json like this:
```json
[
  {
    "type": "triangle",
    "points": [{"lat":"1N", "long":"2E"}, {"lat":"2N","long":"3E"}],
  },
]
```
I get a lot of errors back:

```json
[
{ "path": "0.points", "type": "maxItems", "message: 'should NOT have more than 1 items" }
{ "path": "0.points", "type": "if", "message": "should match then schema" }
{ "path": "0.points", "type": "minItems", "message": "should NOT have less than 3 items" }
{ "path": "0.points", "type": "if", "message": "should match then schema" }
{ "path": "0.points", "type": "minItems", "message": "should NOT have less than 3 items" }
{ "path": "0.points", "type": "if", "message": "should match then schema" }
]
```
What I'd really like to get back is just:
```json
{ "path": "0.points", "type": "minItems", "message": "should NOT have less than 3 items" }
{ "path": "0.points", "type": "if", "message": "should match then schema" }
```

Is it possible to add a flag for this? Or is this something I missed that already exits?

epoberezkin:
[...]

Issue #201, opened 2016-05-28
how to reduce validation errors to chosen path of `anyOf`


mbroadst:
relevant code example:

```
let PointSchema = {
  id: 'point',
  anyOf: [
    {
      type: 'object',
      properties: {
        latitude: { type: 'number' },
        longitude: { type: 'number' }
      },
      required: [ 'latitude', 'longitude' ]
    },
    {
      type: 'array',
      description: 'GeoJSON format',
      items: [ { type: 'number' }, { type: 'number' } ],
      additionalItems: false,
      minItems: 2
    }
  ]
};

const Ajv = require('ajv');
let ajv = Ajv();
ajv.addSchema(PointSchema, 'point');
let validate = ajv.compile(PersonSchema);
let valid = validate({
  name: 'BadBob',
  location: [ 200 ]
});

console.log(validate.errors);
```

results in:

```
[ { keyword: 'type',
    dataPath: '.location',
    schemaPath: 'point/anyOf/0/type',
    params: { type: 'object' },
    message: 'should be object' },
  { keyword: 'minItems',
    dataPath: '.location',
    schemaPath: 'point/anyOf/1/minItems',
    params: { limit: 2 },
    message: 'should NOT have less than 2 items' },
  { keyword: 'anyOf',
    dataPath: '.location',
    schemaPath: 'point/anyOf',
    params: {},
    message: 'should match some schema in anyOf' } ]
```

In this case, the first path of `anyOf` should not have been followed (as the input is not an object but instead an array). Unfortunately, I get back a sort of confusing collection of errors for the user which I'll now have to go back and parse my way through manually it seems. 

Is there a way to reduce the output errors to only the path followed which best matches the input?

mbroadst:
hm, I thought the solution might be to switch to `oneOf` (instead of `anyOf`, so we're only reporting errors for the path chosen), but that results in the following somewhat useless error:

```
testing:  { name: 'BadBob', location: [ 200 ] }
valid:  false
errors:  [ { keyword: 'oneOf',
    dataPath: '.location',
    schemaPath: 'point/oneOf',
    params: {},
    message: 'should match exactly one schema in oneOf' } ]
```

epoberezkin:
anyOf collects one error for each subschema and if no valid subschema is found it also adds anyOf error. Determining which of failing subschemas "best matches" the input is not trivial - it seems to be out of scope of validation and more like in application scope.

[...]

PR #1012, opened 2019-05-23
feat: group anyOf errors under each branch

ForbesLindesay:

It looks like there are a number of issues where people are complaining about how difficult it is to tell what the real errors are when a schema fails to match an `anyOf` constraint. I think the problem is that you can get two conflicting errors for two different branches, and there's nothing to indicate which conditions are "or" and which are "and".

I think this would go a long way towards resolving issues reported in #201 #980 #828 

My solution is a structure like:

```js
[ { keyword: 'anyOf',
    dataPath: '',
    schemaPath: '#/anyOf',
    params:
     { errors:
        [ [ { keyword: 'type',
              dataPath: '',
              schemaPath: '#/anyOf/0/type',
              params: { type: 'number' },
              message: 'should be number',
              schema: 'number',
              parentSchema: { type: 'number' },
              data: 'foo' } ],
          [ { keyword: 'type',
              dataPath: '',
              schemaPath: '#/anyOf/1/type',
              params: { type: 'integer' },
              message: 'should be integer',
              schema: 'integer',
              parentSchema: { type: 'integer' },
              data: 'foo' } ] ] },
    message: 'should match some schema in anyOf',
    schema: [ { type: 'number' }, { type: 'integer' } ],
    parentSchema: { anyOf: [ { type: 'number' }, { type: 'integer' } ] },
    data: 'foo' } ]
```

which indicates that one branch of the `anyOf` had the error "should be number" and the other branch had the error "should be integer".

This fails some of the tests in json-schema-test about recursive schemas, I'm not 100% sure how I should go about fixing those.

BREAKING CHANGE: the error structure has changed for `anyOf` errors.

### DIFF
--- a/lib/dot/anyOf.jst
+++ b/lib/dot/anyOf.jst
@@ -15,6 +15,8 @@
 
   {{# def.setCompositeRule }}
 
+  var anyOf_vErrors_{{=$dataLvl}} = vErrors;
+  var anyOf_errors_{{=$dataLvl}} = null;
   {{~ $schema:$sch:$i }}
     {{
       $it.schema = $sch;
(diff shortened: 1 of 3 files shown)


epoberezkin:
1. As it's a breaking change it could only be introduced with an option, e.g.: groupErrors
2. I would like to look at the proposed errors format in draft-08 JSON-Schema spec (not published yet) - it makes sense to just use it.
3. If this feature is introduced it needs to be done for anyOf, oneOf, and allOf together.

As a workaround it's relatively easy to attribute an error to a branch by looking at schemaPath, so it can be done in the application code by parsing it.

Some of the tests fail because they assert the specific number of errors - they need to be reviewed.

epoberezkin (changes requested):
1. As it's a breaking change it could only be introduced with an option, e.g.: groupErrors
2. I would like to look at the proposed errors format in draft-08 JSON-Schema spec (not published yet) - it makes sense to just use it.
3. If this feature is introduced it needs to be done for anyOf, oneOf, and allOf together.

As a workaround it's relatively easy to attribute an error to a branch by looking at schemaPath, so it can be done in the application code by parsing it.

Some of the tests fail because they assert the specific number of errors - they need to be reviewed.

epoberezkin:
closing...

Issue #993, opened 2019-04-14
OpenApi attribute nullable with enum


left-rite:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**
6.10.0

**Ajv options object**
```javascript
{
  nullable: true
}

```

**JSON Schema**

```json
{
  "type": "object",
  "properties": {
    "propertyOne": { 
      "type": "string",
      "nullable": true,
      "enum": [ "foo", "bar" ]
    }
  }
}

```

**Sample data**

```json
{
  "propertyOne": null
}

```

**Your code**

```javascript
var ajv = new Ajv(options);
var validate = ajv.compile(schema);
var result = validate(data);

```

**Validation result, data AFTER validation, error messages**

```json
[
  {
    "keyword": "enum",
    "dataPath": ".propertyOne",
    "schemaPath": "#/properties/propertyOne/enum",
    "params": {
      "allowedValues": [
        "foo",
        "bar"
      ]
    },
    "message": "should be equal to one of the allowed values"
  }
]

```

**What results did you expect?**
The result to be true, since support for the keyword nullable has be set to true and schema has set nullable to true for the property "propertyOne".

**Are you going to resolve the issue?**
Not sure

epoberezkin:
Interesting. I am not sure that this should be the expected behaviour. It is definitely possible to add null into the enum explicitly.

Could you point to the OpenAPI doc that says that nullable should change the behaviour of enum?

left-rite:
I couldn't find a doc or example so I asked [OAI](https://github.com/OAI/OpenAPI-Specification/issues/1900), essentially the response was validation constraints like enum should only apply when the value is not null.

mattpolzin:
I've got a "fix" for this, if @epoberezkin is ok with calling the current behavior undesirable.

For me the problem with the current behavior is that I cannot add `null` to the `enum` without metaschema validation failing (the enum should only contain the type specified for the property and because we are using `nullable` rather than adding `null` to the array of types `null` is not allowed in the `enum`).

The following fails as mentioned by the OP.
```
{
  "type": "object",
  "properties": {
    "propertyOne": { 
      "type": "string",
      "nullable": true,
      "enum": [ "foo", "bar" ]
    }
  }
}
```

The next variation fails metaschema validation for the reason I mention above.
```
{
  "type": "object",
  "properties": {
    "propertyOne": { 
      "type": "string",
      "nullable": true,
      "enum": [ "foo", "bar", null ]
    }
  }
[...]

PR #1021, opened 2019-06-13
Nullable enum

mattpolzin:
https://github.com/epoberezkin/ajv/issues/993

I moved the check for whether the `nullable` option is `true` **AND** the `nullable` property is `true` into a definition and then used that definition in the existing validation step that adds `null` to the array of types allowed as well as in the `enum` keyword step to add `null` to any enumerated list of values that might be present on a `nullable` property.

I'm very unfamiliar with doT so please let me know if I've created any code that does not meet standards or bets practices.

### DIFF
--- a/lib/dot/definitions.def
+++ b/lib/dot/definitions.def
@@ -130,6 +130,13 @@
 #}}
 
 
+{{## def.$nullable:
+  {{
+    var $isNullable = it.opts.nullable && it.schema.nullable === true;
+  }}
+#}}
+
+
 {{## def.$dataNotType:_type:
   {{?$isData}} ({{=$schemaValue}} !== undefined && typeof {{=$schemaValue}} != _type) || {{?}}
 #}}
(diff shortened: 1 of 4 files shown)


epoberezkin:
Thanks for the PR.

I am afraid it cannot be addressed in such a narrow way. See https://github.com/OAI/OpenAPI-Specification/issues/1900#issuecomment-486772917

Firstly, OpenAPI needs to clarify whether interpretation 1 or 2 in that comment is correct - it would require a major version change. Today's OpenAPI docs are closer to interpretation 2, in which case this PR is simply wrong. 2 is also consistent with the proposal in https://github.com/OAI/OpenAPI-Specification/issues/1389

If OpenAPI decides that interpretation 1 is correct and updates the spec, then ALL schema keywords should be skipped not just "enum". Among standard keywords only "const" and "enum" are affected, but there may be other custom keywords and they should also be skipped.

Either way, before any change is made here, OpenPAI position should be clarified and reflected in the docs.

mattpolzin:
Indeed. I agree with your decision after reading that whole comment thread. I think it’s unfortunate that something can be “nullable” _and_ disallow actually having a null value by omitting it from the enum, but that’s an OpenAPI topic.

PR #1044, opened 2019-07-17
add: `omitDataPath` option to `ajv.errorsText` for ajv-errors custom …

nswbmw:
```js
var Ajv = require('ajv')
var ajv = new Ajv({ allErrors: true })
require('ajv-errors')(ajv)

var validate = ajv.compile({
  type: 'object',
  properties: {
    name: {
      type: 'string',
      minLength: 6,
      errorMessage: 'Your name should more than 6 characters'
    }
  }
})

validate({ name: 'xxx' })

console.log(ajv.errorsText(validate.errors))
// data/name Your name should more than 6 characters
```

But i want get raw errorMessage.

```js
console.log(ajv.errorsText(validate.errors, { omitDataPath: true }))
// Your name should more than 6 characters
```

### DIFF
--- a/lib/ajv.js
+++ b/lib/ajv.js
@@ -404,7 +404,7 @@ function _get$IdOrId(schema) {
  * Convert array of error message objects to string
  * @this   Ajv
  * @param  {Array<Object>} errors optional array of validation errors, if not passed errors from the instance are used.
- * @param  {Object} options optional options with properties `separator` and `dataVar`.
+ * @param  {Object} options optional options with properties `separator`, `dataVar` and `omitDataPath`.
  * @return {String} human readable string with all errors descriptions
  */
 function errorsText(errors, options) {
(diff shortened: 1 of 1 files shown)


epoberezkin:
Thank you for the PR. I do not understand the problem that it solves though - what is the use case?

nswbmw:
I don't want to return dataPath(eg: `data/name`) to user.

```js
console.log(ajv.errorsText(validate.errors))
// data/name Your name should more than 6 characters
```

to

```js
console.log(ajv.errorsText(validate.errors, { omitDataPath: true }))
// Your name should more than 6 characters
```

epoberezkin:
In this case the best approach is not to use errorsText method and process errors in your code to format the output for the users.

errorsText is a simple convenience method to simplify logging all errors in one line - it’s target audience is developers, not users.

epoberezkin (changes requested):
Actually, I think this option can be ok, but I suggest `dataPath: true/false` defaulting to true (instead of `omitDataPath` defaulting to false).

To be merged, it needs to have:

- [ ] tests (it looks like this method has no tests at all - it can go in this file: https://github.com/epoberezkin/ajv/blob/master/spec/ajv.spec.js in one define)
  - [ ] the test of the current behaviour when this option is not present
  - [ ] the test of the current behaviour when this option is true
  - [ ] the test of the new behaviour when this option is false
- [ ] typescript type
- [ ] updated docs

nswbmw:
@epoberezkin Thanks for your proposal, i was busy with my business, sorry for not have the time to complete this pr. closed for any other people could make a new pr.

PR #1055, opened 2019-08-05
Update ajv.d.ts

carlos121493:
optional support first argument


epoberezkin:
I think only the last argument(s) can be optional.

PR #1075, opened 2019-08-27
Updated readme with proposed ReDoS solution

Venryx:
The solution being to run the format validation in a separate thread, with a timeout specified. If it doesn't complete within the timeout period, the data is assumed invalid, and the thread killed.

Granted, this breaks the synchronous nature of AJV validation, and is therefore outside of the library's own validation stack. But it might still be worth noting as a library-external solution to the problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -683,6 +683,7 @@ __Please note__: some formats that Ajv implements use [regular expressions](http
 - making assessment of "format" implementations in Ajv.
 - using `format: 'fast'` option that simplifies some of the regular expressions (although it does not guarantee that they are safe).
 - replacing format implementations provided by Ajv with your own implementations of "format" keyword that either uses different regular expressions or another approach to format validation. Please see [addFormat](#api-addformat) method.
+- running the format validation in a separate thread, with a timeout specified.
 - disabling format validation by ignoring "format" keyword with option `format: false`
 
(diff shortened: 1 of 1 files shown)


epoberezkin:
Do you mean making the whole schema asynchronous (`$async: true` in the schema) and defining custom formats that use another processes to run? If so, it needs clarifying...

Venryx:
Oh, I wasn't aware AJV supported asynchronous validation. That would make it more convenient to add other-thread regex evaluation. (since then it doesn't require a user-created "wrapper/proxy function" for `ajv.validate`)

My suggestion is not very fleshed out. Evaluating regular expressions in another thread is just an approach to solving ReDoS attacks that I've heard used before, so I figured it was worth mentioning somewhere.

epoberezkin:
So the suggestion could be to replace format/pattern with async custom keywords, e.g. asyncFormat and asyncPattern, that would run it in another thread to be non-locking.

Or, maybe better, it can be a plugin that would make standard keywords format and pattern asynchronous provided the schema is async.

So if you or somebody implements such plugin (see plugins section in docs), it can be referenced here.

epoberezkin:
alternatively, the suggestion could be to run the whole validation in a separate thread (that I think can also be achieved with async schema).

PR #1105, opened 2019-10-19
readme: add link to ajv-content-encoding plugin

remko:

Adds a link to ajv-content-encoding-plugin in the README.

Add a link to ajv-content-encoding-plugin in the README.

No.


epoberezkin (changes requested):
Thank you! Happy to reference it as BETA and partial implementation of contentEncoding, listing which values are supported and not supported atm. It can be updated once full support is added, also BETA can be removed as build test and code coverage is added to the repo. Is it ok?

epoberezkin:
readme restructured

Issue #1136, opened 2019-12-18
Add keywords option


franciscomorais:
Add the `keywords` option to the Ajv options to add custom keywords, without having to call `addKeyword`.

**What version of Ajv you are you using?**
6.10.2

**What problem do you want to solve?**
Add keywords object directly to the Ajv constructor, like formats.

```js
new Ajv({
  keywords: {
    constant: {
      validate: function (schema, data) {
        return true;
      }
    }
  }
});
```

**What do you think is the correct solution to problem?**
Copy the implementation of the `formats` option to add keywords.

**Will you be able to implement it?**
yes

epoberezkin:
merged

PR #1137, opened 2019-12-18
Add keywords to Ajv options

franciscomorais:
This PR adds the `keywords` option to the Ajv options to add custom keywords, without having to call `addKeyword`.

#1136

Add the keywords option to the Ajv options to add custom keywords, without having to call addKeyword.

No

### DIFF
--- a/spec/options/options_validation.spec.js
+++ b/spec/options/options_validation.spec.js
@@ -26,9 +26,32 @@ describe('validation options', function() {
       }});
 
       var validate = ajv.compile({ format: 'identifier' });
+
       validate('Abc1') .should.equal(true);
       validate('123') .should.equal(false);
       validate(123) .should.equal(true);
+    });
+  });
+
+  describe('keywords', function() {
+    it('should add keywords from options', function() {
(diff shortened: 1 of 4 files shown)


epoberezkin (changes requested):
At the moment it adds formats, not keywords. The implementation, tests and types need to be updated.

spec/options/options_validation.spec.js
```diff
@@ -26,12 +26,30 @@ describe('validation options', function() {
...
       validate('Abc1') .should.equal(true);
       validate('123') .should.equal(false);
       validate(123) .should.equal(true);
     });
   });
 
+  describe('keywords', function() {
+    it('should add keywords from options', function() {
+      var ajv = new Ajv({ keywords: {
+        identifier: {
```
epoberezkin: I suggest adding type: "string" in the keyword definition and changing the implementation to `/^[a-z_$][a-z0-9_$]*$/i.test(data)` so the example makes sense. Then the same tests can be used as in formats option test.

epoberezkin (changes requested):

PR #1159, opened 2020-02-04
Add custom logger example

Tyharo1:
The current value description for the custom logger option is not easily understandable. Adding an example to this would be greatly beneficial.

The readability and clarity of using a custom logger when using the README.

Added a Javascript example of what a custom logger would like for those looking to use it.

N/A


README.md
```diff
@@ -1151,6 +1151,15 @@ Defaults:
 - _schemas_: an array or object of schemas that will be added to the instance. In case you pass the array the schemas must have IDs in them. When the object is passed the method `addSchema(value, key)` will be called for each schema in this object.
 - _logger_: sets the logging method. Default is the global `console` object that should have methods `log`, `warn` and `error`. Option values:
   - custom logger - it should have methods `log`, `warn` and `error`. If any of these methods is missing an exception will be thrown.
+    ```javascript
+    var ajv = new AJV.default({
```
epoberezkin: Not consistent with other examples, it should be `new Ajv(...`

README.md
```diff
@@ -1151,6 +1151,15 @@ Defaults:
...
+      logger: {
+        log: log => console.log(log),
```
epoberezkin: All (or most) examples in the spec atm are ES5. While it’s worth revisiting, it should stay consistent, so let’s stick to ES5 please. You can create another issue to discuss moving examples to ES6 (with the comment the Ajv code is still ES5 for wider compatibility).

README.md
```diff
@@ -1151,6 +1151,15 @@ Defaults:
...
+        warn: () => null,
```
epoberezkin: warn method is mandatory as docs say.

epoberezkin (commented):
I’ve added the comments, but in general this is a contrived example that simply creates the same behaviour as default, so it is pointless and confusing.

I suggest instead move into errors section, creating a separate subsection Error logging linked from here, and put some explanations there how any logging package can be used as long as it or its wrapper expose the required methods. Does it make sense?

Tyharo1:
@epoberezkin Thank you for the feedback, its greatly appreciated! I believe I've made the desired changes to the example. I'll take some time and add the Error subsection you suggested 👍

Tyharo1:
@epoberezkin The discussed changes have been implemented. Please let me know if I properly addressed the desired changes.

PR #1162, opened 2020-02-11
fix(types) added undefined to getSchema()

boenrobot:

getSchema() may return undefined if the key does not resolve to a previously added schema. Users should be encouraged to check for that. Since there isn't a separate method to check if a keyRef would resolve (without getting and compiling the schema), checking the return value seems like the best way.

Added undefined as a possible return type of getSchema(). While one could check the return value even without this change, linters would complain about it, as an undefined check would always be false according to the types.


epoberezkin:
Thank you!

PR #1166, opened 2020-02-19
feat: Make the validation error more detailed

nassiharel:
This feature will improve the validation message returned from function: .errorsText

*example message*

before change:
data.env should be equal to one of the allowed values

after change:
data.env should be equal to one of the allowed values (nodejs,python,jvm)

Add an optional boolean property to decide if message should be more specific.

```javascript

validator.errorsText(errors, { extraInfo: true });

```

Yes, add tests

### DIFF
--- a/lib/ajv.js
+++ b/lib/ajv.js
@@ -413,11 +413,20 @@ function errorsText(errors, options) {
   options = options || {};
   var separator = options.separator === undefined ? ', ' : options.separator;
   var dataVar = options.dataVar === undefined ? 'data' : options.dataVar;
+  var extraInfo = options.extraInfo === true;
 
   var text = '';
   for (var i=0; i<errors.length; i++) {
     var e = errors[i];
-    if (e) text += dataVar + e.dataPath + ' ' + e.message + separator;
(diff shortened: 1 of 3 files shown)


lib/ajv.js
```diff
@@ -413,11 +413,21 @@ function errorsText(errors, options) {
...
   var text = '';
+  var extraText = '';
   for (var i=0; i<errors.length; i++) {
     var e = errors[i];
-    if (e) text += dataVar + e.dataPath + ' ' + e.message + separator;
+    if (e) {
+      if(extraInfo) {
+        if (e.params && e.params.allowedValues)
+          extraText += ' (' + e.params.allowedValues.join(',') + ')';
+        else if (e.params && e.params.additionalProperty)
```
epoberezkin, line 426: there was a suggestion to log additional property previously, but given that it's data submitted from external source it is not secure. It is covered in https://github.com/epoberezkin/ajv/blob/master/FAQ.md#why-dont-additionalproperties-false-errors-display-the-property-name. I am going to close PR, sorry.

PR #1176, opened 2020-03-21
patternProperties: skip keys with undefined values

### LINKED ISSUE #1173: schema for objects: patternProperties and properties treat `undefined` values differently

Version: `6.12.0`

**Ajv options object**

- None, this is using the default `options`.

**JSON Schema**

```js
{
  type: "object",
  properties: {
    normalProp: { type: "string" }
  },
  patternProperties: {
    patternProp: { type: "string" }
  }
}
```

**Sample data**

```js
{
  normalProp: undefined,
  patternProp: undefined
}
```

**Your code**

```javascript
var ajv = new Ajv({});

console.log(validate(data));
console.log(validate.errors);
```

https://runkit.com/yellowkirby/ajv-issue-pattern-properties

```js
// validate.errors
[{
  dataPath: "['patternProp']"
  keyword: "type"
  message: "should be string"
  params: Object {type: "string"}
  schemaPath: "#/patternProperties/patternProp/type"
}]
```

I expected no validation errors. The sub-schema for `normalProp` and `patternProp` are identical, so they should both treat the literal `undefined` as a non-existent value (https://github.com/epoberezkin/ajv/issues/207#issuecomment-309576041)

I'm hoping somebody more familiar with the codebase has a simple fix 🙏

YellowKirby:

Fixes #1173 

When iterating over object keys that match patternProperties, ignore keys whose data values that are `undefined`. This makes it consistent with how regular properties are handled.

In `properties.jst`, when iterating over properties in an object, check both that the property key matches the given RegExp (this check is not new) *and* that the data value is not `undefined` (this part is new).

First time making changes to this codebase. Not sure if there's anything I missed.

### DIFF
--- a/lib/dot/properties.jst
+++ b/lib/dot/properties.jst
@@ -215,7 +215,7 @@ var {{=$nextValid}} = true;
       }}
 
       {{# def.iterateProperties }}
-        if ({{= it.usePattern($pProperty) }}.test({{=$key}})) {
+        if ({{=$data}}[{{=$key}}] !== undefined && {{= it.usePattern($pProperty) }}.test({{=$key}})) {
           {{
             $it.errorPath = it.util.getPathExpr(it.errorPath, $key, it.opts.jsonPointers);
             var $passData = $data + '[' + $key + ']';
--- a/spec/issues/1173_patternProperties_with_undefined_value.spec.js
+++ b/spec/issues/1173_patternProperties_with_undefined_value.spec.js
@@ -1,0 +1,26 @@
+'use strict';
+
+var Ajv = require('../ajv');
+require('../chai').should();
+
+
+describe('issue #1173: properties matched by patternProperties with `undefined` values', function() {
+  it('should ignore `undefined` values for properties matched by patternProperties ', function() {
+    var schema = {
+      type: "object",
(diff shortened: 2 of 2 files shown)


epoberezkin:
@YellowKirby thanks for the PR and sorry for the delay.

I am a bit reluctant to fix it without major version change and even more reluctant doing it piece by piece.

There are quite a few cases where `undefined` properties would be used, depending on `ownProperties` option - check out other places where `iterateProperties` is used and also `noPropertyInData` - so probably easier to fix there. Both are defined in definitions.def. Also there is `required` keyword - not sure what is the current behaviour, it would depend on some other options...

To do it comprehensively, in the current major version, some option (e.g. `validateUndefined`) should be added:
- without option: keep current behaviour for all keywords, accounting for `ownProperties` option.
- false: ignore undefined properties for all keywords
- true: take undefined properties into account for all keywords

It will result in a much bigger test case.

YellowKirby:
Thank you for the feedback! I think adding a new option makes a lot of sense. I can work on adding that as well as checking more edge cases/combinations with other options when I get some free time.

epoberezkin:
closing it, the code moved to TS

PR #1196, opened 2020-04-25
Pass schema object to processCode function

### LINKED ISSUE #1194: Pass schema to processCode

6.12.0

I am using the `processCode` option to instrument my generated code and it would be useful for my instrumentation to have access to information from my schemas.

I think the schema should be passed as a second parameter to the `processCode` function.

Yes.

awlayton:

This resolves #1194

I passed the schema to `processCode` when it is callled,
as well as updating the type definitions and README for `processCode`.

Nope.


README.md
```diff
@@ -1169,7 +1169,7 @@ Defaults:
   errorDataPath:    'object', // deprecated
   messages:         true,
   sourceCode:       false,
-  processCode:      undefined, // function (str: string): string {}
+  processCode:      undefined, // function (str: string, schema: object): string {}
```
epoberezkin, line 1172: Please also add comment in the lines 1281-2 below. `require('js-beautify').js_beautify` probably won't work anymore after this change, as it probably has some optional second parameter (it needs to be checked), so it has to be mentioned both in the docs and in the release notes what it has to be replaced with - not sure how many users would be affected... I am thinking to release over the next weekend to see if anything breaks. Alternatively it can be released on the `beta` branch in preparation for the next major version. I will think about it, but please add to the docs in any case.

epoberezkin (changes requested):
thank you - please see the comment re docs, all good otherwise.

epoberezkin:
thank you - I will release it on Saturday

PR #1216, opened 2020-05-13
[ajv.d.ts] Add missing argument for validateSchema function

brussee:
Fix types according to implementation and documentation: https://github.com/ajv-validator/ajv/blob/master/lib/ajv.js#L164

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -74,9 +74,10 @@ declare namespace ajv {
     /**
     * Validate schema
     * @param {object|Boolean} schema schema to validate
+    * @param {Boolean} throwOrLogError pass true to throw (or log) an error if invalid
     * @return {Boolean} true if schema is valid
     */
-    validateSchema(schema: object | boolean): boolean;
+    validateSchema(schema: object | boolean, throwOrLogError?: boolean): boolean;
     /**
     * Get compiled schema from the instance by `key` or `ref`.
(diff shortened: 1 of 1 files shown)


brussee:
For people using [patch-package](https://github.com/ds300/patch-package) you may find the following gist useful until this PR is merged: https://gist.github.com/brussee/0011732e20dbcfd4c4e9d33748fa23f4

epoberezkin:
This wasn't documented in readme, let me please think if I want to legitimise this parameter - it was quite ad-hoc. Doc in code was needed mainly to appease linter about "this" (I think).

Why do you need this parameter becoming public api? Its name is quite confusing tbh...

brussee:
@epoberezkin Thanks for your reply and sorry for the delay.

My use case is that I want to know the reason if schema is invalid. The only way to get that information back to the caller is to have the `validateSchema` function throw an error with the message (is there another way?), which it only does right now if you set the `throwOrLogError` parameter to `true`.

I agree the name is confusing and I propose you choose between what I consider the C way vs. the C++ way of returning (status/exception) information from a function, e.g.:
`func foo(in, *err) { err = 'foo'; return -1; }`
`func foo(in) { throw Error('foo'); }`
If you want to support both, like you do now, maybe you can make a global setting instead of a parameter per function, because consistency is preferred in my opinion.

Thanks for your consideration!

brussee:
I just noticed the errors and errorsText are public API already, so I can use that instead.

epoberezkin:
You can indeed get the errors from the instance - thank you.

PR #1219, opened 2020-05-22
Add option for strictNumbers. Resolves #1128.

### LINKED ISSUE #1128: NaN and Infinity should fail type:number with option strictNumbers

https://github.com/epoberezkin/ajv/issues/182#issuecomment-219548557

Option changing type validation. Check if these values are coercible to strings - with this option coercion should fail too

Maybe

issacgerges:
#1128

Added an option for `strictNumbers` which is used in `checkDataType` to conditionally check `isFinite(date)`. Added documentation

The current behavior of `ajv` is that `integer` will always fail `NaN` and pass `Infinity` but `number` currently doesn't. To respect semver, this will continue to behave the same way, and this new option `strictNumbers` is off by default.

I'm unfamiliar with some of the internals of `ajv` and believe that my changes to `checkDataType` are sufficient. Also was unsure why y'all were using `data === data` vs `isFinite(data)` in integer (one will only fail `NaN`, the other will fail `NaN and pos/neg `Infinity`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1156,6 +1156,7 @@ Defaults:
   // strict mode options
   strictDefaults:   false,
   strictKeywords:   false,
+  strictNumbers:    false,
   // asynchronous validation options:
   transpile:        undefined, // requires ajv-async package
   // advanced options:
@@ -1250,7 +1251,9 @@ Defaults:
   - `false` (default) - unknown keywords are not reported
   - `true` - if an unknown keyword is present, throw an error
   - `"log"` - if an unknown keyword is present, log warning
-
+- _strictNumbers_: validate numbers strictly, failing validation for NaN and Infinity. Option values:
+  - `false` (default) - NaN or Infinity will pass validation for numeric types
+  - `true` - NaN or Infinity will not pass validation for numeric types
 
 ##### Asynchronous validation options
 
--- a/lib/compile/util.js
+++ b/lib/compile/util.js
@@ -36,7 +36,7 @@ function copy(o, to) {
 }
 
 
-function checkDataType(dataType, data, negate) {
+function checkDataType(dataType, data, strictNumbers, negate) {
   var EQUAL = negate ? ' !== ' : ' === '
     , AND = negate ? ' || ' : ' && '
     , OK = negate ? '!' : ''
@@ -49,15 +49,18 @@ function checkDataType(dataType, data, negate) {
                           NOT + 'Array.isArray(' + data + '))';
     case 'integer': return '(typeof ' + data + EQUAL + '"number"' + AND +
                            NOT + '(' + data + ' % 1)' +
-                           AND + data + EQUAL + data + ')';
+                           AND + data + EQUAL + data +
+                           (strictNumbers ? (AND + OK + 'isFinite(' + data + ')') : '') + ')';
+    case 'number': return '(typeof ' + data + EQUAL + '"' + dataType + '"' +
+                          (strictNumbers ? (AND + OK + 'isFinite(' + data + ')') : '') + ')';
     default: return 'typeof ' + data + EQUAL + '"' + dataType + '"';
   }
 }
 
 
-function checkDataTypes(dataTypes, data) {
+function checkDataTypes(dataTypes, data, strictNumbers) {
   switch (dataTypes.length) {
-    case 1: return checkDataType(dataTypes[0], data, true);
+    case 1: return checkDataType(dataTypes[0], data, strictNumbers, true);
     default:
       var code = '';
       var types = toHash(dataTypes);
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -1246,11 +1247,13 @@ Defaults:
   - `false` (default) - ignored defaults are not reported
   - `true` - if an ignored default is present, throw an error
   - `"log"` - if an ignored default is present, log warning
-- _strictKeywords_: report unknown keywords in schemas. Option values:
```
issacgerges: This change is unrelated but appeared to be inconsistency in the docs

epoberezkin: sorry, not sure what was the inconsistency?

epoberezkin: "Option values:" is used in most options (if not all)...

lib/compile/util.js
```diff
@@ -50,14 +50,15 @@ function checkDataType(dataType, data, negate) {
     case 'integer': return '(typeof ' + data + EQUAL + '"number"' + AND +
                            NOT + '(' + data + ' % 1)' +
                            AND + data + EQUAL + data + ')';
+    case 'number': return '(typeof ' + data + EQUAL + '"' + dataType + '"' +  (strictNumbers ? (AND + OK + 'isFinite(' + data + ')') : '') + ')';
```
epoberezkin: The reason to use `data === data` is because it's ES5 compatible.
isFinite was only added in ES6 (I think), so it would only be possible if ajv drops ES5 support for generated code...
If I am correct about ES6, then it should be `data === data && data !== Infinity && data !== -Infinity` in the generated code.

spec/options/strictNumbers.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    });
+
+    it('should not impact validation for NaN as type integer', function() {
+      var validate = ajv.compile({type: 'integer'});
+      validate(1).should.equal(true);
+      validate(NaN).should.equal(false);
+      validate(Infinity).should.equal(true);
+    });
+  });
+});
```
epoberezkin: It's very thorough - thanks. I suggest taking `false` and default  test into some function as it's exactly the same test for two option values - so it is clear that the behaviour for `false` and default is the same.

Something like:

```javascript
describe('strictNumbers default', testWithoutStrictNumbers(new Ajv))
describe('strictNumbers = false', testWithoutStrictNumbers(new Ajv({strictNumbers: false})))

function testWithoutStrictNumbers(_ajv) {
  return function () {
  // ...
  }
}
```

epoberezkin (changes requested):

epoberezkin:
@issacgerges thank you - added some comments.

README.md
issacgerges: You're right, my mistake

lib/compile/util.js
issacgerges: It appears as though `isFinite` is available in ES5 (I think it's actually been there since the beginning) although it didn't start hanging off the `Number` prototype since ES6. This code should still work in es5. https://www.ecma-international.org/ecma-262/5.1/#sec-15.1.2.5

issacgerges:
Hey @epoberezkin made the changes you suggested. I ended up leaving `isFinite` and linked to the spec showing it is ES5 compatible. I think you were thinking of `Number.prototype.isFinite` not `isFinite`

issacgerges:
@epoberezkin any other changes you'd like me to make?

lib/compile/util.js
epoberezkin: cool, then it can be used (needs some benchmarking, but https://jsbench.me/ shows them as equal)

spec/options/strictNumbers.spec.js
```diff
@@ -0,0 +1,47 @@
...
+      validate(1).should.equal(true);
+      validate(NaN).should.equal(false);
+      validate(Infinity).should.equal(false);
+    });
+
+    it('should not impact validation for NaN as type integer', function() {
+      var validate = ajv.compile({type: 'integer'});
+      validate(1).should.equal(true);
+      validate(NaN).should.equal(false);
+      validate(Infinity).should.equal(true);
```
epoberezkin: Just noticed it, sorry - I think integer should also not allow Infinity with strictNumbers.

epoberezkin:
> @epoberezkin any other changes you'd like me to make?

made one small comment about integer with strict numbers - sorry, missed it before.

Almost there.

spec/options/strictNumbers.spec.js
epoberezkin: I don't think anybody would want to have a separate option and integers are also numbers (i.e. they would not fail validation with type: 'number'). The idea of type: 'integer' is to impose stricter constraint than number, but in this case it becomes somewhat overlapping rather than stricter.

epoberezkin: And I would also add non-integer examples to the tests, to make sure nothing was broken with the new option (e.g. 1.1 would pass number but fail integer).

issacgerges:
> > @epoberezkin any other changes you'd like me to make?
> 
> made one small comment about integer with strict numbers - sorry, missed it before.
> 
> Almost there.

no worries @epoberezkin, I believe I should have met all your requests this time. let me know!

epoberezkin:
Thank you!

Issue #1228, opened 2020-06-11
Document creating pre-compiled schemas for CSP with ajv-cli


GrahamLea:
**What problem do you want to solve?**
Easily understanding how to use ajv with a Content Security Policy that doesn't include `unsafe-eval`.

**What do you think is the correct solution to problem?**
The README.md should contain a section on how to use `ajv-cli` to create and use pre-compiled schemas. I found how to do this eventually, but only after lots of googling.

**Will you be able to implement it?**
Nope.

epoberezkin:
There is the part about  ajv-cli: https://github.com/ajv-validator/ajv/blob/master/README.md#command-line-interface

The structure definitely needs to be improved - there is a recent ticket about it.

It would be great if you could suggest the amendments based on your experience.

PR #1239, opened 2020-06-30
Document pre-compiled schemas for CSP in README

GrahamLea:
Pre-compiled schemas are a workaround for maintaining a secure Content Security Policy (CSP)

#1228

Added two paragraphs to the README

Unlikely

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,13 +238,14 @@ Ajv is tested with these browsers:
 __Please note__: some frameworks, e.g. Dojo, may redefine global require in such way that is not compatible with CommonJS module format. In such case Ajv bundle has to be loaded before the framework and then you can use global Ajv (see issue [#234](https://github.com/ajv-validator/ajv/issues/234)).
 
 
-### Ajv & Content Security Policies (CSP)
+### Ajv and Content Security Policies (CSP)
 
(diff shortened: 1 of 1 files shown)


GrahamLea:
I've added this under the "Using in browser" section, but it could also fit in the "Security considerations" section.
I didn't add it to the menu at the top as I wasn't sure whether it needed to contain all level 3 headings.

epoberezkin:
Thank you. I think it is important to comment somehow that the schemas, if they successfully compile, are equivalent, but that not everything that can be compile with ajv can be processed with ajv-pack. As is the change may create impression that the compiled schemas are not equivalent.

Re placement - I don't mind either way, it's ok in browser section, there can be a link from security section if you like.

GrahamLea:
> I think it is important to comment somehow that the schemas, if they successfully compile, are equivalent, but that not everything that can be compile with ajv can be processed with ajv-pack.

Ah, yeah, thanks. I wasn't totally sure on what the deal was there. i.e. whether the limitations were in compiling or in the functionality of the compiled schemas. I've revised that text now to be clearer.

Also added a Contents link and a link from the Security section.

PR #1243, opened 2020-07-13
Bugfix should skip extendError if "error" option is "full" for custom keyword validate function

PhoenixHe-NV:
#1244

### DIFF
--- a/lib/dot/custom.jst
+++ b/lib/dot/custom.jst
@@ -180,7 +180,9 @@ var {{=$valid}};
           if (vErrors === null) vErrors = {{=$ruleErrs}};
           else vErrors = vErrors.concat({{=$ruleErrs}});
           errors = vErrors.length;
-          {{# def.extendErrors:false }}
+          {{? $rDef.errors != 'full' }}
+            {{# def.extendErrors:false }}
+          {{?}}
         } else {
           {{= def_customError }}
         }
--- a/spec/issues/1244_custom_validate_should_skip_extenderrors.spec.js
+++ b/spec/issues/1244_custom_validate_should_skip_extenderrors.spec.js
@@ -1,0 +1,43 @@
+'use strict';
+
+var Ajv = require('../ajv');
+require('../chai').should();
+
+
+describe('issue #1244: Should skip extendError if "error" option is "full"', function() {
+  it('should skip extendError if "error" option is "full"', function() {
+    var ajv = new Ajv({
+      verbose: true
+    });
+    ajv.addKeyword("propertyMustBeTrue", {
+      // eslint-disable-next-line no-unused-vars
+      validate: function v(schema, data, _parentSchema, dataPath) {
+        if (data && !data[schema]) {
+          v.errors = [{
+            keyword: "propertyMustBeTrue",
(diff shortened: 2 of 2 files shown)


PhoenixHe-NV:
Will add test later

PhoenixHe-NV:
@epoberezkin Could you please help to review this PR？Our product depends a lot on error reporting in ajv and is currently blocked by this bug. Thanks！

epoberezkin:
@NullMDR Thank you. I am looking. I think there might have been some reasoning behind the decision to call extend anyway, trying to find the traces from the time...

PhoenixHe-NV:
@epoberezkin Any update on this one?

PhoenixHe-NV:
ping @epoberezkin

PhoenixHe-NV:
ping @epoberezkin . Our code depends on ajv will go live soon and because of this bug we still need to maintain a fork of ajv. Could you please help to review this PR?

epoberezkin:
Sorry for the delay - I've reviewed it now.

The reason that the errors are extended with `dataPath` and `schemaPath` properties in all cases but custom **inline** keywords and `errors: "full"` option of a custom keyword, is that "compile", "validate" or "macro" custom keywords (i.e. all but inline keywords) do not have the current schema path to include it in errors. The "full" value of this option is only intended for "inline" keywords - so the current behaviour is not a bug, it is documented here: https://github.com/ajv-validator/ajv/blob/master/CUSTOM.md#reporting-errors-in-custom-keywords.

I think that it would be a breaking change and it would also require changing API to pass `schemaPath` to custom keywords or, maybe, extending errors only with `schemaPath` leaving `dataPath` as is if it is present. We need to think how custom keywords other than inline should change in v7, there are many objectives for keywords - I am working on it now.

Also, currently dataPath is overridden by extendError [only if it is not set](https://github.com/ajv-validator/ajv/blob/master/lib/dot/custom.jst#L68) - so you already can set dataPath in your errors. If you need to have `data` property in errors, you need to use `verbose: true` option in Ajv (although in your test case it would be the whole object rather than the specific property - it is where keyword is defined). Any other information can be included in `params` property of the error.

A smaller possible change would be to only set `.data` property in errors only if it is not already set ([here](https://github.com/ajv-validator/ajv/blob/master/lib/dot/custom.jst#L75)) - that would allow setting it in your custom keyword handler. If this line is changed to `if ({{=$ruleErr}}.data === undefined) {{=$ruleErr}}.data = {{=$data}};` your test would pass. That may be an option for v6 - I will think about it.

epoberezkin:
@NullMDR would this change work for you?

PhoenixHe-NV:
Oh yes thanks @epoberezkin . I'll test again with your proposed changes for check data undefined in extendError. 

What do you mean by "an option for v6"? And what's v7 anyway?

epoberezkin:
> What do you mean by "an option for v6"

this change could be possible in the current major version of Ajv (v6)

> And what's v7 anyway?

the next major version (WIP in v7-alpha branch)

epoberezkin:
closing it, the code has completely changed now.

PR #1251, opened 2020-07-22
Fix uri scheme validation

ChALkeR:
`foo,bar:baz` is not a valid uri.

Refs: https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/419


epoberezkin:
Probably worth updating JSON-Schema-test-suite.

It has to be added to ajv-formats too now...

ChALkeR:
That JSON-Schema-test-suite PR is not landed yet.

ChALkeR:
https://github.com/ajv-validator/ajv-formats/pull/1 for `ajv-formats` :slightly_smiling_face:

PR #1262, opened 2020-08-21
Add _opts property type in Ajv.Ajv

### LINKED ISSUE #1253: using "_opt.defaultMeta" in TypeScript
v6.12.3

**Your typescript code**

```typescript
import metaSchema from 'ajv/lib/refs/json-schema-draft-04.json';

// create Ajv instance(ajv) and more code...
ajv._opts.defaultMeta = metaSchema.id;
...
```

**Typescript compiler error messages**

```
Property '_opts' does not exist on type 'Ajv'.
```

**Describe the change that should be made to address the issue?**
Add type definition for "_opt".
Please tell me If there are other solutions.

I'll do this if a fix is ​​needed.

b4h0-c4t:
resolve #1253 issue.

Add type declaration Ajv.Ajv._opt

The reason why I set any type is other any object( ex. `_metaOpts` ) could not compatible with detailed objects.

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -137,7 +137,7 @@ declare namespace ajv {
     */
     errorsText(errors?: Array<ErrorObject> | null, options?: ErrorsTextOptions): string;
     errors?: Array<ErrorObject> | null;
-    _opts: any;
+    _opts: Options;
   }
 
   interface CustomLogger {
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -137,6 +137,7 @@ declare namespace ajv {
     */
     errorsText(errors?: Array<ErrorObject> | null, options?: ErrorsTextOptions): string;
     errors?: Array<ErrorObject> | null;
+    _opts: any;
   }
 
   interface CustomLogger {


epoberezkin (changes requested):
Shouldn’t it be the same interface as for Options?

b4h0-c4t:
@epoberezkin 
Thanks for your review.
I think `Options` do not have some `x._opts` properties. (e.x. defaultMeta)

b4h0-c4t:
Self-resolved.
`defaultMeta` can translated `meta`.

epoberezkin:
I am thinking it may be best to wait a bit until v7 is released in a week or so...

epoberezkin:
It might be important to access this property though - defaultMeta is not the same as meta, so it should be added too.

Merging it.

alasdairhurst:
@epoberezkin @b4h0-c4t is `defaultMeta` intended to be accessed exclusively via ajv._opts or is it available to be set in the constructor as part of the public API?

right now it's undocumented and i don't see  any recommendation for using it other than from the old v5 docs, but it seems to work if you set it in the constructor as long as you add the relevant metaSchema straight afterwards. Maybe i'm missing something important though.

PR #1289, opened 2020-09-23
fix issue1092

willson-chen:
https://github.com/ajv-validator/ajv/issues/1092

Report all duplicates with uniqueItems: "all" option.

Test cases

### DIFF
--- a/lib/ajv.ts
+++ b/lib/ajv.ts
@@ -91,6 +91,7 @@ interface CurrentOptions {
   $data?: boolean
   allErrors?: boolean
   verbose?: boolean
+  uniqueItems?: boolean | string
   $comment?:
     | true
     | ((comment: string, schemaPath?: string, rootSchema?: AnySchemaObject) => unknown)
@@ -146,7 +147,6 @@ interface RemovedOptions {
   strictDefaults?: boolean
   strictKeywords?: boolean
   strictNumbers?: boolean
-  uniqueItems?: boolean
   unknownFormats?: true | string[] | "ignore"
   cache?: any
   serialize?: (schema: AnySchema) => unknown
@@ -169,7 +169,6 @@ const removedOptions: OptionsInfo<RemovedOptions> = {
   strictDefaults: "It is default now, see option `strict`.",
   strictKeywords: "It is default now, see option `strict`.",
   strictNumbers: "It is default now, see option `strict`.",
-  uniqueItems: '"uniqueItems" keyword is always validated.',
   unknownFormats: "Disable strict mode or pass `true` to `ajv.addFormat` (or `formats` option).",
   cache: "Map is used as cache, schema object as key.",
   serialize: "Map is used as cache, schema object as key.",
--- a/lib/vocabularies/validation/uniqueItems.ts
+++ b/lib/vocabularies/validation/uniqueItems.ts
@@ -21,7 +21,7 @@ const def: CodeKeywordDefinition = {
   error,
   code(cxt: KeywordCxt) {
     const {gen, data, $data, schema, parentSchema, schemaCode, it} = cxt
-    if (!$data && !schema) return
+    if (!$data && !schema || it.opts.uniqueItems === false) return
     const valid = gen.let("valid")
     const itemTypes = parentSchema.items ? getSchemaTypes(parentSchema.items) : []
     cxt.block$data(valid, validateUniqueItems, _`${schemaCode} === false`)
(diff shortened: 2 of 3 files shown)


epoberezkin:
@willson-chen cool - thanks. I was trying to reduce the number of options and iniqueItems was removed too :)

My main concern with doing it this way, is that it can be easily exploited because the number of errors can be up to n^2 (if all items are the same). Also it will be different whether O(n) or O(n^2) algorithm is used.

I think we need to figure out how to have consistent error reporting regardless which algorithm is used... If that is achieved, then it can be just done under allErrors option, without the need for an additional option.

If somebody wants to disable uniqueItems completely they can remove the keyword and either use {strict: false} or more granular {keywords: {uniqueItems: true}}, so no need for a separate option either.

epoberezkin:
For example, instead of reporting error params as `{i, j}` (which is quite odd btw :) it could report each group of duplicates
 as `{duplicates: [...]}` with all duplicates collected in one array (and without allErrors there will be only two duplicates there). In this case when all items are the same there will be just one error.

It probably can be done if the keyword only reports the errors after the full scan, not in the process, collecting the duplicates in some data structure.

I would probably also add a warning when n^2 algorithm is used, maybe with option to suppress (e.g. allowSlowUniqueItems) - it will be consistent with the rest strict mode options.

epoberezkin:
sorry, it got automatically closed when I merged v7-alpha into v7-beta... - I don't seem to be able to edit it.

willson-chen:
@epoberezkin Sorry, I am late, next I will try to solve this problem, but before that I want to know:
1. Regardless of algorithm O(n) or O(n^2), must the error report be kept exactly the same?
2. If the `errors` are summarized in the `errorsText` method, the `errors` with the same `keyword` will be stored in the same array, for example, in the following example:
```
var ajv = new Ajv({allErrors: true});
const schema = {
  "type": "array",
  "items": {"type": "string"},
  "uniqueItems": true
}

var data = ["1", "1", "1",1,"1","1"];

var valid = validate(data);

if (valid) console.log('Valid!');
else {
  console.log('Invalid: ' + ajv.errorsText(validate.errors));
}
```
After processing `errors` in the `errorsText` method, we get two arrays:
```
[
  {
    keyword: 'type',
    dataPath: '/3',
    schemaPath: '#/items/type',
    params: { type: 'string' },
    message: 'should be string'
  }
]
```
and 
```
[
  {
    keyword: 'uniqueItems',
    dataPath: '',
    schemaPath: '#/uniqueItems',
    params: { i: 4, j: 5 },
    message: '5 and 4'
  },
  {
    keyword: 'uniqueItems',
    dataPath: '',
    schemaPath: '#/uniqueItems',
    params: { i: 2, j: 4 },
    message: '4 and 2'
  },
  {
    keyword: 'uniqueItems',
    dataPath: '',
    schemaPath: '#/uniqueItems',
    params: { i: 1, j: 2 },
    message: '2 and 1'
  },
  {
    keyword: 'uniqueItems',
    dataPath: '',
    schemaPath: '#/uniqueItems',
    params: { i: 0, j: 1 },
    message: '1 and 0'
  }
]
```
Then deal with the same error in the above array. Is this feasible?

epoberezkin:
Hi @willson-chen - thank you.

1. ON^2 vs ON - the list of errors will inevitably be different as ON algorithm doesn’t consider items when the type is wrong, so it won’t report any duplicates there. The shape of the errors should be the same though - indices of all same duplicates will be collected in one error object (with possibly multiple groups of duplicates, each group reported as a separate error). Alternatively, duplicates property could be array of arrays of indices - not sure what is more ergonomic for users - this or multiple errors.

2. In your example you have ON algorithm (because there is type in items and it’s primitive), but if it were ON2 algorithm you would have roughly N^2/2 errors. So collecting them in array and then merging would be very expensive for large arrays - not only you have N2 execution, but also N2 allocation... Possibly, there should be some configurable threshold after which it switches from N2 iteration with smaller constant factor (comparing using equal) to N iteration with larger constant factor - stable-stringifying and storing strings in hash. Benchmarks can help find what is the reasonable threshold.

Also, as @ChALkeR pointed out at some point, large hash maps increase insertion time (as they get “full” and stop being OlogN), so maybe it’s worth experimenting with benchmarks further and either switch to Maps (maybe they are faster) or to sorting - that is surprisingly faster on large arrays...

Either way, I think some additional data structure to track indices of duplicates should be kept using some hash maps so when something matches something it can check that it’s a known duplicate and add the index to the existing list or if it’s a new duplicate pair to create a new array for duplicates indices. Once all duplicates lists are collected, it would report all errors (or one error - TBD) at once, in a loop. A classic CS problem...

So in your example you should only have one error in any case.

PR #1318, opened 2020-11-07
Turn ErrorObject into a tagged union type

vecerek:
https://github.com/ajv-validator/ajv/issues/1319: The `ErrorObject`'s params are inaccessible in TypeScript. Example code:
```ts
const validationError: ajv.ValidationError = new ajv.ValidationError([{
    dataPath: ".someKey",
    keyword: "enum",
    params: { allowedValues: ["1", "2", "3"] },
    schemaPath: "",
}]);
validationError.errors[0].params.allowedValues; // TypeError, see below
```

```
$ tsc --target ES5 --noImplicitAny --noEmit spec/typescript/index.ts
spec/typescript/index.ts:60:34 - error TS2339: Property 'allowedValues' does not exist on type 'ErrorParameters'.
  Property 'allowedValues' does not exist on type 'RefParams'.

60 validationError.errors[0].params.allowedValues;
```

Turned `ErrorObject` into a tagged union type, the tag being the keyword; based on the [errors definition file](https://github.com/ajv-validator/ajv/blob/master/lib/dot/errors.def#L93-L123). I also discovered that there might have been one error params type missing, namely the `OneOfParams` type. So, I added it.

Yes, I didn't understand very well the code and the project structure, so I couldn't verify if all keywords are correct and that the relationship between a keyword and its params type is also correct.

### DIFF
--- a/lib/ajv.d.ts
+++ b/lib/ajv.d.ts
@@ -295,21 +295,48 @@ declare namespace ajv {
     dataVar?: string;
   }
 
-  interface ErrorObject {
-    keyword: string;
+  type ErrorObject = {
     dataPath: string;
     schemaPath: string;
-    params: ErrorParameters;
-    // Added to validation errors of propertyNames keyword schema
-    propertyName?: string;
     // Excluded if messages set to false.
     message?: string;
     // These are added with the `verbose` option.
     schema?: any;
     parentSchema?: object;
     data?: any;
-  }
+  } & (
+    | { keyword: "false schema"; params: NoParams }
+    | { keyword: "$ref"; params: RefParams }
+    | { keyword: "additionalItems"; params: LimitParams }
+    | { keyword: "additionalProperties"; params: AdditionalPropertiesParams }
+    | { keyword: "anyOf"; params:  NoParams }
+    | { keyword: "const"; params: EnumParams }
+    | { keyword: "contains"; params: NoParams  }
+    | { keyword: "dependencies"; params: DependenciesParams }
+    | { keyword: "enum"; params: EnumParams }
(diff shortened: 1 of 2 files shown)


epoberezkin:
@vecerek it is already done in v7 (it's under npm "beta" tag atm, in v7-beta branch) - I recommend migrating to it.

I will not be merging it to v6, sorry. Typescript support was always an afterthought in v6, while v7 is fully rewritten in typescript.

vecerek:
Thank you. I'll migrate over to v7 then 🙂

vecerek:
@epoberezkin I've just migrated over to v7 and noticed that although the params are accessible, the type safety is lost. `params` is just a `Record<string, any>`. Should I try and port the tagged union solution over to v7?

epoberezkin:
@vecerek sorry for the delay - busy week...

ErrorObject has to be open to allow users defining additional keywords.

There is a type DefinedError that extends ErrorObject and is [exported](https://github.com/ajv-validator/ajv/blob/master/lib/ajv.ts#L29) - this is the tagged union for all bundled keywords. If you need to access error parameters you would need to cast an error to this type. See this test: https://github.com/ajv-validator/ajv/blob/master/spec/types/error-parameters.spec.ts

Maybe there is a better solution, but I didn't figure out how to make it better. It cannot be just closed tagged union.

So if you just use bundled keywords, you just need this type, but if you define your own keyword(s) you would need to define your type that combines DefinedError with your keyword(s) error(s).

This reminds me that such type should be added to ajv-keywords.

It probably should be better documented.

epoberezkin:
Also see the example in readme: https://github.com/ajv-validator/ajv#getting-started

PR #1353, opened 2020-12-16
Add user-friendly message for maxLength validation

nicksrandall:
The error message "should not have fewer than 5 items" feels weird for a string of characters. I think we should revert back to v6 terminology and say "characters" instead of "items".


epoberezkin:
It's actually an accidental wrong change from v6 - it's incorrectly copied from maxItems - thank you!

epoberezkin:
in 7.0.1

PR #1354, opened 2020-12-17
errorsText: Fallback to this.errors for null/undefined

orgads:
Backwards compatibility with v6.

Fallback to this.errors when the input is null/undefined.

### DIFF
--- a/lib/core.ts
+++ b/lib/core.ts
@@ -577,9 +577,10 @@ export default class Ajv {
   }
 
   errorsText(
-    errors: ErrorObject[] | null | undefined = this.errors, // optional array of validation errors
+    errors: ErrorObject[] | null | undefined = undefined, // optional array of validation errors
     {separator = ", ", dataVar = "data"}: ErrorsTextOptions = {} // optional options with properties `separator` and `dataVar`
   ): string {
+    if (!errors) errors = this.errors;
     if (!errors || errors.length === 0) return "No errors"
     return errors
       .map((e) => `${dataVar}${e.dataPath} ${e.message}`)


epoberezkin:
I am probably missing something - what does it change? It would help to have a failing test that this change fixes or at least an example (to begin with, there probably has to be the test anyway if it is a regression)...

orgads:
If the user calls `ajv.errorsText(null)`, it used to use `this.errors`, but now it just shows `No errors`.

epoberezkin:
That's what I thought - but it means it was incorrect in v6.

The use case is that if you call it as `ajv.errorsText()` it should use `ajv.errors`, but if you call it as `ajv.errorsText(validate.errors)`, and `validate.errors` can be `null`, it should always use `validate.errors`, and if it is `null` it should return "No errors".

I may still be missing something, but it may be that there ~~was~~ is a bug in v6?

orgads:
I see. Then how do you suggest to pass options, but still use `ajv.errors`? Pass it directly: `ajv.errorsText(ajv.errors, { ... })`?

epoberezkin:
yes, that was the idea. Agree it's not the nicest API - we can change it in v8 which would be Q1'20 anyway.

orgads:
Q1'21 probably :)

orgads:
So that's what we did. I'll close it then. But you can probably expect it to pop from other users :)

epoberezkin:
right :) Classic New year mistake. Anyway, open to suggestions on the different API... Maybe Ajv should export the function to pass `validate.errors` to and the method only accepts options and doesn't accepts options (everybody would break though:)

epoberezkin:
how was it even working if you pass null?... Maybe it was working because users never combined `ajv.validate` approach with `ajv.compile` approach, but whoever did combine it on the same instance would have bugs in v6...

epoberezkin:
or do docs suggest to pass null?...

epoberezkin:
I've actually found some cases when I had to pass ajv.errors explicitly - it is an annoying API...

orgads:
There, [on v6](https://github.com/ajv-validator/ajv/blob/v6/lib/ajv.js#L412) it was:
```
  errors = errors || this.errors;
  if (!errors) return 'No errors';
```

[The docs](https://github.com/ajv-validator/ajv/tree/v6#errorstextarrayobject-errors--object-options---string) don't say much about this argument.

epoberezkin:
Yeah, it was bug-prone I think.

PR #1355, opened 2020-12-17
README: Fix broken links

orgads:


epoberezkin:
cool - thank you! (the build failed because GitHub doesn't expose secrets on PRs, but it also shows that it tries to update the website on PRs and it shouldn't happen...)

PR #1356, opened 2020-12-17
Add gitattributes to force eol=lf

orgads:
Without this, eslint fails on Windows when core.autocrlf is true

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.js eol=lf
+*.ts eol=lf


epoberezkin:
I think it's better to have environment specific settings in global git config, rather than in projects (there may be some historic noise in .gitignore, but I'd rather we remove it than add more). Also I use windows with git bash - never had this problem...

orgads:
You probably don't have `core.autocrlf=true`.

epoberezkin:
Possibly, either way, let’s not add platform specific configuration - it is better to have in your global git settings

orgads:
Why not?

PR #1376, opened 2021-01-04
Imports that are generated from standalone code generation are invalid

joelgallant:
When using ajv's 'standalone' module, imports to modules seem to be invalid. It looks like the files they should reference are in the `lib` directory. I don't know about ajv's internals, but this fixes my use case.

I searched for `require`'s of "dist" modules (the one I had troubles with was ecs2length), replacing ones that presumably should be "lib" modules.

I don't have context about why this is the way it is, maybe I'm missing something.

### DIFF
--- a/lib/compile/index.ts
+++ b/lib/compile/index.ts
@@ -108,7 +108,7 @@ export function compileSchema(this: Ajv, sch: SchemaEnv): SchemaEnv {
   if (sch.$async) {
     _ValidationError = gen.scopeValue("Error", {
       ref: ValidationError,
-      code: _`require("ajv/dist/compile/error_classes").ValidationError`,
+      code: _`require("ajv/lib/compile/error_classes").ValidationError`,
     })
   }
 
--- a/lib/vocabularies/validation/const.ts
+++ b/lib/vocabularies/validation/const.ts
@@ -17,7 +17,7 @@ const def: CodeKeywordDefinition = {
   code(cxt: KeywordCxt) {
     const eql = cxt.gen.scopeValue("func", {
       ref: equal,
-      code: _`require("ajv/dist/compile/equal")`,
+      code: _`require("ajv/lib/compile/equal")`,
     })
     // TODO optimize for scalar values in schema
     cxt.fail$data(_`!${eql}(${cxt.data}, ${cxt.schemaCode})`)
--- a/lib/vocabularies/validation/enum.ts
+++ b/lib/vocabularies/validation/enum.ts
@@ -21,7 +21,7 @@ const def: CodeKeywordDefinition = {
     const useLoop = schema.length >= it.opts.loopEnum
     const eql = cxt.gen.scopeValue("func", {
       ref: equal,
-      code: _`require("ajv/dist/compile/equal")`,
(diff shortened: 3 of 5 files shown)


joelgallant:
Judging by the failing tests from this (func4 is not a function), I think this issue might be stemming from AJV's in-tree tests vs how it's distributed? (ie. the requires are correct when in this repo, but the NPM distributed version of ajv no longer has those files)

maciejgoscinski:
+1
I experienced exactly this problem in my issue here:
https://github.com/ajv-validator/ajv-cli/issues/98

I'm using an npm-distributed ajv and ajv-cli. The require()'s in the generated code refer to nonexistent paths.

Other than that, would it seem useful to anyone else, to have an optional ES-style of imports/exports used in the codegen?

epoberezkin:
It would be good to have a minimal setup to reproduce the issue - I don’t think I understand it...

npm package includes compiled typescript in dist folder - this is what compiled code requires, whether it is used as standalone or inside closures returned by ajv methods.

Whether it is standalone or not, it is plain node.is-compatible JavaScript - it does work as is for me.

Could you explain the problem you experience and how to reproduce it?

epoberezkin:
> requires are correct when in this repo, but the NPM distributed version of ajv no longer has those file

I did just check - dist folder with these files is included in npm package.

Are you using plain node.js or is there some build step that would somehow process these "requires"? If the latter then it may be that it is either configured incorrectly or is not 100% compatible with node.js

epoberezkin:
For example of where it works, please check gajus/table package - it uses ajv-cli to compile schemas into standalone code - it was released with ajv v7 a few days ago.

maciejgoscinski:
@epoberezkin I just created an empty repository to demonstrate, and you're right indeed - the aforementioned directories in dist do exist, working with both npm and yarn.
I'm investigating what could be the reason for failure in my project repository and why could that be happening for others, too.

@joelgallant by any chance, are you using yarn workspaces? Do you define your ajv as a dependency, or devDependency?

maciejgoscinski:
@joelgallant I managed to discover and fix the problem on my side.
Make sure your dependencies are in line and check the actual resolved version numbers in node_modules package.jsons of AJV and AJV-cli.

For example, as of today, these latest versions go well together (use exact versions without ~ or ^):
```
    "ajv": "7.0.3",
    "ajv-cli": "4.0.1"
```

It might be the case that if your project has been alive for some time, your package manager was not bumping the deps due to package-lock or yarn.lock, even though the ones specified in package.json were already increased from 6.x to 7.x.
In my case the yarn upgrade command wasn't working as I expected, and there was an outdated version still being downloaded.

joelgallant:
Interesting. I am using yarn workspaces, and it was resolving to AJV v6 (because ajv was a transitive dependency of the package using generated code).

So it looks like this PR is wrong, and I'll need to make v7 a peerDependency. That makes a lot more sense now.

epoberezkin:
Great - it’s simpler than I thought :) I was worried about the possibility of some not 100% nodejs compatible build tool :)

They don’t have to be fixed versions - any changes from 7.0.0 and 4.0.0 of ajv/ajv-cli will be backwards compatible until the next major version.

PR #1443, opened 2021-02-15
Add URI import option

ninofiliu:

No opened issue

AJV can be imported directly from a URL and besides that being nice to know when coding a client app, that's the only way to import it when using it with Deno

I used [Skypack](https://www.skypack.dev/) as the standard because IMHO that's the most promising service to deliver ES modules but don't hesitate to use another service in the example


ninofiliu:
Eh, whatever, nobody seems to care about it, I'm closing

PR #1446, opened 2021-02-16
[RFC] JTD Schema Object

erikbrinkman:
Adds a type like the `JSONSchemaType` but for JTD schemas. Notably, because unions are tagged in JTD by default, the discriminator type can determine if the sub schemas match each of the union elements, and there's no indecisiveness between `undefined` and `null`.

This is very much an RFC in that there's a few design decisions that should probably be made before adding this, and there's a few behaviors of this current submission that aren't ideal.

Added the type, and tests to verify typescripts behavior, although some of the verified behavior is not ideal.

There are three major decisions that should probably be decided on, and then there's a number of behaviors that I've called out as odd in the code that should probably also be resolved.

1. The first major decision is if this should only accept schemas that fit the JTD schema definition? For example, JTD says `definitions` is only allowed in a root schema, but typescript doesn't like forbidding properties (but it can be done). Should this definition error if anything extra is specified in a schema (at least anything that's expressly forbidden).
2. Should this type definition be "complete" (not sure if this is really the correct word). I tend to lean towards `iii`, but that's a preference. My thoughts here are messy, but basically the type could:
   1. typecheck if the schema is valid for that type. e.g. this would allow the empty schema for anything. This is valid according to the JTD schema, but invalid for a type assertion.
   2. typecheck if the schema is valid for an instance. e.g. this would allow `Schema<number | string> = { type: "string" }` on other words, validating the schema would produce an instance of the presented type, even if it's narrower.
   3. Both of these. This would mean there are types where the valid schema is `never`, like the above `number | string` since it's not a tagged union.
3. How should AJV extensions be handled, e.g. currently there's no support for `Date` being a valid `timestamp` or specific fields in `metadata` like untagged unions that I'm not sure can be done well.

Other notes from the code that don't fall under the above decisions:
1. for `enum` types there's no way to make sure that everything was specified. This technically means that only `ii` above is possible, but could potentially be overlooked. There might be a way to create a sorted tuple type, but it would add the complexity that enum type must be in sorted order. Not clear how feasible this is or what the runtime of checking it would be for large enumerations.

### DIFF
--- a/lib/ajv.ts
+++ b/lib/ajv.ts
@@ -26,7 +26,6 @@ export {KeywordCxt}
 export {DefinedError} from "./vocabularies/errors"
 export {JSONType} from "./compile/rules"
 export {JSONSchemaType} from "./types/json-schema"
-export {JTDSchema} from "./types/jtd-schema"
 export {_, str, stringify, nil, Name, Code, CodeGen, CodeGenOptions} from "./compile/codegen"
 
 import type {AnySchemaObject} from "./types"
--- a/lib/ajv.ts
+++ b/lib/ajv.ts
@@ -26,6 +26,7 @@ export {KeywordCxt}
 export {DefinedError} from "./vocabularies/errors"
 export {JSONType} from "./compile/rules"
 export {JSONSchemaType} from "./types/json-schema"
+export {JTDSchema} from "./types/jtd-schema"
 export {_, str, stringify, nil, Name, Code, CodeGen, CodeGenOptions} from "./compile/codegen"
 
 import type {AnySchemaObject} from "./types"
--- a/lib/types/jtd-schema.ts
+++ b/lib/types/jtd-schema.ts
@@ -1,0 +1,103 @@
+/** required keys of an object, not undefined */
+type RequiredKeys<T> = {
+  [K in keyof T]-?: undefined extends T[K] ? never : K
+}[keyof T]
+
+/** optional or undifined-able keys of an object */
+type OptionalKeys<T> = {
+  [K in keyof T]-?: undefined extends T[K] ? K : never
+}[keyof T]
+
+/** type is true if all elements of T extend T */
+type AllExtend_<T, E> = T extends E ? true : false
+type AllExtend<T, E> = false extends AllExtend_<T, E> ? false : true
+
+/** type is true if type is identically string */
+type IsString_<T> = T extends string ? (string extends T ? true : false) : false
+type IsString<T> = false extends IsString_<T> ? false : true
+
+/** gets only the string literals of a type or null if a type isn't a string literal */
+type NullableStringLiterals<T> = T extends string ? (string extends T ? null : T) : null
+
+/** numeric strings */
+type NumberType = "float32" | "float64" | "int8" | "uint8" | "int16" | "uint16" | "int32" | "uint32"
+
+/** string strings */
+type StringType = "string" | "timestamp"
+
+/** actual schema */
+export type JTDSchema<T, D extends Record<string, unknown> = Record<string, never>> = (
+  | // refs - accepts any key in definitions of the right type
+  {[K in keyof D]: T extends D[K] ? {ref: K} : never}[keyof D]
+  // numbers - only accepts number or union of number literals
+  | (AllExtend<T, number | null> extends true ? {type: NumberType} : never)
+  // booleans - accepts any boolean, incluting just true or false literals
+  | (AllExtend<T, boolean | null> extends true ? {type: "boolean"} : never)
+  // strings - only accepts the type string
+  // TODO accept Date?
+  | (IsString<Exclude<T, null>> extends true ? {type: StringType} : never)
+  // enums - only accepts union of string literals
+  | (// TODO we can't actually check that everything in the union was specified
+    null extends NullableStringLiterals<Exclude<T, null>>
+      ? never
+      : {enum: NullableStringLiterals<Exclude<T, null>>[]})
+  // arrays - only accepts arrays, could be array of unions to be resolved later
+  | (T extends (infer E)[]
+      ? {
+          elements: JTDSchema<E, D>
+        }
+      : never)
+  // all object like
+  | (keyof T extends string
+      ?
+          | // values
+          {
+              values: JTDSchema<T[keyof T], D>
+            }
+          // explicit keys
+          | ((RequiredKeys<Exclude<T, null>> extends never
+              ? {
+                  properties?: Record<string, never>
+                }
+              : {
+                  properties: {[K in RequiredKeys<T>]: JTDSchema<T[K], D>}
+                }) &
+              (OptionalKeys<Exclude<T, null>> extends never
+                ? {
+                    optionalProperties?: Record<string, never>
+                  }
+                : {
+                    optionalProperties: {
+                      [K in OptionalKeys<T>]: JTDSchema<Exclude<T[K], undefined>, D>
+                    }
+                  }) & {
+                additionalProperties?: boolean
+              })
+          // tagged unions
+          | {
+              [K in keyof T]-?: T[K] extends string
+                ? {
+                    discriminator: K
+                    mapping: {
+                      // TODO currently allows descriminator to be present in schema
+                      [M in T[K]]: JTDSchema<Omit<T extends {[D in K]: M} ? T : never, K>, D>
+                    }
+                  }
+                : never
+            }[keyof T]
+          // empty
+          // TODO as written, and object like could get an empty schema which
+          // potentially breaks guarantees, but does match the spec
+          | Record<string, never>
+      : never)
+) & {
+  // extra properties
+  // TODO handle ajv extensions to metadata?
+  metadata?: {[name: string]: unknown}
+  // TODO these should only be allowed at the top level, but typescript doesn't like saying no
+  definitions?: {[K in keyof D]: JTDSchema<D[K], D>}
+} & (null extends T
+    ? {
+        nullable: true
+      }
+    : {nullable?: false})
--- a/lib/types/jtd-schema.ts
+++ b/lib/types/jtd-schema.ts
@@ -8,16 +8,34 @@ type OptionalKeys<T> = {
   [K in keyof T]-?: undefined extends T[K] ? K : never
 }[keyof T]
 
-/** type is true if all elements of T extend T */
-type AllExtend_<T, E> = T extends E ? true : false
-type AllExtend<T, E> = false extends AllExtend_<T, E> ? false : true
+/** type is true if T is a union type */
+type IsUnion_<T, U extends T = T> = false extends (
+  T extends unknown ? ([U] extends [T] ? false : true) : never
+)
+  ? false
+  : true
+type IsUnion<T> = IsUnion_<T>
 
-/** type is true if type is identically string */
-type IsString_<T> = T extends string ? (string extends T ? true : false) : false
-type IsString<T> = false extends IsString_<T> ? false : true
+/** type is true if T is identically E */
+type TypeEquality<T, E> = [T] extends [E] ? ([E] extends [T] ? true : false) : false
 
 /** gets only the string literals of a type or null if a type isn't a string literal */
-type NullableStringLiterals<T> = T extends string ? (string extends T ? null : T) : null
+type EnumString<T> = [T] extends [never]
+  ? null
+  : T extends string
+  ? string extends T
+    ? null
+    : T
+  : null
+
+/** true only if all types are array types (not tuples) */
+// NOTE relies on the fact that tuples don't have an index at 0.5, but arrays
+// have an index at every number
+type IsElements<T> = [T] extends [readonly unknown[]]
+  ? undefined extends T[0.5]
+    ? false
+    : true
+  : false
 
 /** numeric strings */
 type NumberType = "float32" | "float64" | "int8" | "uint8" | "int16" | "uint16" | "int32" | "uint32"
@@ -26,78 +44,108 @@ type NumberType = "float32" | "float64" | "int8" | "uint8" | "int16" | "uint16" 
 type StringType = "string" | "timestamp"
 
 /** actual schema */
-export type JTDSchema<T, D extends Record<string, unknown> = Record<string, never>> = (
-  | // refs - accepts any key in definitions of the right type
-  {[K in keyof D]: T extends D[K] ? {ref: K} : never}[keyof D]
-  // numbers - only accepts number or union of number literals
-  | (AllExtend<T, number | null> extends true ? {type: NumberType} : never)
-  // booleans - accepts any boolean, incluting just true or false literals
-  | (AllExtend<T, boolean | null> extends true ? {type: "boolean"} : never)
-  // strings - only accepts the type string
-  // TODO accept Date?
-  | (IsString<Exclude<T, null>> extends true ? {type: StringType} : never)
-  // enums - only accepts union of string literals
-  | (// TODO we can't actually check that everything in the union was specified
-    null extends NullableStringLiterals<Exclude<T, null>>
-      ? never
-      : {enum: NullableStringLiterals<Exclude<T, null>>[]})
-  // arrays - only accepts arrays, could be array of unions to be resolved later
-  | (T extends (infer E)[]
+export type JTDSchemaType<T, D extends Record<string, unknown> = Record<string, never>> = (
+  | // refs - where null wasn't specified, must match exactly
+  ({[K in keyof D]: [T] extends [D[K]] ? {ref: K} : never}[keyof D] & {nullable?: false})
+  // nulled refs - if ref is nullable and nullable is specified, then it can
+  // match either null or non-null definitions
+  | (null extends T
       ? {
-          elements: JTDSchema<E, D>
-        }
+          [K in keyof D]: [Exclude<T, null>] extends [Exclude<D[K], null>] ? {ref: K} : never
+        }[keyof D] & {nullable: true}
       : never)
-  // all object like
-  | (keyof T extends string
-      ?
-          | // values
-          {
-              values: JTDSchema<T[keyof T], D>
-            }
-          // explicit keys
-          | ((RequiredKeys<Exclude<T, null>> extends never
+  // all other types
+  | ((
+      | // numbers - only accepts the type number
+      (true extends TypeEquality<Exclude<T, null>, number> ? {type: NumberType} : never)
+      // booleans - accepts the type boolean
+      | (true extends TypeEquality<Exclude<T, null>, boolean> ? {type: "boolean"} : never)
+      // strings - only accepts the type string
+      | (true extends TypeEquality<Exclude<T, null>, string> ? {type: StringType} : never)
+      // strings - only accepts the type Date
+      | (true extends TypeEquality<Exclude<T, null>, Date> ? {type: "timestamp"} : never)
+      // enums - only accepts union of string literals
+      // TODO we can't actually check that everything in the union was specified
+      | (null extends EnumString<Exclude<T, null>> ? never : {enum: EnumString<Exclude<T, null>>[]})
+      // arrays - only accepts arrays, could be array of unions to be resolved later
+      | (false extends IsUnion<Exclude<T, null>>
+          ? true extends IsElements<Exclude<T, null>>
+            ? T extends readonly (infer E)[]
               ? {
-                  properties?: Record<string, never>
+                  elements: JTDSchemaType<E, D>
                 }
-              : {
-                  properties: {[K in RequiredKeys<T>]: JTDSchema<T[K], D>}
-                }) &
-              (OptionalKeys<Exclude<T, null>> extends never
+              : never
+            : never
+          : never)
+      // values
+      | (false extends IsUnion<Exclude<T, null>>
+          ? true extends TypeEquality<keyof Exclude<T, null>, string>
+            ? T extends Record<string, infer V>
+              ? {
+                  values: JTDSchemaType<V>
+                }
+              : never
+            : never
+          : never)
+      // properties
+      | (false extends IsUnion<Exclude<T, null>>
+          ? null extends EnumString<keyof Exclude<T, null>>
+            ? never
+            : ([RequiredKeys<Exclude<T, null>>] extends [never]
                 ? {
-                    optionalProperties?: Record<string, never>
+                    properties?: Record<string, never>
                   }
                 : {
-                    optionalProperties: {
-                      [K in OptionalKeys<T>]: JTDSchema<Exclude<T[K], undefined>, D>
+                    properties: {[K in RequiredKeys<T>]: JTDSchemaType<T[K], D>}
+                  }) &
+                ([OptionalKeys<Exclude<T, null>>] extends [never]
+                  ? {
+                      optionalProperties?: Record<string, never>
                     }
-                  }) & {
-                additionalProperties?: boolean
-              })
-          // tagged unions
-          | {
-              [K in keyof T]-?: T[K] extends string
-                ? {
-                    discriminator: K
-                    mapping: {
-                      // TODO currently allows descriminator to be present in schema
-                      [M in T[K]]: JTDSchema<Omit<T extends {[D in K]: M} ? T : never, K>, D>
+                  : {
+                      optionalProperties: {
+                        [K in OptionalKeys<T>]: JTDSchemaType<Exclude<T[K], undefined>, D>
+                      }
+                    }) & {
+                  additionalProperties?: boolean
+                }
+          : never)
+      // discriminator
+      | (true extends IsUnion<Exclude<T, null>>
+          ? null extends EnumString<keyof Exclude<T, null>>
+            ? never
+            : {
+                [K in keyof Exclude<T, null>]-?: Exclude<T, null>[K] extends string
+                  ? {
+                      discriminator: K
+                      mapping: {
+                        // TODO currently allows descriminator to be present in schema
+                        [M in Exclude<T, null>[K]]: JTDSchemaType<
+                          Omit<T extends {[C in K]: M} ? T : never, K>,
+                          D
+                        >
+                      }
                     }
-                  }
-                : never
-            }[keyof T]
-          // empty
-          // TODO as written, and object like could get an empty schema which
-          // potentially breaks guarantees, but does match the spec
-          | Record<string, never>
-      : never)
+                  : never
+              }[keyof Exclude<T, null>]
+          : never)
+      // empty schema
+      // NOTE there should only be one type that extends Record<string, never> so unions
+      // shouldn't be a worry
+      | (T extends Record<string, never> ? unknown : never)
+      // null
+      // NOTE we have to check this too because null as an exclusive type also
+      // qualifies for the empty schema
+      | (true extends TypeEquality<T, null> ? unknown : never)
+    ) &
+      (null extends T
+        ? {
+            nullable: true
+          }
+        : {nullable?: false}))
 ) & {
   // extra properties
-  // TODO handle ajv extensions to metadata?
-  metadata?: {[name: string]: unknown}
-  // TODO these should only be allowed at the top level, but typescript doesn't like saying no
-  definitions?: {[K in keyof D]: JTDSchema<D[K], D>}
-} & (null extends T
-    ? {
-        nullable: true
-      }
-    : {nullable?: false})
+  metadata?: Record<string, unknown>
+  // TODO these should only be allowed at the top level
+  definitions?: {[K in keyof D]: JTDSchemaType<D[K], D>}
+}
--- a/spec/types/jtd-schema.spec.ts
+++ b/spec/types/jtd-schema.spec.ts
@@ -1,0 +1,212 @@
+/* eslint-disable @typescript-eslint/no-empty-interface */
+import type {JTDSchema} from "../.."
+import chai from "../chai"
+const should = chai.should()
+
+describe("JTDSchema typechecks", () => {
+  it("should typecheck number schemas", () => {
+    const numf: JTDSchema<number> = {type: "float64"}
+    should.exist(numf)
+    const numi: JTDSchema<number> = {type: "int32"}
+    should.exist(numi)
+    // @ts-expect-error
+    const numl: JTDSchema<number> = {type: "int64"}
+    should.exist(numl)
+  })
+
+  it("should typecheck boolean schemas", () => {
+    const bool: JTDSchema<boolean> = {type: "boolean"}
+    should.exist(bool)
+    // boolean literals can't be reduced
+    const boolTrue: JTDSchema<true> = {type: "boolean"}
+    should.exist(boolTrue)
+  })
+
+  it("should typecheck string schemas", () => {
+    const str: JTDSchema<string> = {type: "string"}
+    should.exist(str)
+    const time: JTDSchema<string> = {type: "timestamp"}
+    should.exist(time)
+  })
+
+  it("should typecheck enumeration schemas", () => {
+    const enumerate: JTDSchema<"a" | "b"> = {enum: ["a", "b"]}
+    should.exist(enumerate)
+    // don't need to specify everything
+    const enumerateMissing: JTDSchema<"a" | "b" | "c"> = {enum: ["a", "b"]}
+    should.exist(enumerateMissing)
+    // @ts-expect-error
+    const enumerateNumber: JTDSchema<"a" | "b" | 5> = {enum: ["a", "b"]}
+    should.exist(enumerateNumber)
+    // @ts-expect-error
+    const enumerateString: JTDSchema<"a" | "b"> = {type: "string"}
+    should.exist(enumerateString)
+  })
+
+  it("should typecheck elements schemas", () => {
+    const elements: JTDSchema<number[]> = {elements: {type: "float64"}}
+    should.exist(elements)
+    // homogenous tuples works
+    const tupleHomo: JTDSchema<[number, number]> = {elements: {type: "float64"}}
+    should.exist(tupleHomo)
+    // not heterogeneous
+    const tupleHeteroNum: JTDSchema<[number, string]> = {
+      // @ts-expect-error
+      elements: {type: "float64"},
+    }
+    should.exist(tupleHeteroNum)
+    const tupleHeteroString: JTDSchema<[number, string]> = {
+      // @ts-expect-error
+      elements: {type: "string"},
+    }
+    should.exist(tupleHeteroString)
+  })
+
+  it("should typecheck values schemas", () => {
+    const values: JTDSchema<Record<string, number>> = {values: {type: "float64"}}
+    should.exist(values)
+    const valuesDefined: JTDSchema<{prop: number}> = {values: {type: "float64"}}
+    should.exist(valuesDefined)
+  })
+
+  it("should typecheck properties schemas", () => {
+    const properties: JTDSchema<{a: number; b: string}> = {
+      properties: {a: {type: "float64"}, b: {type: "string"}},
+    }
+    should.exist(properties)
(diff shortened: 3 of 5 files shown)


epoberezkin:
This is great - I was thinking to add it, but I didn't have an idea how to make discriminator work. From briefly looking at PR - I am definitely not on this level of TS typefoo - some things there go over my head - need to read slowly :)

Some comments on the raised questions:

> 1. The first major decision is if this should only accept schemas that fit the JTD schema definition? For example, JTD says definitions is only allowed in a root schema, but typescript doesn't like forbidding properties (but it can be done). Should this definition error if anything extra is specified in a schema (at least anything that's expressly forbidden).

On a similar question, when I was writing a meta-schema for JTD schemas I've made it to explicitly prevent definitions anywhere but in the root, but given that JTD itself is not designed for such purpose, it doubled meta-schema size. I avoided the repetition on the code level, but on the data level the schema is still twice bigger - potentially meaning some memory/performance penalties (although not double time) for what could have been effectively handled much simpler - by creating `definitions` keyword that would throw the exception anywhere but in the root at compilation time and do nothing else. I am thinking to change to that.

On the question here, I am absolutely ok with not enforcing this restriction on type level. Typescript does flag unknown properties, and we should not allow unknown properties (if possible - need to read the code attentively) but trying to prevent definitions from appearing anywhere other than in the root would also make this type heavier (not sure how much heavier though - you may know how to do it in a simple way). So happy to leave as is and happy to add this restriction too if it doesn't complicate things too much.

I think it goes up to the higher level question - what is this type purpose?

The options are:

1) define the type that effectively replaces the need for syntax and semantic validation. From your other questions, it seems follows that it is not possible in full anyway.
2) define the type that assists in manually writing JTD schemas in a way that the schema is consistent - with data type itself, and still allow some errors be caught at schema validation/compilation time. This seems a simpler and the most valuable part - so I'd stick with that.

> 2. Should this type definition be "complete" (not sure if this is really the correct word). I tend to lean towards iii, but that's a preference. My thoughts here are messy, but basically the type could:

same, i + ii makes sense

> i. typecheck if the schema is valid for that type. e.g. this would allow the empty schema for anything. This is valid according to the JTD schema, but invalid for a type assertion.

assuming you mean checking if the schema is valid according to JTD spec but not necessarily consistent with instance type. From goal 2 above, in isolation it seems of low value - although we could of course define a generic type for any schema as well - as `JSONSchemaType` does to get to `SomeJSONSchema`, we could make SomeJTDSchema here as well.

> ii. typecheck if the schema is valid for an instance. e.g. this would allow Schema<number | string> = { type: "string" } on other words, validating the schema would produce an instance of the presented type, even if it's narrower.

assuming you mean the type of the instance - then yes.

> 3. How should AJV extensions be handled, e.g. currently there's no support for Date being a valid timestamp or specific fields in metadata like untagged unions that I'm not sure can be done well.

It should allow Date in place of timestamp type - see `vocabularies/jtd/type.ts`, although it is neither tested nor documented - needs to be improved.

I would not worry about handling extensions inside metadata field on the type level. Ajv only allowed such extensions (and only inside metadata) for two reasons:

1) to have meta-schema - If I was writing a separate JTD validator, I would not do it this way, but this approach fits better with Ajv model.
2) to simplify migration from JSON Schema - you could simply drop JSON Schema keywords there (and add them to the instance of course).

In general, I think users should really avoid using such extensions in the new schema, based on the differences between design objectives of JTD and JSON Schema.

I have spent some time objecting to user defined keywords JSON Schema, but given that JSON Schema approaches the level of complexity of general purpose programming language (and you can actually use ajv with JSON Schema to interpret another JSON based general purpose language with some extension keywords - see my abandoned JSON Script), having user define keywords in JSON Schema fits this model, and it can be helpful for traversing and folding the trees with some arbitrary logic on the way.

JTD on another hand is fulfilling a much smaller, but, arguably, as important  objective: define a JSON based specification for defining schemas of cross-platform JSON messages, specifically in a language-heterogenous environment, in such way that a simple schema can be used to generate types for multiple languages - hence the name (which I proudly contributed to:).

It is much more restrictive by design, either because certain data structures/types are not present in some important languages and using them would undermine its cross-platform-ness, or because these data structures are difficult to extend reliably, specifically:
- tuples/heterogenous arrays - better avoided in JSON messages, as they are neither ubiquitous nor extensible
- untagged unions - not present even in some advanced languages
- 64 bits integers

So in a way, JTD guides your JSON message / API modelling in such a way that you would not have challenges adding services in any language later. So I'd strongly discourage using any extensions and instead move any value/semantic validation logic to the application code.

> for enum types there's no way to make sure that everything was specified. This technically means that only ii above is possible, but could potentially be overlooked. There might be a way to create a sorted tuple type, but it would add the complexity that enum type must be in sorted order. Not clear how feasible this is or what the runtime of checking it would be for large enumerations.

Yep, I had the same issue in JSONSchemaType - didn't realise it is solvable with restrictions. It is actually not such a bad idea - maybe it can be a separate version of this type (or extra parameter) or we could just make this call that this type in opinionated in this way - some people enforce sorting of everything via linting  and I tend to gravitate towards sorting things - it makes them easier to find in large lists and avoids large diffs at a point when the unsorted list grows out of control and you eventually sort it... Maybe it is better to start with sorted. Up to you, maybe deserves a separate PR, but I am on board with enforcing sorting of enums in schemas - unless it is super complex. I would do it in JSONSchemaType too then (in v8).

Anyway, sorry for the brain dump, what you've done is great - I'll read the code carefully - it'll take some effort :) - and comment. All your tradeoffs make sense to me based on the above.

lib/ajv.ts
```diff
@@ -26,6 +26,7 @@ export {KeywordCxt}
 export {DefinedError} from "./vocabularies/errors"
 export {JSONType} from "./compile/rules"
 export {JSONSchemaType} from "./types/json-schema"
+export {JTDSchema} from "./types/jtd-schema"
```
epoberezkin: I do not think it needs to be added to ajv.ts, it should probably be in jtd.ts?

I would also rename it to JTDSchemaType for consistency sake (although happy to debate it and change in v8 if you have some strong arguments for not using Type syffix).

With JTD you would be improrting ajv and the type from this file:

```typescript
import Ajv, {JTDSchemaType} from "ajv/dist/jtd"
```

lib/types/jtd-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+type OptionalKeys<T> = {
+  [K in keyof T]-?: undefined extends T[K] ? K : never
+}[keyof T]
+
+/** type is true if all elements of T extend T */
+type AllExtend_<T, E> = T extends E ? true : false
+type AllExtend<T, E> = false extends AllExtend_<T, E> ? false : true
+
+/** type is true if type is identically string */
+type IsString_<T> = T extends string ? (string extends T ? true : false) : false
```
epoberezkin: I am wondering why the second line is necessary here? The first line seems to say that for T to be identically string, T should extend string and string should extend T - which I thought would be enough, but it is probably not correct given that the second line is needed... And I think for majority of T the first line is enough, but there are some Ts when it is not - what are they?

As a side comment - Typescript is weird really if you just can have type level equality without these trick :)

AllExtend I understand even less, but it looks similar - what is the example there?

Maybe it is worth defining a utility type `TypeEquality<T1, T2>` that would only evaluate to true only if T1 is identical to T2 - although I am not sure if what happens here with string can be generalised - maybe not...

lib/types/jtd-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+type RequiredKeys<T> = {
+  [K in keyof T]-?: undefined extends T[K] ? never : K
+}[keyof T]
+
+/** optional or undifined-able keys of an object */
+type OptionalKeys<T> = {
+  [K in keyof T]-?: undefined extends T[K] ? K : never
+}[keyof T]
+
+/** type is true if all elements of T extend T */
```
epoberezkin: It's probably typo, should be `type is true if all elements of E extend T` (not certain, but seems so).

lib/types/jtd-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+type AllExtend_<T, E> = T extends E ? true : false
+type AllExtend<T, E> = false extends AllExtend_<T, E> ? false : true
+
+/** type is true if type is identically string */
+type IsString_<T> = T extends string ? (string extends T ? true : false) : false
+type IsString<T> = false extends IsString_<T> ? false : true
+
+/** gets only the string literals of a type or null if a type isn't a string literal */
+type NullableStringLiterals<T> = T extends string ? (string extends T ? null : T) : null
```
epoberezkin: maybe worth renaming to EnumString - shorter and connects to enum in schema - took me some time to "compute" it...

lib/types/jtd-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+/** actual schema */
+export type JTDSchema<T, D extends Record<string, unknown> = Record<string, never>> = (
+  | // refs - accepts any key in definitions of the right type
+  {[K in keyof D]: T extends D[K] ? {ref: K} : never}[keyof D]
+  // numbers - only accepts number or union of number literals
+  | (AllExtend<T, number | null> extends true ? {type: NumberType} : never)
+  // booleans - accepts any boolean, incluting just true or false literals
+  | (AllExtend<T, boolean | null> extends true ? {type: "boolean"} : never)
+  // strings - only accepts the type string
+  // TODO accept Date?
```
epoberezkin: yes - Date should be supported and require timestamp type

lib/types/jtd-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+/** numeric strings */
+type NumberType = "float32" | "float64" | "int8" | "uint8" | "int16" | "uint16" | "int32" | "uint32"
+
+/** string strings */
+type StringType = "string" | "timestamp"
+
+/** actual schema */
+export type JTDSchema<T, D extends Record<string, unknown> = Record<string, never>> = (
+  | // refs - accepts any key in definitions of the right type
+  {[K in keyof D]: T extends D[K] ? {ref: K} : never}[keyof D]
```
epoberezkin: would it work if there are multiple definitions that match the required type.

I'd never thought it's possible...

lib/types/jtd-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+      ? {
+          elements: JTDSchema<E, D>
+        }
+      : never)
+  // all object like
+  | (keyof T extends string
+      ?
+          | // values
+          {
+              values: JTDSchema<T[keyof T], D>
```
epoberezkin: Does it mean it would allow values with schema matching the type of any property?

Or would it only allow it if all properties have the same type?

Could we maybe say that the type has to extend `{[K in string]?: T}` to allow "values"?

By the way, a side note, I am going to add code generation for parser/serializer - we could support Map in this context with some extra property in metadata. Just an idea, no need for this PR

lib/types/jtd-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+                additionalProperties?: boolean
+              })
+          // tagged unions
+          | {
+              [K in keyof T]-?: T[K] extends string
+                ? {
+                    discriminator: K
+                    mapping: {
+                      // TODO currently allows descriminator to be present in schema
+                      [M in T[K]]: JTDSchema<Omit<T extends {[D in K]: M} ? T : never, K>, D>
```
epoberezkin: it does look like these two Ds mean different things... Or it's over my head how it works - need to read it slowly :)

lib/types/jtd-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+                    }
+                  }
+                : never
+            }[keyof T]
+          // empty
+          // TODO as written, and object like could get an empty schema which
+          // potentially breaks guarantees, but does match the spec
```
epoberezkin: I don't think we need to allow empty schema for the sake of just matching the schema. Could we make that it is only allowed in cases when none of the previous types were matched, even at the cost of reducing the space (as I did in JSONSchemaType with Known - not sure if it was a good idea tbh)?

spec/types/jtd-schema.spec.ts
```diff
@@ -0,0 +1,212 @@
+/* eslint-disable @typescript-eslint/no-empty-interface */
+import type {JTDSchema} from "../.."
+import chai from "../chai"
+const should = chai.should()
+
+describe("JTDSchema typechecks", () => {
+  it("should typecheck number schemas", () => {
+    const numf: JTDSchema<number> = {type: "float64"}
+    should.exist(numf)
```
epoberezkin: if the only reason why should.exist is needed is to prevent typescript complaining could we prefix with underscores and remove all should.exist to reduce the noise?

spec/types/jtd-schema.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+    // boolean literals can't be reduced
+    const boolTrue: JTDSchema<true> = {type: "boolean"}
+    should.exist(boolTrue)
+  })
+
+  it("should typecheck string schemas", () => {
+    const str: JTDSchema<string> = {type: "string"}
+    should.exist(str)
+    const time: JTDSchema<string> = {type: "timestamp"}
+    should.exist(time)
```
epoberezkin: \+ Date

spec/types/jtd-schema.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+      elements: {type: "string"},
+    }
+    should.exist(tupleHeteroString)
+  })
+
+  it("should typecheck values schemas", () => {
+    const values: JTDSchema<Record<string, number>> = {values: {type: "float64"}}
+    should.exist(values)
+    const valuesDefined: JTDSchema<{prop: number}> = {values: {type: "float64"}}
+    should.exist(valuesDefined)
```
epoberezkin: As I commented above, I think in this case "values" should not be allowed - not sure if it is possible though.

spec/types/jtd-schema.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    const union: JTDSchema<A | B> = {
+      discriminator: "type",
+      mapping: {
+        a: {properties: {a: {type: "float64"}}},
+        b: {
+          optionalProperties: {b: {type: "string"}},
+        },
+      },
+    }
```
epoberezkin, line 172: this is really cool

spec/types/jtd-schema.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+        a: {properties: {a: {type: "float64"}}},
+      },
+    }
+    should.exist(unionMissing)
+  })
+
+  it("should typecheck empty schemas", () => {
+    const empty: JTDSchema<Record<string, never>> = {}
+    should.exist(empty)
+    // probably shouldn't accept this
```
epoberezkin: yes - I agree

epoberezkin:
That's very cool. Some comments is just me not understanding things and maybe not possible to achieve - probably need to just read slower - but any comments would be great. Thank you!

epoberezkin:
It's also worth supporting readonly arrays - it was just added to JSONSchemaType.

lib/types/jtd-schema.ts
erikbrinkman: The answer ties to a lot of the choices I've made, and it's a design decision, one I'm open to having. In this case it has to do with how typescript handles `X extends Y ? ... : ...` when X is a union. In that case typescript distributes the conditional across every element of the union (see [distributive conditional types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#distributive-conditional-types)). In this case, the decision that I've made is that that the `JDTSchema<number | string>` shouldn't create a valid schema. However, if you write `T extends string ? string extends T ? MyType<T> : never : never` of `string | number` because of distributive conditional types, that will result in `string | never` which is just `string`. Thus, this solution uses this trick to assert that the type isn't a union with string as a member, but is specifically only the type string.

Yes, it is a little weird, and mostly has to do with the way they wrote the conditional type system. I think the general principle is that typescript is meant to mimic javascript, so you care more about extending (e.g. having the relevant properties) than you care about strict matching, e.g. only having the specific properties.

All extend is similar but instead is saying, every element of the union extends the type. This way `AllExtend<1 | 2 | 3, number>` is still true, and `AllExtend<true, bool>` is still true. This is "necessary" because JTD only defines enums for strings, not for numbers or bools, so we need to extend them. Arguably, `AllExtend` is an interesting design decision. I could see the argument that `JDTSchemaType` should reject that typing because JTD can't actually verify that type.

TypeEquality should work in place of IsString, so I can change that. Depending on the answer / discussion about the above question, they might all become type equality.

Finally, for simple types like this, the typescript playground is a good place to test out how things are resolving.

lib/types/jtd-schema.ts
erikbrinkman: this should. You can think of the first part as going through the definitions making the type `{def_name: { ref: def_name }}` only if T extends the type assigned to that definition. Then by indexing, we end up taking the union of all of those, so effectively it will only match the refs with appropriate type.

lib/types/jtd-schema.ts
erikbrinkman: That seems like a reasonable name, I called it `Nullable` because I'm kind of abusing the type and using the sentinel null to indicate anything that's not a string literal. This is necessary because we need to reject the type if there are any nulls present, i.e. it's not only a union of string literals. Still, I think EnumString convey this roughly and it's an internal name, so calling out the connection is good.

lib/types/jtd-schema.ts
erikbrinkman: To be clear with the way that AjvJTD works. `JTDSchemaType<Date>` should be `{ type: "timestamp" }`?

lib/types/jtd-schema.ts
erikbrinkman: I initially disagree, but you're correct. given what I think makes the most sense, this should only allow types of arbitrary keys.

lib/types/jtd-schema.ts
erikbrinkman: They do, I should change that, I ran out of letters. The more left D is meant to mean Discriminator. Basically that's the condition saying only take the type where {Discriminator: Mapping}, and the second D is just the Definitions type that's passed through.

Good catch!

lib/types/jtd-schema.ts
erikbrinkman: Given both the comment about rejecting in the test, and what I've written below, I think the only type that should allow an empty schema is the empty object `Record<string, never>` and maybe when it's implemented, the empty map `Map<string, never>`

spec/types/jtd-schema.spec.ts
erikbrinkman: Both eslint and typescript complained, and I wasn't sure of a good solution. If both are set to ignore underscores, then yeah, that makes much more sense.

lib/ajv.ts
erikbrinkman: This makes much more sense

erikbrinkman (commented):
I'm going to try and summarize some of the things I mentioned in some of the comments here, and may contradict some of what I wrote. Thanks for going over this so carefully. I think you ultimately caught some stuff I missed.

I want to make sure to hammer out this question of the types purpose to make sure we're on the same page. The first I expressed in one of the comments, but I'll express it again here. I ultimately this is the most important quality of the `JTDSchemaType` and that is that this should be true:
```
ajv.compile<T>(schema: JTDSchema<T>): (obj:unknown) => obj is T
```
This means that empty schemas should only be allowed for empty objects. values should only be allowed for types that are indexed by any string (contrary to what I wrote earlier), types like `1 | 2 | 3` should not be allowed as JTD can not properly verify them, etc.

The second property that I think is good, but parts can be done gradually, is that the type should attempt to prevent unexpected errors, e.g. the schema should try and match the actual type as much as possible. This means that while the having the schema `{ type: "string" }` for the type `number | string` is valid, i.e. if the validator returns true, then the object is that type, the schema was likely written improperly, and thus shouldn't be allowed. Similarly, if possible, definitions should only be allowed in the root, discriminator fields shouldn't be allowed in mapping objects, enumerated types should make sure that every element of the union is specified, etc. Most of these are somewhat hard to guarantee. I think in scope of this first commit is just union checking, but the others could be added later as they make sense.

- **definitions in nested schemas:** for now I'm going to not focus on disallowing this since it should fail fast and not provide unexpected errors. I'll look into see how difficult it will be / how complicated it will make the type, but I don't think it's that high priority.
- **dates:** I'm going to add support, shouldn't be hard.
- **metadata:** since this is mainly for migration purposes, and is complicated, and may not be able to typecheck, this seems out of scope for now. Also, the general notes around extensions seem to indicate that this should ideally be targeted at first class JTD users, not migrators.
- **some schema:** I think there's use for this, but I think it's somewhat out of scope. It seems like that may be better suited to being part of the vocabularies in some way, since internally they may want to use those types anyway.
- **union to tuple:** See [this issue](https://github.com/microsoft/TypeScript/issues/13298) for the long discussion about it. There seem to be some solutions that generate all permutations. In some sense, that's the ideal solution, but obviously without a compact way to represent all permutations, that type is not feasible for enumerations larger than a handful. This seems worth exploring, but as noted above, probably an extension.

Finally, what do you mean by readonly arrays, just `readonly T[]`? I think that should work fine, but I'll need to add tests to assert that.

lib/types/jtd-schema.ts
epoberezkin: Thanks a lot - got it - need to get into habit of using the playground indeed.

>  I could see the argument that JDTSchemaType should reject that typing because JTD can't actually verify that type.

Interesting. I think the current trade-off is better - to align JTD Schema with the existing type, rather than to restrict type to JTD schema capabilities (there is type code generation approach for the latter). We could add JTDSchemaStrict that would be stricter in that regard

lib/types/jtd-schema.ts
epoberezkin: NullableEnumString is also ok - up to you, just so it's clearer that it is related to enum.

lib/types/jtd-schema.ts
epoberezkin: I think so, yes. Need to test that it works - it only runs standard JTD tests atm.

lib/types/jtd-schema.ts
epoberezkin: great

spec/types/jtd-schema.spec.ts
epoberezkin: eslant is definitely set to ignore names that start from underscore, I "think" typescript would ignore them too.

epoberezkin:
> ultimately this is the most important quality of the JTDSchemaType and that is that this should be true:

yes

> This means that empty schemas should only be allowed for empty objects. values should only be allowed for types that are indexed by any string (contrary to what I wrote earlier)

yes

> types like 1 | 2 | 3 should not be allowed as JTD can not properly verify them

based on validation === type guard statement, yes, although in some comment I wrote that maybe there should be a separate strict version of this type, better to keep things simple - if application code does some additional validation further down it should "upgrade" the type (rather than have the same type).

agreed

> This means that while the having the schema { type: "string" } for the type number | string is valid, i.e. if the validator returns true, then the object is that type, the schema was likely written improperly, and thus shouldn't be allowed

yes

> Similarly, if possible, definitions should only be allowed in the root

if not too complex - ok

> discriminator fields shouldn't be allowed in mapping objects

yes, but not as important, as it would fail compilation - data type would still be correct.

> enumerated types should make sure that every element of the union is specified, etc.

I am in support of "sorted enum" requirement if this type is used

> Most of these are somewhat hard to guarantee. I think in scope of this first commit is just union checking, but the others could be added later as they make sense.

yes, let's ship it - it's good

> Finally, what do you mean by readonly arrays, just readonly T[]? I think that should work fine, but I'll need to add tests to assert that.

yes. see #1447 

agreed with all points, need to read "union to tuple" but ok too :)

> They do, I should change that, I ran out of letters. The more left D is meant to mean Discriminator. Basically that's the condition saying only take the type where {Discriminator: Mapping}, and the second D is just the Definitions type that's passed through.

That's what I thought. It's just scoping rules are not intuitive... I am not anti-words in type parameters :)

> Given both the comment about rejecting in the test, and what I've written below, I think the only type that should allow an empty schema is the empty object Record<string, never> and maybe when it's implemented, the empty map Map<string, never>

yes

erikbrinkman:
Everything should be addressed, check the tests to make sure things are as you expect. What's implemented here is essentially the JTDSchemaStrict. That is, it won't accept `1 | 2 | 3` etc. I still think this is best because it means the types will be correct after validating. I'm willing to loosen the restrictions, but I think it's important to think about what that would look like. There could be a flag passed `JTDSchemaType` for strict or not. Alternatively, there could be utility types that transform a strict a type into a loose type. One of the benefits of that approach is that you could do something like:
```
type Validatable = Loosen<MyType>
ajv.compile<Validatable>(schema)
```
which would allow you to know the type that was validated, but also assert afterwards, or just cast.

I'm definitely open to other suggestions too, but my general thinking was, if it starts strict, the type is easier to loosen than it is to make more strict and make previously passing schemas fail.

Other miscellaneous updates:
- restricting types (e.g. only root definitions) proved difficult enough that I'm definitely not including it here
- still looking into sorted tuples, but will probably submit as a separate pr
- dates and readonly arrays are supported
- I still couldn't find a way to get eslint or typescript to be okay with _variables. I switched it to just use the void operator to reduce the noise. If you think underscores should still be possible, do you think you could take a stab at getting them to work?
- some of the types have been cleaned up. Most of the _ types were removed in favor or better tricks, notably boxing unions in a singleton tuple to prevent conditional expansion.

erikbrinkman:
Also, looking into it more, it seems that there is no way to compare type literals in typescript, so there isn't a good way to do union to tuple 😢

epoberezkin:
> if it starts strict, the type is easier to loosen than it is to make more strict and make previously passing schemas fail.

agreed

looks good to me! ok to merge/release?

erikbrinkman:
Yeah, if the tests seem to cover everything, this should be good to squash and merge

epoberezkin:
@erikbrinkman I was wondering - if you have a recursive type that represents say a tree, then JTDSchemaType would probably require a recursive schema and it should all type check? Or would typescript just blow up? I didn't try it yet, but it probably should work?

Also, it would be great to add some advanced demos/examples in the tests to showcase what is possible.

erikbrinkman:
It seems like it should generally work, because the schema only needs to check one branch, e.g. as since the ref type checks, the rest works.

I tested it and it seems to work:
```
type LinkedList = { val: number, next: LinkedList } | null
const schema: JTDSchemaType<LinkedList, { node: LinkedList }> = {
    definitions: {
        node: {
            properties: {
                val: { type: "float64" },
                next: { ref: "node" }
            },
            nullable: true
        },
    },
    ref: "node"
}
```
typechecked fine in the playground.

I'm happy to add some demos / examples. Do you have a pointer to what you're thinking of?

Finally, this should be its own issue, but it's small, so I'll just mention it here. I missed one case, which is the type `unknown` should also match an empty schema. I'll add a PR to update that soon.

Also, do you want to open up an issue to discuss what the relaxed schema should look like? Do you want to wait until people try the current schema and raise particular issues?

epoberezkin:
> I'm happy to add some demos / examples. Do you have a pointer to what you're thinking of?

just some more relatable complex example combining multiple primitives in one schema. Recursive case would also be good (it was unexpected btw that you need to explicitly pass definitions as parameters - is it maybe avoidable?)

> what the relaxed schema should look like?

I don’t think it’s needed to be honest. It would encourage having one type where there should be two - one after JTD validation and another after additional application level validation (anything that cannot be expressed with JTD)

> unknown should also match an empty schema

indeed

erikbrinkman:
@epoberezkin my real question was where. Is there a place where you show some examples of JSONSchemaType that I could append with JTDSchemaType?

As for the question of needing to type the definitions, to me it makes sense. You current specify the type for the schema, why wouldn't you specify the type for the definitions, as they're basically mini schemas in and of themselves. If these go untyped, it becomes a little less clear to me how a recursive schema, or any complex definitions would work. They work now because you essentially specify the recursion, and typescript verifies that it's consistent. Without that I'm not sure.

It does bring up this idea though. Currently JTDSchemaType type takes a typescript type, and converts it to the type of an valid schema for that type. This is useful for error checking. However, another route would be to go the opposite direction, e.g. allow any schema type, and convert the schema type to a typescript type. Then you could just type `ajv.compile({ ...schema... })` and typescript could automatically infer the type of the output.

I think it should be pretty easy to do something like this, but part of the utility depends on the ability to customize the signature of compile (the other question you tagged me in). It'd be easier if we kind of assumed that it'd only be used for inferred types, and therefore wouldn't need to handle unions in the same way as a user specified type. The major problem I see is that a lot of the types won't get inferred correctly due to the way typescript works. If you write: `{ type: "float64" }` the inferred type of typescript is `{ type: string }` so the types would have to be very verbose as `{ type: "float64" as "float64" }` unless there's a new better way to type string literals.

I decided to play around with this, and it seems to generally work. See the playground [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAaghgGwJYBM7Aig+gHgMoA0UAIlBAB4YB2KAzlAEoQDGA9gE4o63DtJUBzIgFcqAayqsA7lQB8sqAF4oACgCwAKCjaoeMpQg16AbyihIALigAiAGYJW6AGwAWa1AC+UAPxQqwgFsAIwh2KAtNHV19ajooU3YIWysePkFPH1hEVHRMXGIAbTwC60TbawBdCqJiBQitHT0KWJMoMHZWSHZgJAhaKyY2Tm5efiEoUQlpOQzfUwKAaSh+KDEIEFZbXRL2ztCevsqKq3hkNAxsfB2OroPaSsXqkgUveqiqCAA3UM0ASigAD6qJoGIzxPzCBAIOBBBAQKy8YTQLy+fxQ8J+L6hX6aTTmaCnHIXfAKZQglrglBJfhIHqsKj9RgsDhcVJjETiSQyF6ZQnnPL4IjFaxU2w0ukMyp1LJnXKXQhMoas0aCIgfb7seS4jS2UTMCVQNgBMBIOEklS0ZgACwgATgVjwvysKlYQQAVlZJlyqP9FApXW7lvQ+XKSfFIjpPqxUFBLTa7QBuCPaRLAYTsKiqAOeznTJ1QANBmVEzBhv3gqMxgMJqCp9OZxEQGseJMaDzatgM4BQT7ZflKQ2sY2miAqYzJqCi8VIemM8cNKLaSRUqzzxfrto3fa9OcTjeL3sIVdmcDwmz2RzAVzuOD0OwOZxuTwEPf7nQfSjHspWazLiA3u8-3cdsFzfDwX1A-c0WhWEz0bKBbzMdgkQg-cQKicCJ2-GwgIQwDWCpaxNA8HENDxU8oAAGX4NYUGongB1MQ8rH8YJQjVZorGo8RMHo7svCBaDW00XUqH1GdMzhKgVH4MAuJo3ikB4fNWJCMI1x0OFuzYURu2UAAGVsoikK0R1UWSoAAQkUZRoP+DT1wAakcnSqGAIyNws5RZIAOg-dyJ3QnQ6wzQddNbEDNE7BjZJzKYZAHUQpw+FBWyQLYVEPEsZKoMBfnsidou7ViBwsgAyMrllyvzmigCqqrAGrKCa7t6t8-yWp8w8PO0IqoGQHhKMMAcpJyvKep7aMUCgApWKIAbgCGqgKmEtsgA). It looks like typescript is smart in that it handles infinite recursion, but not in a human readable way, e.g. it seems to lazily expand the type as necessary.

epoberezkin:
>  Is there a place where you show some examples of JSONSchemaType that I could append with JTDSchemaType?

I think it should be in "getting started" in readme, but I was thinking to add to the tests too, similar to json-schema types test.

It all deservers a separate big section in json-type-definition.md, particularly if data type is added (probably JTDDataType<typeof schema>?)

> allow any schema type, and convert the schema type to a typescript type.

That's the idea of JTD I think - then you don't really need to define the type.
Having a utility type would also help to define data types to be used in other contexts.

> It'd be easier if we kind of assumed that it'd only be used for inferred types, and therefore wouldn't need to handle unions in the same way as a user specified type

yes - sounds ok.

> { type: "float64" as "float64" }

You can make the whole schema `as const`, it is not too bad: https://www.typescriptlang.org/play?#code/MYewdgzgLgBBwAsCmBbAhjAvDA3gKBhiQBtUkwoIAuXAwmKATwAckaAiAM2JDSgDYALOzoBfPKJhoIMUJCh48TVjADKiVGgAqLJFga6QnOBvSK50E8nQBGGuuvbd+-IRJkK1WvQOsO3XigAZgAmEUJxcSA

It works in [your playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAaghgGwJYBM7Aig+gHgMoA0UAIlBAB4YB2KAzlAEoQDGA9gE4o63DtJUBzIgFcqAayqsA7lQB8sqAF4oACgCwAKCjaoeMpQg16AbyihIALigAiAGYJW6AGwAWa1AC+UAPxQqwgFsAIwh2KAtNHV19ajooU3YIWysePkFPH1hEVHRMXGIAbTwC60TbawBdCqJiBQitHT0KWJMoMHZWSHZgJAhaKyY2Tm5efiEoUQlpOQzfUwKAaSh+KDEIEFZbXRL2ztCevsqKq3hkNAxsfB2OroPaSsXqkgUveqiqCAA3UM0ASigAD6qJoGIzxPzCBAIOBBBAQKy8YTQLy+fxQ8J+L6hX6aTTmaCnHIXfAKZQglrglBJfhIHqsKj9RgsDhcVJjETiSQyF6ZQnnPL4IjFaxU2w0ukMyp1LJnXKXQhMoas0aCIgfb7seS4jS2UTMCVQNgBMBIOEklS0ZgACwgATgVjwvysKlYQQAVlZJlyqP9FApXW7lvQ+XKSfFIjpPqxUFBLTa7QBuCPaRLAYTsKiqAOeznTJ1QANBmVEzBhv3gqMxgMJqCp9OZxEQGseJMaDzatgM4BQT7ZflKQ2sY2miAqYzJqCi8VIemM8cNKLaSRUqzzxfrto3fa9OcTjeL3sIVdmcDwmz2RzAVzuDwEPf7nQfSjHspWazLiA3+-r2-fxdo6FYTPRs7wXRd2zA38wNfGwP2sTQvDgehOx4HENDxU8oAAGX4NYUBwngB1MQ8rH8YJQjVZorBw8RMAI7svCBADW00XUqH1GdMzhKgVH4MBqNwuikFQ0jAhCMI1x0OFuzYURu2UAAGVsoikK0R1UPioAAQkUZQAP+ST1wAaiM2SqGAZSN005Q+IAOifCyJwgqI6wzQc5NbCDNBQ7s+JzKYZAHUQpw+FBWyQLYVEPEteKoMBfgMicfIhAIB00gAydLlji+zmigTLsrAXLKGK7sCrshzStsw9LO0ZLkB4LDDAHbjYvi2qe2jFAoAKMiiAa4AmqoCoWLbIA) too

It's all good - it's worth adding it and it can be used it type signatures.

erikbrinkman:
I did not know about const assertions, that's really useful and undoes a lot of that frustration. This might not be too bad afterall.

epoberezkin:
It would be cool to have type for SomeJTSSchemaType to make sure the schema is valid (maybe it is just JSONSchemaType without parameters) and then JTDDataType that computes type for the data from the schema.

And actually using JSONSchemaType might also force the types to be correct without `as const`?
Although it probably won't work - it would be too generic...

So it would work like this:

```
const schema = { // I was thinking specifying type here would work, but probably not
    definitions: {
        node: {
            properties: {
                val: { type: "float64" },
                next: { ref: "node" }
            },
            nullable: true,
        }
    },
    ref: "node"
} as const

type MyData = JTDDataType<typeof schema>
```

Probably JTDDataType can make sure that the schema is valid and it removes the need for schema validation actually...

erikbrinkman:
yeah, I think with a fully defined JTDDataType, that should work to both validate and produce the correct type. The downside being that incorrect types will just be never instead of throwing an immediate error. There's some talk about adding an invalid type to make things like this more obvious, but nothing currently exists.

Issue #1447, opened 2021-02-16
Readonly arrays not handled properly


LinusU:
**What version of Ajv are you using? Does the issue happen if you use the latest version?** 7.1.0/master, yes

**Ajv options object**

```javascript
{}
```

**JSON Schema**

```json
{
  "additionalProperties": false,
  "properties": {
    "items": { "type": "array" },
  },
  "required": ["items"],
  "type": "object"
}
```

**Sample data**

```text
n/a
```

**Your code**

```typescript
interface TransactionInput {
  items: readonly { count: number, productId: string }[]
}

const transactionInputValidator = ajv.compile<TransactionInput>({
  additionalProperties: false,
  properties: {
    items: { type: 'array' }, // <--- Error: TS wants this to be "object"
  },
  required: ['items'],
  type: 'object'
})
```

**Validation result, data AFTER validation, error messages**

```text
n/a
```

**What results did you expect?**

TypeScript to allow it to be an array

**Are you going to resolve the issue?**

Yes

epoberezkin:
merged - thank you!

PR #1448, opened 2021-02-16
fix: allow readonly arrays in interfaces

LinusU:
Before this change, a readonly array would be inferred as "object":

```ts
interface TransactionInput {
  items: readonly { count: number, productId: string }[]
}

const transactionInputValidator = ajv.compile<TransactionInput>({
  additionalProperties: false,
  properties: {
    items: { type: 'array' }, // <--- Error: TS wants this to be "object"
  },
  required: ['items'],
  type: 'object'
})
```

Now this is properly inferred as "array".

-----

#1447

Changed `T extends any[]` to `T extends readonly any[]`. This works since `readonly T[]` is a subset of `T[]`, and thus `T[]` always satisfies `readonly T[]`.

The one word changed 😄

### DIFF
--- a/lib/types/json-schema.ts
+++ b/lib/types/json-schema.ts
@@ -38,7 +38,7 @@ export type JSONSchemaType<T, _partial extends boolean = false> = (T extends num
       } & {length: T["length"]}
       minItems: T["length"]
     } & ({maxItems: T["length"]} | {additionalItems: false})
-  : T extends any[]
+  : T extends readonly any[]
   ? {
       type: JSONType<"array", _partial>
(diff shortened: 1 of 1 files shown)


epoberezkin:
that's because of the conditional fall-through to objects - at first I thought there is something more weird going on :)

Thank you!

LinusU:
Thanks for the quick merge ❤️ 

When do you think this will be part of a release? I'm currently blocked on this 😅

epoberezkin:
today then ;)

epoberezkin:
https://github.com/ajv-validator/ajv/releases/tag/v7.1.1

PR #1455, opened 2021-02-21
add unknown => empty mapping in JTDSchemaType

erikbrinkman:
Makes sure that unknown types can be validated with an empty schema

Added the above change, and also some tests, notably a test that shows that typescript doesn't error when specifying the discriminator again in the schema (although I'm not sure why). And a test of a recursive schema.

should be pretty straightforward

### DIFF
--- a/lib/types/jtd-schema.ts
+++ b/lib/types/jtd-schema.ts
@@ -133,6 +133,8 @@ export type JTDSchemaType<T, D extends Record<string, unknown> = Record<string, 
       // NOTE there should only be one type that extends Record<string, never> so unions
       // shouldn't be a worry
       | (T extends Record<string, never> ? unknown : never)
+      // unknown also allows empty schema
+      | (unknown extends T ? unknown : never)
       // null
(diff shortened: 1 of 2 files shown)


epoberezkin (approved):

epoberezkin:
I had to revert it - I added JTDSchemaType to compile/validate type signatures and some minimal test for it, but this PR fails type inference in another place - [here](https://github.com/ajv-validator/ajv/blob/master/spec/errors.spec.ts#L530) - see [this error](https://github.com/ajv-validator/ajv/runs/1959426615?check_suite_focus=true). Need to look into it later or maybe you can...

Also, strangely "compile" works without explicit JSONSchemaType annotation ([see here](https://github.com/ajv-validator/ajv/blob/master/spec/types/json-schema.spec.ts#L122)) but JTDSchemaType required explicit type annotation ([see here](https://github.com/ajv-validator/ajv/blob/master/spec/types/jtd-schema.spec.ts#L33)) - any idea why?

epoberezkin:
cc @erikbrinkman ^^

PR #1457, opened 2021-02-22
Documentation for JTDSchemaType

erikbrinkman:

Adds some documentation to `JTDSchemaType`

See above

I'm not really sure how much detail you want. This seems like enough to me, but I'm definitely open. I'm not sure the usage of JTD in the readme is correct. Can you just import AJV or do you need to import AjvJTD? Or pass the `jtd: true` option? Can you import nested directories in typescript, (e.g. does `import AjvJTD from "ajv/dist/jtd";` work?) ?I've only used the types in deno.

Again, this is stacked, so ignore the first commit.

### DIFF
--- a/docs/json-type-definition.md
+++ b/docs/json-type-definition.md
@@ -325,14 +325,14 @@ Most straightforward types should work with `JTDSchemaType`, e.g.
 interface MyType {
     num: number;
     optionalStr?: string;
-    nullableEnum: "1" | "2";
+    nullableEnum: "v1.0" | "v1.2" | null;
     values: Record<string, number>;
 }
 
 const schema: JTDSchemaType<MyType> = {
     properties: {
         num: { type: "float64" },
-        nullableEnum: { enum: ["1", "2"], nullable: true },
+        nullableEnum: { enum: ["v1.0", "v1.2"], nullable: true },
         values: { values: { type: "int32" } },
     },
     optionalProperties: {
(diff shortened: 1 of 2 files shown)


docs/json-type-definition.md
```diff
@@ -313,6 +314,73 @@ Unlike JSON Schema, JTD does not allow to reference:
...
+The type `JTDSchemaType` can be used to validate that the written schema matches the type you expect to validate. This type is strict such that if typescript compiles, you should require no further type guards. The downside of this is that the types that `JTDSchemaType` can verify are limited to the types that JTD can verify. If a type doesn't verify, `JTDSchemaType` should resolve to `never`, throwing an error when you try to assign to it. This means that types like `1 | 2 | 3`, or general untagged unions (outside of unions of string literals) cannot be used with `JTDSchemaType`.
+
+### Most Schemas
+
+Most straightforward types should work with `JTDSchemaType`, e.g.
+```typescript
+interface MyType {
+    num: number;
+    optionalStr?: string;
+    nullableEnum: "1" | "2";
```
epoberezkin: should it be `nullableEnum: "1" | "2" | null`
Or am I missing something?

Also, I'd use string strings here, for people accidentally not to think you can use numbers...

epoberezkin (commented):
great - one comment there

erikbrinkman:
both of those made sense. Any comment on the typescript in the README?

epoberezkin:
> Any comment on the typescript in the README?

looks ok - merging it. Would be good to add example with compile, but let's understand first why it requires explicit type annotation (question in another PR)

PR #1459, opened 2021-02-22
Migrate to ES imports

erikbrinkman:
Removes the use of "require" in es code. This should allow using ajv with deno, and arguably makes the code cleaner by removing one dependency on common-js-like syntax.

This replaces almost all instances of require with `import * as <name> from "<path>"`. It's not clear to me why these weren't using es imports, and I could believe there is a reason so feel free to reject it. All tests passed on my local machine.

If there's some reason why this is a bad idea that I missed. I scanned for similar PRs / issues, but didn't find anything.

### DIFF
--- a/lib/ajv.ts
+++ b/lib/ajv.ts
@@ -31,7 +31,7 @@ export {_, str, stringify, nil, Name, Code, CodeGen, CodeGenOptions} from "./com
 import type {AnySchemaObject} from "./types"
 import AjvCore from "./core"
 import draft7Vocabularies from "./vocabularies/draft7"
-import draft7MetaSchema = require("./refs/json-schema-draft-07.json")
+import * as draft7MetaSchema from "./refs/json-schema-draft-07.json"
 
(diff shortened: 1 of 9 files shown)


epoberezkin:
It's not clear to me why these weren't using es imports

> because, embarrassingly, I thought that I can only import them with `import default` and that requires using esModuleInterop that, allegedly, forces downstream dependencies to also use this setting, and also generates some extra noise in JavaScript... I didn't know you can import them like that and require just worked :)

epoberezkin (approved):

erikbrinkman:
Yeah, the import syntax is a little confusing, but `import * as name from "path"` allows you to namespace an entire module.

As a side note, you've now approved a few of these PRs. Were you planning on merging them later? I don't have write access, so I can't merge them myself. I'm just trying to make sure you weren't waiting for me to do anything else.

epoberezkin:
> Were you planning on merging them later

Yes, getting there :)

PR #1460, opened 2021-02-22
types: expose SchemaValidateFunction

koba04:

I'm trying to update `ajv` from v6 to v7.
In that way, we had to define the type of a validation function locally even `ajv` has the type definition.
This PR allows us to use `SchemaValidateFunction` to define a validation function.

- https://ajv.js.org/docs/keywords.html#define-keyword-with-validation-function
- https://github.com/kintone/js-sdk/pull/636/files#diff-d5a911011e45cad491e7becc159811e9b989c609694dc91a5aea51c797f17d40R14-R18

I've exposed `SchemaValidateFunction` from the entry point of `ajv`.

### DIFF
--- a/lib/ajv.ts
+++ b/lib/ajv.ts
@@ -15,6 +15,7 @@ export {
   AnySchema,
   ValidateFunction,
   AsyncValidateFunction,
+  SchemaValidateFunction,
   ErrorObject,
   ErrorNoParams,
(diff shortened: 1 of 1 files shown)


epoberezkin (approved):

koba04:
Thank you!

PR #1472, opened 2021-03-02
feat: expose PropertiesSchema and RequiredMembers types

t7yang:
If we have a properties schema or required members that share across multiple schema then we would like to declare it separately. So, we need `PropertiesSchema` and `RequiredMembers` types to help us declare value.

Export type `PropertiesSchema` and type `RequiredMembers`.

No.


epoberezkin:
You may be better off with having them in definitions (but you still can type-check them using these types) - I will merge it.

@erikbrinkman added a very advanced type for JTDSchemaType with discriminated union support do check it out.

PR #1473, opened 2021-03-02
fix: Array type in JSONSchemaType should allow ReadonlyArray too

t7yang:
Currently developers can not use `ReadonlyArray` in `JSONSchemaType` like `enum`, `items` (for tuple), `required`, ...

This because in TypeScript, mutable array can assign to read-only array but read-only can not assign to mutable array.

[Example](https://www.typescriptlang.org/play?#code/DYUwLgBAZg9jBcEBKICGATGA7YBPAggE6Gq4A8AzmIQJZYDmAfANwBQrsMEAvBANoByAIwCANBAEAmAQF02rUJABGqQoiq0GfOexWEe-YWInSZEVBQgBjbFTYQIQA)

Convert every mutable array types to read-only array types in `lib/types/json-schema.ts`.

No.

### DIFF
--- a/lib/types/json-schema.ts
+++ b/lib/types/json-schema.ts
@@ -34,7 +34,7 @@ export type JSONSchemaType<T, _partial extends boolean = false> = (T extends num
       // JSON AnySchema for tuple
       type: JSONType<"array", _partial>
       items: {
-        [K in keyof T]-?: JSONSchemaType<T[K]> & Nullable<T[K]>
+        readonly [K in keyof T]-?: JSONSchemaType<T[K]> & Nullable<T[K]>
       } & {length: T["length"]}
       minItems: T["length"]
     } & ({maxItems: T["length"]} | {additionalItems: false})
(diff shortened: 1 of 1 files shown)


epoberezkin:
Just wondering - why do we need readonly arrays in schemas?

t7yang:
Said data like this:

```ts
enum Foo {
  a = 'A',
  b = 'B'
}

// FOO can be the value of `enum` in a schema, and also expose to outside world to use
// for some reasons, we freeze this tuple to prevent it from mutation
export const FOO = Object.freeze([Foo.a, Foo.b]);

interface Bar {
  foo: Foo;
}

// currently TypeScript complain about can not assign readonly array to mutable array
const schema: JSONSchemaType<Bar> = {
  type: 'object',
  properties: {
    foo: { type: 'string', enum: FOO }
  },
  required: ['foo']
}
```

We do have other alternatives like create a new mutable tuple for `enum` to bypass this issue, but kind of cumbersome work.

The main idea is to enhance the capability of `JSONSchemaType` to accept both mutable and read-only array.
I think this can make DX better :)

PR #1481, opened 2021-03-10
fix: reexport PropertiesSchema and RequiredMember types in ajv.ts

t7yang:
Last time PR https://github.com/ajv-validator/ajv/pull/1472 did not reexport this two types in `ajv.ts`.

Reexport `PropertiesSchema` and `RequiredMember` types in `ajv.ts`.

No.


epoberezkin:
I think it's not as widely needed as the main type, they can be imported directly from where they are defined:

```
import type {PropertiesSchema, RequiredMembers} from "ajv/dist/types/json-schema"
```

t7yang:
I think it is weird to import from dist, kind of workaround, but it do work.
So, I am okay, if you decide to not merge this PR.

PR #1483, opened 2021-03-11
Mark JSONSchemaType's Record<string,any> required property as optional

synaestheory:
This ensures that the required type can be omitted per: https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.5.3

The current json-schema-validation spec suggests that the `required` property can be omitted, behaving as if an empty array were provided.  This updated the type definition for that property to allow it to be omitted.

The relevant text from the spec is the last line in this section:
https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.5.3

It was introduced with Draft 6:
https://tools.ietf.org/html/draft-wright-json-schema-validation-01#section-6.17

I marked the `required` keyword as optional for JSONSchemaType's Record format and updated all references that I could find for relevant documentation and code comments. I also updated the required vocabulary definition.

There is a peculiar compilation error occuring while attempting to run the tests. The failure occurs with only the first commit (that updates the type using the optional property flag via typescript).

I have since attempted to also update the vocabulary for the `required` keyword but am still seeing the same type of compilation error and cannot quite make sense of how that change seemingly breaks the compilation (only while running tests).

### DIFF
--- a/docs/json-schema.md
+++ b/docs/json-schema.md
@@ -363,7 +363,11 @@ _invalid_: `{a: 1, b: 2, c: 3}`
 
 ### `required`
 
-The value of the keyword should be an array of unique strings. The data object to be valid should contain all properties with names equal to the elements in the keyword value.
+The value of this keyword MUST be an array. Elements of this array, if any, MUST be strings, and MUST be unique.
+
+An object instance is valid against this keyword if every item in the array is the name of a property in the instance.
+
+Omitting this keyword has the same behavior as an empty array.
 
 **Example**
 
--- a/lib/types/json-schema.ts
+++ b/lib/types/json-schema.ts
@@ -59,7 +59,7 @@ export type JSONSchemaType<T, _partial extends boolean = false> = (T extends num
       type: JSONType<"object", _partial>
       // "required" type does not guarantee that all required properties are listed
       // it only asserts that optional cannot be listed
-      required: _partial extends true ? Readonly<(keyof T)[]> : Readonly<RequiredMembers<T>[]>
+      required?: _partial extends true ? Readonly<(keyof T)[]> : Readonly<RequiredMembers<T>[]>
       additionalProperties?: boolean | JSONSchemaType<T[string]>
       unevaluatedProperties?: boolean | JSONSchemaType<T[string]>
       properties?: _partial extends true ? Partial<PropertiesSchema<T>> : PropertiesSchema<T>
--- a/lib/types/json-schema.ts
+++ b/lib/types/json-schema.ts
@@ -53,7 +53,6 @@ export type JSONSchemaType<T, _partial extends boolean = false> = (T extends num
   : T extends Record<string, any>
   ? {
       // JSON AnySchema for records and dictionaries
-      // "required" is not optional because it is often forgotten
       // "properties" are optional for more concise dictionary schemas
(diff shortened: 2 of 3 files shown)


synaestheory:
I do see the spec test failing after this change, the way it is failing is very peculiar to me and I do not currently have the time to look further into it. I will take a look another look a bit later, but if anyone has some quick insights that might help that would be appreciated.

epoberezkin:
This is by design.

The only scenario when you would want to omit `required` is when all your properties in the data type are optional (and probably there can be a way to check it), but this is relatively rare. The current workarounds for this scenario are:
- use `required: []` in your schema. The benefit is that it makes it explicit that all properties are optional (and not that the "required" was simply forgotten, which is quite a common mistake).
- do not use JSONSchemaType

If at least one of the properties in your data type is non-optional (=required) you have to use required keyword to have schema consistent with the data.

lib/types/json-schema.ts
```diff
@@ -53,13 +53,12 @@ export type JSONSchemaType<T, _partial extends boolean = false> = (T extends num
...
   ? {
       // JSON AnySchema for records and dictionaries
-      // "required" is not optional because it is often forgotten
       // "properties" are optional for more concise dictionary schemas
       // "patternProperties" and can be only used with interfaces that have string index
       type: JSONType<"object", _partial>
       // "required" type does not guarantee that all required properties are listed
       // it only asserts that optional cannot be listed
-      required: _partial extends true ? Readonly<(keyof T)[]> : Readonly<RequiredMembers<T>[]>
+      required?: _partial extends true ? Readonly<(keyof T)[]> : Readonly<RequiredMembers<T>[]>
```
epoberezkin, line 61: A possible improvement is to make required optional if RequiredMembers is an empty array, if RequiredMembers then required must be used in the schema.

docs/json-schema.md
```diff
@@ -363,7 +363,11 @@ _invalid_: `{a: 1, b: 2, c: 3}`
 
 ### `required`
 
-The value of the keyword should be an array of unique strings. The data object to be valid should contain all properties with names equal to the elements in the keyword value.
+The value of this keyword MUST be an array. Elements of this array, if any, MUST be strings, and MUST be unique.
+
+An object instance is valid against this keyword if every item in the array is the name of a property in the instance.
+
+Omitting this keyword has the same behavior as an empty array.
```
epoberezkin, line 370: No need to make this doc precise/formal - it purposefully tries to be more concise, not covering on the edge cases - there is I-D for formal specification.

lib/vocabularies/validation/required.ts
```diff
@@ -25,12 +25,13 @@ const def: CodeKeywordDefinition = {
...
   type: "object",
   schemaType: "array",
+  allowUndefined: true,
   $data: true,
   error,
   code(cxt: KeywordCxt) {
     const {gen, schema, schemaCode, data, $data, it} = cxt
     const {opts} = it
-    if (!$data && schema.length === 0) return
+    if (typeof schema === 'undefined' || (!$data && schema.length === 0)) return
```
epoberezkin, line 34: there is no need to check for undefined - this function would not be called if required is not present in the schema

epoberezkin:
closing it - the implementation to remove "required" in case of all properties optional will be different.

PR #1484, opened 2021-03-11
simplify some of the JTDSchema

erikbrinkman:

The JTD schema was a little complex, it still is, but arguably less so.

This puts everything behind a guard for that schema type, and then instead of having things be unions of everything, it does sequential checks

It should be relatively straightforward.


lib/types/jtd-schema.ts
```diff
@@ -28,11 +31,35 @@ type EnumString<T> = [T] extends [never]
...
+
+/** true if the type is a properties type */
+type IsProperties<T> = false extends IsUnion<Exclude<T, null>>
+  ? null extends EnumString<keyof Exclude<T, null>>
+    ? false
+    : true
+  : false
+
+/** true if type is discriminator type */
+type IsDiscriminator<T> = true extends IsUnion<Exclude<T, null>>
```
epoberezkin: It is possibly to further refactor into something like

```
IsRecord<T, Union extends boolean> = Union extends IsUnion<Exclude<T, null>>
   ? null extends EnumString<keyof Exclude<T, null>>
     ? false
     : true
   : false
```

(spent some time looking for the difference between the two:)

lib/types/jtd-schema.ts
```diff
@@ -48,94 +75,87 @@ export type JTDSchemaType<T, D extends Record<string, unknown> = Record<string,
...
+        : never
+      : // values
+      true extends IsValues<T>
+      ? T extends Record<string, infer V>
+        ? {
+            values: JTDSchemaType<V>
+          }
+        : never
+      : // properties
+      true extends IsProperties<T>
```
epoberezkin: ... and then if you like refactoring idea this would be `IsRecord<T, false>`...

lib/types/jtd-schema.ts
```diff
@@ -48,94 +75,87 @@ export type JTDSchemaType<T, D extends Record<string, unknown> = Record<string,
...
-      | (false extends IsUnion<Exclude<T, null>>
-          ? true extends TypeEquality<keyof Exclude<T, null>, string>
-            ? T extends Record<string, infer V>
-              ? {
-                  values: JTDSchemaType<V>
+              }) & {
+            additionalProperties?: boolean
+          }
+      : // discriminator
+      true extends IsDiscriminator<T>
```
epoberezkin: ... and this `IsRecord<T, true>`

Or they could stay separate, defined via `IsRecord`.

epoberezkin:
Great - let me know what you think - happy to merge as is or with refactoring.

epoberezkin (dismissed):

lib/types/jtd-schema.ts
erikbrinkman: yeah, I like the brevity!

erikbrinkman (commented):
yeah, I 'like the brevity. I'll make that change

epoberezkin (approved):

PR #1485, opened 2021-03-11
add union support to JSONSchemaType

### LINKED ISSUE #1302: Limitation: JSONSchemaType<T> does not support unions

ajv@7.0.0-beta.1

**Ajv options object**

N/A (type error)

**JSON Schema**

See code below

**Sample data**

N/A (type error)

**Your code**

```typescript
type VariableGroupValue = string | number | { query: string };

const variableGroupValueSchema: JSONSchemaType<VariableGroupValue> = {
  anyOf: [
    {
      type: 'string',
    },
    {
      type: 'number',
    },
    {
      type: 'object',
      properties: {
        query: {
          type: 'string',
        },
      },
      required: ['query'],
      additionalProperties: false,
    },
  ],
};
```

```
const variableGroupValueSchema: JSONSchemaType<VariableGroupValue, false>
Type '{ anyOf: ({ type: "string"; } | { type: "number"; } | { type: "object"; properties: { query: { type: "string"; }; }; required: "query"[]; additionalProperties: false; })[]; }' is not assignable to type 'JSONSchemaType<VariableGroupValue, false>'.
  Type '{ anyOf: ({ type: "string"; } | { type: "number"; } | { type: "object"; properties: { query: { type: "string"; }; }; required: "query"[]; additionalProperties: false; })[]; }' is not assignable to type '{ type: "object"; required: "query"[]; additionalProperties: boolean; properties?: PropertiesSchema<{ query: string; }> | undefined; patternProperties?: { [pattern: string]: never; } | undefined; propertyNames?: JSONSchemaType<...> | undefined; dependencies?: { ...; } | undefined; minProperties?: number | undefined;...'.
    Type '{ anyOf: ({ type: "string"; } | { type: "number"; } | { type: "object"; properties: { query: { type: "string"; }; }; required: "query"[]; additionalProperties: false; })[]; }' is missing the following properties from type '{ type: "object"; required: "query"[]; additionalProperties: boolean; properties?: PropertiesSchema<{ query: string; }> | undefined; patternProperties?: { [pattern: string]: never; } | undefined; propertyNames?: JSONSchemaType<...> | undefined; dependencies?: { ...; } | undefined; minProperties?: number | undefined;...': type, required, additionalPropertiests(2322)
Peek Problem (Alt+F8)
No quick fixes available
```

image

It complains that the top level schema object doesn't have the keys: `type`, `required`, `additionalPropertiests`. I thought I could use `anyOf` on it's own. It doesn't really make sense to use `type` at the top level because it could be one of three different types but it looks like it expects it to be an object. I also tried `oneOf` with the same result.

Any suggestions or workarounds welcome, especially if my schema is wrong. Thanks

Expected no type error 

No.

erikbrinkman:

fixes #1302

somewhat anyway

In addition to the full JSONSchemaType, it can not also be an empty `anyOf` or `oneOf` construct containing any array of valid schemas. I thought about supporting combinations, but I'm not sure how much sense that makes.

A typescript error moved positions in one of the tests, although where there error should throw doesn't seem super relevant.

This does add some weird nested possibilities, but these are all valid schemas.

It's important to note that this is very rudimentary, but it does open some typed schemas up.

Also, the changes to the body of JSONSchemaType besides the part at the end is just whitespace.

### DIFF
--- a/lib/types/json-schema.ts
+++ b/lib/types/json-schema.ts
@@ -85,12 +85,42 @@ export type JSONSchemaType<T, _partial extends boolean = false> = (
       else?: PartialSchema<T>
       not?: PartialSchema<T>
     })
+  // these two unions allow arbitrary unions of types
   | {
       anyOf: readonly JSONSchemaType<T, _partial>[]
     }
   | {
       oneOf: readonly JSONSchemaType<T, _partial>[]
     }
+  // this union allows for { type: (primitive)[] } style schemas
+  | ({
+      type: (T extends number
+        ? JSONType<"number" | "integer", _partial>
+        : T extends string
+        ? JSONType<"string", _partial>
+        : T extends boolean
+        ? JSONType<"boolean", _partial>
+        : never)[]
+    } & (T extends number
+      ? {
+          minimum?: number
+          maximum?: number
+          exclusiveMinimum?: number
+          exclusiveMaximum?: number
+          multipleOf?: number
+          format?: string
+        }
+      : T extends string
+      ? {
+          type: JSONType<"string", _partial>
+          minLength?: number
+          maxLength?: number
+          pattern?: string
+          format?: string
+        }
+      : T extends boolean
+      ? unknown
+      : never))
 ) & {
   [keyword: string]: any
   $id?: string
--- a/lib/types/json-schema.ts
+++ b/lib/types/json-schema.ts
@@ -7,75 +7,91 @@ type JSONType<T extends string, _partial extends boolean> = _partial extends tru
   ? T | undefined
   : T
 
-export type JSONSchemaType<T, _partial extends boolean = false> = (T extends number
-  ? {
-      type: JSONType<"number" | "integer", _partial>
-      minimum?: number
-      maximum?: number
-      exclusiveMinimum?: number
-      exclusiveMaximum?: number
-      multipleOf?: number
-      format?: string
+export type JSONSchemaType<T, _partial extends boolean = false> = (
+  | ((T extends number
+      ? {
+          type: JSONType<"number" | "integer", _partial>
+          minimum?: number
+          maximum?: number
+          exclusiveMinimum?: number
+          exclusiveMaximum?: number
+          multipleOf?: number
+          format?: string
+        }
+      : T extends string
+      ? {
+          type: JSONType<"string", _partial>
+          minLength?: number
+          maxLength?: number
+          pattern?: string
+          format?: string
+        }
+      : T extends boolean
+      ? {
+          type: "boolean"
+        }
+      : T extends [any, ...any[]]
+      ? {
+          // JSON AnySchema for tuple
+          type: JSONType<"array", _partial>
+          items: {
+            readonly [K in keyof T]-?: JSONSchemaType<T[K]> & Nullable<T[K]>
+          } & {length: T["length"]}
+          minItems: T["length"]
+        } & ({maxItems: T["length"]} | {additionalItems: false})
+      : T extends readonly any[]
+      ? {
+          type: JSONType<"array", _partial>
+          items: JSONSchemaType<T[0]>
+          contains?: PartialSchema<T[0]>
+          minItems?: number
+          maxItems?: number
+          minContains?: number
+          maxContains?: number
+          uniqueItems?: true
+          additionalItems?: never
+        }
+      : T extends Record<string, any>
+      ? {
+          // JSON AnySchema for records and dictionaries
+          // "required" is not optional because it is often forgotten
+          // "properties" are optional for more concise dictionary schemas
+          // "patternProperties" and can be only used with interfaces that have string index
+          type: JSONType<"object", _partial>
+          // "required" type does not guarantee that all required properties are listed
+          // it only asserts that optional cannot be listed
+          required: _partial extends true ? Readonly<(keyof T)[]> : Readonly<RequiredMembers<T>[]>
+          additionalProperties?: boolean | JSONSchemaType<T[string]>
+          unevaluatedProperties?: boolean | JSONSchemaType<T[string]>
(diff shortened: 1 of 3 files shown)


epoberezkin:
Took me some type to understand why it even works - I was stuck thinking that `T extends number`, would never match if T is `string | number` - but this works because conditional types distribute unions...

Interesting :)

>  I thought about supporting combinations, but I'm not sure how much sense that makes

did you mean to support, for example, that `JSONSchemaType<number | string>` is `{type: ["number", "string"]}`?
If so - yes - as I commented in the issue, not doing it would break type coercion...

Also - am I right that it does not require that all union members are "matched" by some anyOf element, but only that all elements in anyOf match some member in the union (and there can be some unmatched union members)? Or because of how conditional types distribute it would require that all union members are matched, although some can be matched more than once?

Either way, this is cool, this should go into v8 branch I just started - it is a breaking change - I am planning to release beta in a week or so and it will be the new major version by the end of March.

erikbrinkman:
yeah, distributive unions still throw me.

By combinations, I actually meant where some of the properties are specified, e.g. `{ type: "object", anyOf: [{ properties: ... }, { properties: ... }] }`.

I think I'm missing how supporting or not supporting that breaks coercion. Can you elaborate?

You're correct, it doesn't require they all be matched, and they could be matched several types. The type can be thought of as an array of schemas of each element. To ensure they were all matched is essentially the same problem as turning a string union into a tuple, which doesn't seem to be possible, at least neatly in current typescript.

epoberezkin:
> By combinations, I actually meant where some of the properties are specified, e.g. { type: "object", anyOf: [{ properties: ... }, { properties: ... }] }.

Yes, this is needed I think - it is very common with JSON schema to have common parts refactored (type or shared properties)... But maybe it is impossible

> I think I'm missing how supporting or not supporting that breaks coercion. Can you elaborate?

For example, you have these two schemas:

Schema1:

```javascript
{
  type: ["number", "string"],
  minimum: 0,
  minLength: 1
}
```

Schema 2:

```javascript
{
  anyOf: [
    {type: "number", minimum: 0},
    {type: "string", minLength: 1}
  ]
}
```

Without coercion they are equivalent. With coercion (`coerceTypes: true` option) if you validate, for example, -1, the first schema would fail it (and there would be no coercion), the second would pass it (and it would coerce it to string).

epoberezkin:
I am happy to merge as is btw, I will just add to the docs the limitations - it is still better than what there is now...

erikbrinkman:
I wasn't aware of the coerceTypes option. Let me make sure I understand though. JSONSchema allows irrelevant options to be in a schema? I don't think JSONSchemaType does... but I'm not certain.

From the discussion on #1302, I think Just adding the merger on primitive types might be enough? I need to rebase anyway, so I'll give that a shot.

erikbrinkman:
Added a few updates:

1. this now supports unions of primitives, in the way I vaguely described somewhere. The one downside is there's a little bit of copy pasta, but I'm not sure how to succinctly factor it out, at least in a satisfying way. The only way I can think would be to define an interface like `NumberProps` and `StringProps` to use twice, but I don't love that solution`.
2. it now allows the following checked `anyOf` / `oneOf` expressions: `$id`, `$ref`, `$defs`, `definitions`, and arbitrary keywords. If you think this is bad, it's not hard to undo
3. added some documentation at the end of the typescript section on unions.

The biggest no here is there's still no support for partial unions, but I still don't have any clue as to how to make that type check appropriately.

@epoberezkin to your earlier comment about coercion, currently either schema is now a valid, so it would be up to a users desired output.

epoberezkin:
That's great!

I am planning to release it in v8 (v8 branch) - it's already released as v8.0.0-beta.0 - just in case :)

> The only way I can think would be to define an interface like NumberProps and StringProps to use twice, but I don't love that solution`

Why don't you like it? Seems ok to me.
If you do, I'd call them NumberKeywords and StringKeywords - this is what they are, it probably would make the whole thing less bulky? Up to you.

>  about coercion, currently either schema is now a valid, so it would be up to a users desired output.

Yep, it's great!.

Thank you!

epoberezkin:
> JSON Schema allows irrelevant options to be in a schema?

Yes. JSONSchemaType would now allow it too in case of a simple union, but it is definitely not a good practice to mix them like this... Ajv by default warns on such union types as `type: [number, string]` (see strict-mode.md) - in most cases it is a bad idea - but there are cases when you do need them (like coercion or for better tooling), so this warning can be disabled with `allowUnionTypes: true` (or the whole strictTypes restriction can be disabled with `strictTypes: false`).

> I don't think JSONSchemaType does...

no, it is strict in not allowing any non-applicable or unknown members.

epoberezkin:
I tried to point it to v8 branch via GitHub UI, but it simply added all commits from v8 here... Probably v8 needs to be merged in. I will probably make one more release of v7 (with JTDSchema changes and other small things) and then merge v8 to master.

erikbrinkman:
I think I might be able to select the branch I'm using for the PR. I'll try to see if I can target and rebase onto v8, and consider merging the type copying.

epoberezkin:
That's ok, I will do it

lib/types/json-schema.ts
```diff
@@ -102,22 +108,9 @@ export type JSONSchemaType<T, _partial extends boolean = false> = (
...
-          maximum?: number
-          exclusiveMinimum?: number
-          exclusiveMaximum?: number
-          multipleOf?: number
-          format?: string
-        }
+      ? NumberKeywords
       : T extends string
-      ? {
-          type: JSONType<"string", _partial>
```
epoberezkin: Btw is this intentional that type: is removed? I was wondering before why it was needed...

erikbrinkman: lol, yeah, it was copy pasta, but redundant, so it wasn't caught. One of the benefits of going ahead with pulling the keywords out.

PR #1519, opened 2021-03-29
Add error path to strict mode tuple warning

### LINKED ISSUE #1509: strict mode tuple warning should give full path to location of problem
7.2.1

I have a large autogenerated schema (created using typescript-json-schema), and ajv's strict mode is giving warnings about it like:

```
strict mode: "items" is 2-tuple, but minItems or maxItems/additionalItems are not specified or different
```

There are lots of possible `"items"` in the schema! Which is it referring to?

Other strict mode warnings give the full path in the schema to the place where the issue is occurring, e.g.

```
strict mode: use allowUnionTypes to allow union type keyword at "#/anyOf/0/properties/workbench/properties/volume" (strictTypes)
```

It would be great if the tuple warning did the same!

Maybe?

asaid-0:
Fixes #1509 
Added error schema path to the warning message
No

### DIFF
--- a/lib/vocabularies/applicator/items.ts
+++ b/lib/vocabularies/applicator/items.ts
@@ -46,11 +46,12 @@ export function validateTuple(
   })
 
   function checkStrictTuple(sch: AnySchemaObject): void {
+    const {opts, errSchemaPath} = it
     const l = schArr.length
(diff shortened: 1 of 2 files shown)


asaid-0:
@epoberezkin could you please review this PR?

epoberezkin (approved):

epoberezkin:
Thank you!

Issue #940, opened 2019-02-09
Reduce ajv code / bundle size


epoberezkin:
**What version of Ajv you are you using?**
6.8.1

**What problem do you want to solve?**
Browser bundle size

**What do you think is the correct solution to problem?**
Move functions/constants from shared template definitions into separate files (or reused templates)

**Plan**

- [x] Approve adding ajv-dist to cdnjs: https://github.com/cdnjs/cdnjs/issues/13315
- [x] Publish ajv-dist to npm
- [x] Submit PR  to update cdnjs to pull ajv bundles from ajv-dist https://github.com/cdnjs/cdnjs/pull/13316
- [x] Update ~~travis~~ script to publish ajv-dist to npm on every tag - done with GitHub action
- [x] Merge PR to update cdnjs to pull ajv bundles from ajv-dist https://github.com/cdnjs/cdnjs/pull/13316
- [x] Deprecate bundle (/dist) in this package
- [x] ~~Update docs with the new bundle URLs (or maybe the bundle URLs can stay? TBC - they will be the same)~~
- [x] in ajv 7.0 - remove bundles (/dist) from this package

sindresorhus:
The install size is also bloated. Currently it's at 1.5 MB, which is a lot: https://packagephobia.now.sh/result?p=ajv

Users are complaining about it: https://github.com/sindresorhus/electron-store/issues/60

I'm not sure why `ajv` publishes both the lib and dist files, but that could probably be improved: https://github.com/epoberezkin/ajv/blob/d10720734d806fd25606dafd47b64e97ee7d1e7d/package.json#L8-L9

epoberezkin:
@sindresorhus cdnjs.com makes bundles available via npm distribution... Probably the answer is to publish ajv-dist to cdnjs rather than ajv. It probably should be done in the next major version.

epoberezkin:
See the TODO list above

sindresorhus:
Thanks for the quick response. Your plan sound good to me. 👌

epoberezkin:
v7-alpha has a smaller bundle size (120->100k minified)

epoberezkin:
Ok, this is mostly done - bundles are now pushed to ajv-dist on releases, and ajv-dist to npm via workflows - as soon as cdnjs/cdnjs#13316 is merged (or maybe I need to create a new one) the bundles will be in cdnjs.

epoberezkin:
The bundles for v7 are now available from cdnjs via ajv-dist

PR #1526, opened 2021-04-01
Smaller package

realityking:

Continuous improvement to #940 

While digging through the ajv source to look into the ESM creation, I noticed two opportunities to slim down the package size a little bit:

- SourceMaps are generated by Typescript. Since the code is not minified, I don't think this is a major aid in debugging (especially since Node.js, by default, doesn't take them into consideration when creating stack traces) they are big of dead weight. Saving: 268 KB
- The `scripts/` folder is included in the package. This was originally done in fe18ed52f89bc690838940228f2185bae9646df5 to ship the `compile-dots` script. Since that script no longer exists, it might be safe to no longer ship these. Saving: 7KB.

Together this would slash more than 25% of the package size (excluding dependencies).

An additional option could be to not ship the `lib/` folder (another 322 KB) but I'm not sure if some other ajv project might consume those files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
   "files": [
     "lib/",
     "dist/",
-    "scripts/",
     ".tonic_example.js"
   ],
   "scripts": {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,6 +7,7 @@
     "types": ["node"],
     "allowJs": true,
     "target": "ES2018",
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "sourceMap": false
   }
 }


epoberezkin:
Merged - thinking now :)

Would source maps not be used in browser bundles created by `bundle` script?

They are published to cdn via ajv-dist.

I suspect they are used - so rather than not generating them, they should be removed before publishing - e.g. could be in prepublish script.

epoberezkin:
Re lib - given that v8 is released less than a week ago - I think it's ok to drop it, ajv packages do not depend on anything there.

realityking:
> Would source maps not be used in browser bundles created by bundle script?

I double checked this. The script used for bundling currently only generates source maps in the minifaction step, there's no source map handling happening in the Browserify step. I got an idea how to improve yet + get rid of the source map in the npm package. I'll submit a PR in the next few days.

> Re lib - given that v8 is released less than a week ago - I think it's ok to drop it, ajv packages do not depend on anything there.

I saw that you tried it but it didn't work out - too bad. Maybe something to attempt in v9.

PR #1542, opened 2021-04-08
Added keyword name and schema path to keyword usage validation error

### LINKED ISSUE #1539: Add the keyword name to the reported error message in internal validation (validateKeywordUsage)
8.0.5

If the keyword is malformed according to the keyword schema definition, the current message that `ajv` throws is as follows (as an example):
```
 keyword value is invalid: data must be object
```

It's really annoying to try to track down _which_ specific keyword definition is malformed at this point, especially if we have multiple schemas precompiled one after another, and/or they are not overly simple.

Add the keyword name to the message

```
 keyword 'foo' value is invalid: data must be object
```

or

```
 keyword 'foo': keyword value is invalid: data must be object
```

Sure. For this particular change, we only have to change this:
https://github.com/ajv-validator/ajv/blob/b5642ea8655c3922de7f2403cc99d668871d35fb/lib/compile/validate/keyword.ts#L164

into:
```typescript
const msg = "keyword '" + keyword + "' value is invalid: " + self.errorsText(def.validateSchema.errors)
```

or

```typescript
const msg = "keyword '" + keyword + "': keyword value is invalid: " + self.errorsText(def.validateSchema.errors)
```

On the other hand, there might be more places that could use a little bit more verbose error reporting, so I'm open to extra suggestions.

pcwiek:
resolves #1539

Added keyword name & schema path to the reported error message during keyword usage validation.

I'm not entirely sure if we need that one extra test, but I added it anyway to be explicit and actually check the full message and not just parts of it.

### DIFF
--- a/lib/compile/validate/keyword.ts
+++ b/lib/compile/validate/keyword.ts
@@ -144,7 +144,7 @@ export function validSchemaType(
 }
 
 export function validateKeywordUsage(
-  {schema, opts, self}: SchemaObjCxt,
+  {schema, opts, self, errSchemaPath}: SchemaObjCxt,
   def: AddedKeywordDefinition,
   keyword: string
 ): void {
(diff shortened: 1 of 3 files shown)


pcwiek:
Ah, my bad. I have to re-run `prettier`

epoberezkin:
thank you!

PR #1564, opened 2021-04-24
Some improvements to JSONSchemaType for records

### LINKED ISSUE #1491: JSONSchemaType support for Record
**Version**
v8

**Problem**
Currently JSONSchemaType does not support types like `Record<string, T | undefined>` (or `{[K in string]?: T}`).

**Solution**

For the type for example:

```typescript
type MyRecord = Record<string, number | undefined>
```

JSONSchema should be:

```typescript
const myRecordSchema: JSONSchemaType<MyRecord> = {
  type: "object"
  additionalProperty: {type: "number"}
}
```

It should also allow `propertyName` keyword.

Possibly, below should also type check (although it can be expressed with `properties` but it would be more verbose:

```typescript
type MyEnumRecord = Record<"a" | "b" | "c" | "d", number | undefined>

const myEnumRecordSchema: JSONSchemaType<MyEnumRecord> = {
  type: "object"
  propertyNames: {enum: ["a", "b", "c", "d"]}, // it won't check for completeness
  additionalProperty: {type: "number"}
}
```

(Although it would require properly supporting enum, which currently is not type checked at all)

erikbrinkman:

fixes #1491

For `JSONSchemaType`: additional properties doesn't required `type: "string"` since it's
implied. `required` is no long required for objects without any required
members.

1. While I think that ultimately this makes the type support better for records, It does make the type definition itself a bit messier, so I'm open to the perspective that that messiness isn't worth the improvement
2. I chose to keep `required` required for partial types. I think that makes sense, but it's worth calling out.
3. The tests are only type tests, I can add schema validation tests too if you think it's necessary

### DIFF
--- a/lib/types/json-schema.ts
+++ b/lib/types/json-schema.ts
@@ -88,20 +88,24 @@ export type JSONSchemaType<T, _partial extends boolean = false> = (
           // "properties" are optional for more concise dictionary schemas
           // "patternProperties" and can be only used with interfaces that have string index
           type: JSONType<"object", _partial>
-          // "required" type does not guarantee that all required properties are listed
-          // it only asserts that optional cannot be listed
-          required: _partial extends true ? Readonly<(keyof T)[]> : Readonly<RequiredMembers<T>[]>
           additionalProperties?: boolean | JSONSchemaType<T[string]>
           unevaluatedProperties?: boolean | JSONSchemaType<T[string]>
           properties?: _partial extends true ? Partial<PropertiesSchema<T>> : PropertiesSchema<T>
           patternProperties?: {[Pattern in string]?: JSONSchemaType<T[string]>}
-          propertyNames?: JSONSchemaType<string>
+          propertyNames?: Omit<JSONSchemaType<string>, "type"> & {type?: "string"}
           dependencies?: {[K in keyof T]?: Readonly<(keyof T)[]> | PartialSchema<T>}
           dependentRequired?: {[K in keyof T]?: Readonly<(keyof T)[]>}
           dependentSchemas?: {[K in keyof T]?: PartialSchema<T>}
           minProperties?: number
           maxProperties?: number
-        }
+        } & (// "required" type does not guarantee that all required properties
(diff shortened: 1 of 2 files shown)


lib/types/json-schema.ts
```diff
@@ -88,20 +88,24 @@ export type JSONSchemaType<T, _partial extends boolean = false> = (
...
           dependentRequired?: {[K in keyof T]?: Readonly<(keyof T)[]>}
           dependentSchemas?: {[K in keyof T]?: PartialSchema<T>}
           minProperties?: number
           maxProperties?: number
-        }
+        } & (// "required" type does not guarantee that all required properties
+        // are listed it only asserts that optional cannot be listed.
+        // "required" is not necessary if it's a non-partial type with no required keys
+        _partial extends true
+          ? {required: Readonly<(keyof T)[]>}
```
epoberezkin, line 105: I am wondering - why "required" is required in partial schemas? I am probably missing something, as I write the example, it does not seem required (even when it has to be). This example type checks:

```typescript
type MyUnion = {tag: "a"; a: number} | {tag: "b"; b: number}

const myUnionSchema: JSONSchemaType<MyUnion> = {
  type: "object",
  required: ["tag"],
  oneOf: [
    {
      properties: {
        tag: {const: "a"},
        a: {type: "number"},
      },
    },
    {
      properties: {
        tag: {const: "b"},
        b: {type: "number"},
      },
    },
  ],
};
```

but it probably it should have `required: ["a"]` (or "b") in subschemas...

Either way, it's unrelated issue - I am going to merge it as is - thank you very much!

epoberezkin (approved):

lib/types/json-schema.ts
erikbrinkman: I think it doesn't matter because because partial schemas are just that, partial:

https://github.com/ajv-validator/ajv/blob/c9e0676094e04011b64187b2ec3b747a86198cd7/lib/types/json-schema.ts#L4

Issue #1521, opened 2021-03-29
JSONSchemaType requires "required" and "additionalProperties", unless it is oneOf schema


psteinroe:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**
8.0.1, yes

**Ajv options object**

```javascript
{}
```

**Your code**
```typescript
type Test = {
  some: string;
  other: boolean;
};

const t: JSONSchemaType<Test> = {
  type: 'object',
  properties: {
    other: {
      type: 'boolean',
    },
    some: {
      type: 'string',
    },
  },
};
```

**What results did you expect?**
Should be a valid type, instead the compiler throws

`TS2322: Type '{ type: "object"; properties: { other: { type: "boolean"; }; some: { type: "string"; }; }; }' is not assignable to type 'JSONSchemaType<test, false>'.   Type '{ type: "object"; properties: { other: { type: "boolean"; }; some: { type: "string"; }; }; }' is not assignable to type '{ oneOf: readonly JSONSchemaType<test, false>[]; } & { [keyword: string]: any; $id?: string; $ref?: string; $defs?: { [x: string]: JSONSchemaType<Known, true>; }; definitions?: { ...; }; }'.     Property 'oneOf' is missing in type '{ type: "object"; properties: { other: { type: "boolean"; }; some: { type: "string"; }; }; }' but required in type '{ oneOf: readonly JSONSchemaType<test, false>[]; }'`

epoberezkin:
I will check it - what typescript version do you use? Could you try upgrading to 4.2.3?

aMoniker:
I'm running into the same error, using typescript 4.2.3 and ajv 8.0.1 (same error in 8.0.0 as well)

psteinroe:
I upgraded to 4.2.3 and still have the error.

epoberezkin:
I understand the problem - this is not a bug, but with the addition of the union support error reporting became a bit more confusing - maybe something can be done about it (cc @erikbrinkman).

The minimal correct schema for this type is (and it compiles correctly):

```
const schema: JSONSchemaType<Test> = {
  type: "object",
  properties: {
    other: {
      type: "boolean",
    },
    some: {
      type: "string",
    },
  },
  required: ["other", "some"],
}
```

The reason for that is that the properties in the type are required, but JSON Schema without `required` keyword makes them optional.

Please note that this schema still allows additional properties.

[...]

PR #1583, opened 2021-05-01
Add checks for strict null checks

erikbrinkman:

It addresses some of the issues with #1521 by checking if strict null checks is enabled, and if not, setting all exported types from json-schema to a string literal indicating the problem.

1. created an internal wrapper type that checks if strictNullChecks is enabled. If so, resolves to the normal type, if not, resolves to a string literal type with an error.
2. renamed all external types to `UncheckedExternalType` and made internal, then added an external type that wraps it with the type above.
3. *minor* removed the default setting and the exposure of `_partial` from `JSONSchemaType`. This shouldn't have been used in the past as one could just use `PartialSchema` and it was arguably marked as private with the underscore name, but it did still used to be exposed.
4. *minor* changed `_partial` to `IsPartial` to conform with standard capitalization of types, which is now also less relevant as that parameter isn't exported.

- I added wrappers around everything that was exported, but potentially that's not necessary, or desired? It's not super clear, so this seemed like the safe option.
- The case could be made that removing `_partial` from `JSONSchemaType` is backwards compatibility breaking. If you're worried about that I can add it back, but ultimately think it should be removed
- I renamed everything `Unchecked...` that's pretty verbose and maybe not clear, so I'm open to other names.
- I've tested this in the typescript playground (the current version), but there are no built-in tests for it. I decided that adding a whole new type of tests that needed a custom tsconfig was overkill for this feature, but I'm open to the idea that it should be tested.

### DIFF
--- a/lib/types/json-schema.ts
+++ b/lib/types/json-schema.ts
@@ -1,9 +1,15 @@
 /* eslint-disable @typescript-eslint/no-empty-interface */
-export type SomeJSONSchema = JSONSchemaType<Known, true>
+type StrictNullChecksWrapper<Name extends string, Type> = undefined extends null
+  ? `strictNullChecks must be true in tsconfig to use ${Name}`
+  : Type
 
-export type PartialSchema<T> = Partial<JSONSchemaType<T, true>>
+export type SomeJSONSchema = UncheckedJSONSchemaType<Known, true>
 
-type JSONType<T extends string, _partial extends boolean> = _partial extends true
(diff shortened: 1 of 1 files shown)


epoberezkin:
> The case could be made that removing _partial from JSONSchemaType is backwards compatibility breaking. If you're worried about that I can add it back, but ultimately think it should be removed

I don't think anybody used it, it was effectively private, so it's cool that it is fully private now.

epoberezkin (approved):

PR #1584, opened 2021-05-03
JTD timestamp option

jrr:

https://github.com/ajv-validator/ajv/issues/1543

 - added `timestamp` configuration option, ~~available only under JTD~~
 - added tests for its behavior
 - adjusted timestamp check accordingly
 
- You might have a better idea of how to make different options available to the different flavors of AJV. (see the new `AjvOptions` type)
- I didn't see a great place to test this in an  existing suite (e.g. in the submodule JSON data) so I added a whole new test file.

### DIFF
--- a/lib/2019.ts
+++ b/lib/2019.ts
@@ -1,5 +1,5 @@
 import type {AnySchemaObject} from "./types"
-import AjvCore, {AjvOptions} from "./core"
+import AjvCore, {Options} from "./core"
 
 import draft7Vocabularies from "./vocabularies/draft7"
 import dynamicVocabulary from "./vocabularies/dynamic"
@@ -11,7 +11,7 @@ import addMetaSchema2019 from "./refs/json-schema-2019-09"
 const META_SCHEMA_ID = "https://json-schema.org/draft/2019-09/schema"
 
 class Ajv2019 extends AjvCore {
-  constructor(opts: AjvOptions = {}) {
+  constructor(opts: Options = {}) {
     super({
       ...opts,
       dynamicRef: true,
--- a/lib/2019.ts
+++ b/lib/2019.ts
@@ -1,5 +1,5 @@
 import type {AnySchemaObject} from "./types"
-import AjvCore, {Options} from "./core"
+import AjvCore, {AjvOptions} from "./core"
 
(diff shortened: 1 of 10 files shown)


lib/vocabularies/jtd/type.ts
```diff
@@ -41,7 +53,7 @@ const def: CodeKeywordDefinition = {
         break
       case "timestamp": {
         const vts = useFunc(gen, validTimestamp)
-        cond = _`${data} instanceof Date || (typeof ${data} == "string" && ${vts}(${data}))`
+        cond = timestampCode(data, vts, cxt.it.opts.timestamp)
```
epoberezkin: It would be better to pass ctx to timestampCode and only call `useFunc` in case it is needed (it adds generated code to create a variable in the scope of the generated function).

lib/2019.ts
```diff
@@ -1,5 +1,5 @@
 import type {AnySchemaObject} from "./types"
-import AjvCore, {Options} from "./core"
+import AjvCore, {AjvOptions} from "./core"
```
epoberezkin: Not sure what is the best way to deal with diverging options... It is a bit weird that the type that is used in constructor is different from the type that is exported (in the bottom)... Maybe it is better just to allow this option in all classes and ignore it, at least until the next major version?

As long as it is documented it is JTD only option it is ok I think.

spec/jtd-timestamps.spec.ts
```diff
@@ -0,0 +1,34 @@
+import _AjvJTD from "./ajv_jtd"
+import assert = require("assert")
+
+describe("JTD Timestamps", () => {
+  it("Accepts dates or strings by default", () => {
```
epoberezkin: normally it would be "should accept... "/"should enforce ..."

epoberezkin:
It's great - added some comments - thank you!

lib/2019.ts
jrr: Alright, I restored the old Options and documented `timestamp` in a few places.

epoberezkin (approved):

epoberezkin:
Thank you - will merge/release this week!

Issue #1586, opened 2021-05-04
JTDDataType doesn't recognize the boolean type


m00s:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**
Using `v8.2.0`, currently the latest version

**Your code**
I can reproduce the issue with the minimal example on the [documentation](https://ajv.js.org/guide/typescript.html#utility-type-for-jtd-data-type)
```typescript
import Ajv, {JTDDataType} from "ajv/dist/jtd"
const ajv = new Ajv()

const schema = {
  properties: {
    foo: {type: "boolean"}
  },
  optionalProperties: {
    bar: {type: "string"}
  }
} as const

type MyData = JTDDataType<typeof schema>

// type inference is not supported for JTDDataType yet
const validate = ajv.compile<MyData>(schema)

const validData: any = {
  foo: true,
  bar: "abc"
}

if (validate(validData)) {
  // validData.foo is of type `unknown` instead of `boolean`
  console.log(validData.foo)
} else {
  console.log(validate.errors)
}
```

**What results did you expect?**

I expect to have the type of `foo` as `boolean`, not as `unknown`

<img width="845" alt="Screen Shot 2021-05-04 at 11 42 04" src="https://user-images.githubusercontent.com/4352637/116986584-3e190680-acce-11eb-803f-9fbd7f9c8619.png">

**Are you going to resolve the issue?**

Yes

epoberezkin:
merged

PR #1587, opened 2021-05-04
Add boolean support to JTDDataType

m00s:

#1586 

- Add support for boolean type to JTDDataType
- Add test case

No


epoberezkin:
@m00s Thank you!
cc @erikbrinkman - looks like we missed it...

PR #1592, opened 2021-05-05
ci(build): add parallel test coverage calculation

Fdawgs:

At present, parallel jobs are run in the build workflow but only one coverage report is submitted to Coveralls.
As an example, compare the number of job reports in [this build of Fastify](https://coveralls.io/builds/39298323) with a report for an [Ajv build](https://coveralls.io/builds/39351320).

This PR resolves this by collecting all coverage from parallel jobs, allowing Coveralls to generate an aggregated score.
This will also allow developers to see where coverage is missing dependent on OS or node.js version.

- Added `coverage` job
- Added `parallel` and `flag-name` options (see supporting [documentation here](https://github.com/coverallsapp/github-action#inputs)) to `coveralls` step of `build` job

N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,3 +35,15 @@ jobs:
         uses: coverallsapp/github-action@master
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+          parallel: true
+          flag-name: run-${{ matrix.node-version }}-${{ matrix.os }}
+
+  coverage:
+    needs: build
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


epoberezkin:
There is no difference in which node version the code is executed - the coverage will be the same.

I don't think it's necessary to send multiple reports.

Fdawgs:
> 
> 
> There is no difference in which node version the code is executed - the coverage will be the same.
> 
> I don't think it's necessary to send multiple reports.

Cool, no worries @epoberezkin!

Issue #1585, opened 2021-05-04
Can't get version 8.1.0 to run in IE11 


mummybot:
This might be as simple as a one word answer: **"no"**, but I can't find anywhere explicitly stating it.

**The version of Ajv you are using**
8.0.1
**The environment you have the problem with**
Internet Explorer 11
**Your code (please make it as small as possible to reproduce the issue)**
```js
const options = {
  code: { es5: true },
};
const ajv = new Ajv(options);
ajv.compile(schema);
ajv.validate(formData);
```
It breaks on the [`class` declaration](https://github.com/ajv-validator/ajv/blob/master/lib/ajv.ts#L11).

I have tried transpiling it with babel:
```js
module: {
  rules: [
    {
      test: /node\_modules\/.*(ajv)\/.*\.js$/
      use: {
        loader: 'babel-loader',
        options: { presets: '@babel/preset-env' }
      }
    }
  ]
}
```
As well as adding `targets: { ie: '11' }`.
There are various subsequent errors, but eventually I got stuck on it not accepting the `pattern` key in a JSON schema: with or without a RexExp around the [compileSchema function](https://github.com/ajv-validator/ajv/blob/master/lib/compile/index.ts#L164).

epoberezkin:
It is probably because of unicode regexp flag that is required by JSON Schema spec - there was an option recently added to disable it: https://github.com/ajv-validator/ajv/pull/1530

mummybot:
That was exactly the issue! IE11 works fine with:
```js
const options = {
  unicodeRegExp: false,
  code: { es5: true },
};
const ajv = new Ajv(options);

// webpack.config.js
module: {
  rules: [
    {
      test: /node\_modules\/.*(ajv)\/.*\.js$/
      use: {
        loader: 'babel-loader',
        options: { presets: '@babel/preset-env' }
      }
    }
  ]
}
```

Thank you!

mummybot:
Out of curiosity, it may be worth adding this to the docs, as the es5 option gives the appearance of making it work, which it does so long as a schema contains no patterns... Do you want a PR?

epoberezkin:
Yes to docs - thank you.

mummybot:
Hope this is in the right place for you @epoberezkin

PR #1594, opened 2021-05-05
Ajv in Internet Explorer 11 FAQs documentation

mummybot:
Docs in the FAQs for getting Ajv to workin Internet Explorer 11 as discussed in #1585


docs/faq.md
```diff
@@ -96,3 +96,31 @@ There are two possible approaches:
...
+
+```js
+const options = {
+  unicodeRegExp: false, // this enables 'pattern' RegExp
+  code: { es5: true },  // uses es5 specific code branches within Ajv
+};
+const ajv = new Ajv(options);
+```
+
+and transpile Ajv within the `node_modules` of your project.
```
epoberezkin: isn't it better to recompile from typescript? This way you'd have correct source maps (probably)... Just thinking...

epoberezkin: Also, I'd probably move a bit more condensed version "to using in browser" section instead of here - I think it's more likely to be found there...

mummybot: I've condensed it, and moved it to environments.md in a warning (below DOJO, old skool).

epoberezkin:
It seems so

PR #1604, opened 2021-05-12
Update getting-started.md

venky19666:
const data is defined before access


epoberezkin:
constant must be defined before it is accessed, this change makes it invalid.

PR #1624, opened 2021-05-25
Update getting-started.md

ivansieder:
Corrected documentation

Textual change inside the Documentation

No


epoberezkin:
not sure if this clarification is needed - code "performance" is an approximate metric by definition.

epoberezkin:
ah - sorry - that's not what you corrected :)

epoberezkin:
yes - neither reads well tbh, mine is also grammatically incorrect :) Should I just drop approximately?

ivansieder:
Sadly my mother tongue isn't English, it's just what I've learned at school an what sounds right ^^ Translating it with Deepl results in something like "If your JSON string is valid, specialized parsing is approximately as fast as JSON.parse"

PR #1626, opened 2021-05-25
Track evaluated properties with pattern properties of schema = true

P0lip:

https://github.com/ajv-validator/ajv/issues/1625

`patternProperties` that are always valid are tracked now to ensure `unevaluatedProperties` keyword 'works'.

`checkForUnevaluatedProperties` check?

### DIFF
--- a/lib/vocabularies/applicator/patternProperties.ts
+++ b/lib/vocabularies/applicator/patternProperties.ts
@@ -1,9 +1,10 @@
 import type {CodeKeywordDefinition} from "../../types"
 import type {KeywordCxt} from "../../compile/validate"
-import {schemaProperties, usePattern} from "../code"
+import {allSchemaProperties, usePattern} from "../code"
 import {_, not, Name} from "../../compile/codegen"
-import {checkStrictMode} from "../../compile/util"
+import {alwaysValidSchema, checkStrictMode} from "../../compile/util"
 import {evaluatedPropsToName, Type} from "../../compile/util"
+import {AnySchema} from "../../types"
 
 const def: CodeKeywordDefinition = {
   keyword: "patternProperties",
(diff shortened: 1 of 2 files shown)


lib/vocabularies/applicator/patternProperties.ts
```diff
@@ -12,9 +13,16 @@ const def: CodeKeywordDefinition = {
...
     const {gen, schema, data, parentSchema, it} = cxt
     const {opts} = it
-    const patterns = schemaProperties(it, schema)
-    // TODO mark properties matching patterns with always valid schemas as evaluated
-    if (patterns.length === 0) return
+    const patterns = allSchemaProperties(schema)
+    const validPatterns = patterns.filter((p) => !alwaysValidSchema(it, schema[p] as AnySchema))
+    const checkForUnevaluatedProperties =
+      parentSchema.unevaluatedProperties !== true &&
+      parentSchema.unevaluatedProperties !== undefined
```
epoberezkin, line 20: I don't think you can limit it to checking parent - we should mark unevaluatedProperties regardless - it can be higher up in the schema...

lib/vocabularies/applicator/patternProperties.ts
```diff
@@ -12,9 +13,16 @@ const def: CodeKeywordDefinition = {
   code(cxt: KeywordCxt) {
     const {gen, schema, data, parentSchema, it} = cxt
     const {opts} = it
-    const patterns = schemaProperties(it, schema)
-    // TODO mark properties matching patterns with always valid schemas as evaluated
-    if (patterns.length === 0) return
+    const patterns = allSchemaProperties(schema)
+    const validPatterns = patterns.filter((p) => !alwaysValidSchema(it, schema[p] as AnySchema))
```
epoberezkin, line 17: Multiple negation is confusing. `validPatterns` contains the patterns whose schemas are NOT always valid...

lib/vocabularies/applicator/patternProperties.ts
```diff
@@ -51,18 +59,22 @@ const def: CodeKeywordDefinition = {
...
-            {
-              keyword: "patternProperties",
-              schemaProp: pat,
-              dataProp: key,
-              dataPropType: Type.Str,
-            },
-            valid
-          )
+          const alwaysValid = !validPatterns.includes(pat)
+          if (!alwaysValid) {
```
epoberezkin, line 63: ... and here it's double negated again...

It would probably be better to have the list of patters that have alwaysValidSchemas and simply check that the pattern is not included there.

epoberezkin (changes requested):

epoberezkin:
I will update it

epoberezkin:
@P0lip the updated condition to bypass validation is:

```typescript
if (
  patterns.length === 0 ||
  (alwaysValidPatterns.length === patterns.length &&
    (!it.opts.unevaluated || it.props === true))
) {
  return
}
```

to spell it, we want to skip validation in one of two cases:
- there are no patterns in the schema at all
- all patterns have always valid schemas AND we either do not track unevaluated properties (`!it.opts.unevaluated`) or all properties are already evaluated (`it.props === true`)

Does it make sense?

P0lip:
Yup, make sense,
Thanks!

lib/vocabularies/applicator/patternProperties.ts
P0lip: True, I agree it may lead to slight confusion.

PR #1641, opened 2021-06-07
Repair example code for standalone config

SirLich:

Unclear documentation, which caused me some confusion using the stand-alone configuration, as I am still learning the difference between require, and es6-style imports. 

A single line of documentation, which should make using standalone configuration a little easier. 

No thank you. I hope its OK to make such a small PR. Thanks for the great library.


erikbrinkman (approved):
It looks like this happened with the change to babel6, and was completely missed. Thanks for the fix!

Issue #1655, opened 2021-06-18
tsc --build fails when AnyValidateFunction<unknown> is used and esModuleInterop is enabled


root-talis:
I just started using Ajv and ran into a problem when compiling my project. I've set [esModuleInterop option](https://www.typescriptlang.org/tsconfig#esModuleInterop) of TypeScript compiler to `true` and got a list of weird-looking error messages pointing into the internals of ajv. What's even more strange, the problem appears only if `AnyValidateFunction<unknown>` typehint is used in my code. Disabling `esModuleInterop` also "fixes" the problem, but then my project doesn't work. Also, keeping this option enabled [seems to be recommended](https://github.com/tsconfig/bases/blob/main/bases/node16.json).

I think I've come up with a solution, but I'm not sure if it won't break anything. Please review the issue and the proposed fix.

Contents
===
1. Issue Description
2. Proposed fix

## 1. Issue Description

**The version of Ajv you are using**:
version 8.6.0

**The environment you have the problem with**:
node v16.3.0, typescript v4.3.4, `tsc --build` with `"esModuleInterop": true`

**Your code (please make it as small as possible to reproduce the issue)**

_package.json:_
```
{
  "dependencies": {
    "ajv": "^8.6.0"
  },
  "devDependencies": {
    "typescript": "^4.3.4"
  }
}
```

_tsconfig.json:_
```
{
  "compilerOptions": {
    "rootDir": "./",
    "outDir": "./build",

    "lib": ["es2020"],
    "module": "commonjs",
    "target": "es2020",

    "strict": true,
    "esModuleInterop": true,
    "resolveJsonModule": true
  }
}
```

_app.ts:_
```
import {AnyValidateFunction} from "ajv/lib/types/index";
function Foo(): AnyValidateFunction<unknown>|null { return null }
Foo()
```

**Results and error messages in your platform**
```
~/foo$ yarn run tsc --build
yarn run v1.22.5
$ tsc --build
node_modules/ajv/lib/compile/resolve.ts:101:3 - error TS2349: This expression is not callable.
  Type '{ default: { (schema: SchemaObject, opts: Options, cb?: Callback | undefined): void; (schema: SchemaObject, cb: Callback): void; }; }' has no call signatures.

101   traverse(schema, {allKeys: true}, (sch, jsonPtr, _, parentJsonPtr) => {
      ~~~~~~~~

  node_modules/ajv/lib/compile/resolve.ts:5:1
[...]

PR #1656, opened 2021-06-18
Fixed esModuleInterop compatibility error

root-talis:
Issue #1655.

Fixed imports in `lib/compile/resolve.ts`.

I'm not sure how this can be covered in tests, so no tests supplied. Also, I'm not sure that this fix will not break any other usecases.

### DIFF
--- a/lib/compile/resolve.ts
+++ b/lib/compile/resolve.ts
@@ -1,9 +1,9 @@
 import type {AnySchema, AnySchemaObject} from "../types"
 import type Ajv from "../ajv"
 import {eachItem} from "./util"
-import * as equal from "fast-deep-equal"
-import * as traverse from "json-schema-traverse"
-import * as URI from "uri-js"
+import equal from "fast-deep-equal"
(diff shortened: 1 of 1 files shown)


heartforit:
I fell on this as well. This is very annoying. It is not in the master branch, as you can see here for yourself:
https://github.com/ajv-validator/ajv/blob/master/lib/compile/resolve.ts
Maybe somone reversed it.

heartforit:
I opened pull-request:
https://github.com/heartforit/ajv/tree/heartforit-patch-1

heartforit:
Adding  `"ajv": "git+https://github.com/heartforit/ajv.git"` as dependency in package.json could fix it for now.

smelamud:
I've also stuck on this error. I don't know, why this PR was closed, this fix is really needed. I've just reproduced this bug in my Node.js project, while the same ajv version in a React project compiles OK. Possibly, because the React project uses webpack.

Issue #1683, opened 2021-07-09
Adding an option to change the regex engine to RE2


efebarlas:
**What problem do you want to solve?**
The regex engine of Node.js uses a backtracking algorithm which may result in slow regex validation times for schemas containing unsafe regexes, possibly resulting in ReDoS if the web service waits for the validation to be completed.
**What do you think is the correct solution to problem?**
We can add an option to Ajv to determine the regex engine to use (RE2 or Node.js). RE2 is a regex engine which guarantees worst-case linear time complexity (although it runs slower than a backtracking algorithm for some regexes). It's best if this option would default to Node.js for reasons I will detail below.

The vast majority of regexes behave the same in RE2 as in Node.js (the RE2 test suite works by performing regression testing with a traditional backtracking algorithm and a variety of regex engine flavors). However, there are known, and acknowledged, mismatches between RE2 and Node.js, for example the inclusion of vertical tabs as a whitespace character. There could be other, unknown mismatches as well. It is known that different regex engines may interpret regexes slightly differently, as detailed in [this post](https://davisjam.medium.com/why-arent-regexes-a-lingua-franca-esecfse19-a36348df3a2).

Then, a developer could opt-in to using RE2 if they're sure their regex works correctly in RE2. A try-catch block would try to compile the regex in RE2. If it doesn't - meaning the regex contains extended features such as backreferences, which RE2 cannot support -, we can fallback on the Node.js engine and use self.logger.log to display a brief warning about the fallback.
**Will you be able to implement it?**
Yes.

There are other options for integrating RE2, and I can explain them if this solution isn't found suitable.

Also related:  ajv-validator/ajv-formats#7

epoberezkin:
See https://github.com/ajv-validator/ajv/pull/1684#issuecomment-880945417

The option `regExp` would have type:

```
type RegExpEngine = (pattern: string) => RegExpLike

interface RegExpLike {
  test: (s: string) => boolean
}
```

efebarlas:
Should the interface support a 'test' function or should it be something like this:
```
interface RegExpLike { 
  compile: (s: string) => boolean
  match: (s: string) => boolean 
}
```

 This would allow for separation of regex compilation in ajv.compile from regex validation in ajv.validate, right?

epoberezkin:
Needs to be double checked but I don’t think Ajv uses anything other than test at the moment

efebarlas:
[...]

PR #1684, opened 2021-07-09
Option to use RE2 (defaults on native regex engine)

efebarlas:
Signed-off-by: Efe Barlas <ebarlas@purdue.edu>

#1683
-I added an option on using RE2 or the native regex engine. This option defaults on the native regex engine.
-I added a dependency to node-re2. The unpacked size of node-re2 is 1.26 MB, but I do not know the minified size.
-I changed the 'usePattern' function to read the option, and to try and compile the regex in RE2. If RE2 throws an error, the system produces a log message, and falls back to the native regex engine.
-I haven't added tests which demonstrates the correctness of regex validation when RE2 is used. I'm waiting for a green light from the maintainers of ajv to do so.
Answer to [this comment](https://github.com/ajv-validator/ajv/pull/1684#issuecomment-877357736):
-When RE2 is enabled, some test cases in json-schema-test fail, because RE2 and ECMA 262, section 15.10.1 differ in their definition of whitespace (JS includes \v, the other does not).


lib/core.ts
```diff
@@ -219,7 +220,8 @@ type RequiredInstanceOptions = {
     | "validateSchema"
     | "validateFormats"
     | "int32range"
-    | "unicodeRegExp"]: NonNullable<Options[K]>
+    | "unicodeRegExp"
+    | "useRe2"]: NonNullable<Options[K]>
```
davisjam: FWIW, I think this engine is usually spelled "RE2", not "Re2".

lib/vocabularies/code.ts
```diff
@@ -90,8 +92,23 @@ export function callValidateCode(
...
+    try {
+      const engine = new Re2(pattern)
+      return gen.scopeValue("pattern", {
+        key: pattern,
+        ref: engine,
+        code: _`new ${useFunc(gen, re2)}(${pattern})`,
+      })
+    } catch (e) {
+      self.logger.log(
+        "Warning: Fallback to default regex engine. Please verify the safeness of your regex"
```
davisjam: If this is left in, it could be clearer: "Warning: Your regex is not supported by RE2. Falling back to ..."

lib/vocabularies/code.ts
```diff
@@ -90,8 +92,23 @@ export function callValidateCode(
   return context !== nil ? _`${func}.call(${context}, ${args})` : _`${func}(${args})`
 }
 
-export function usePattern({gen, it: {opts}}: KeywordCxt, pattern: string): Name {
+export function usePattern({gen, it: {self, opts}}: KeywordCxt, pattern: string): Name {
   const u = opts.unicodeRegExp ? "u" : ""
+  const useRe2 = opts.useRe2
+  if (u === "u" && useRe2) {
```
davisjam: Can you comment on the reason to restrict the RE2 protection to only the unicodeRegExp?

davisjam:
Questions and notes:

- If RE2 is used as the default engine, does the test suite pass?
- If merged, updates should be made to the documentation on the AJV website that describes ReDoS concerns, to document this option.

lib/vocabularies/code.ts
efebarlas: RE2 always parses regexes as Unicode. The unicodeRegExp option is true by default, which means a developer must knowingly change that option. We could use RE2 nevertheless, but in those cases, RE2 will most likely not behave as the developer anticipates. 

In that case, we can convert a non-Unicode regex to be interpreted in Unicode, but I do not know how to implement this as of now.

lib/core.ts
efebarlas: useRe2 changed to useRE2

lib/vocabularies/code.ts
efebarlas: Warning message changed to be more explanatory

epoberezkin:
I am not terribly excited about the idea of making node-re2 the dependency... I understand the problem though.

A better approach could be to make an option `regExp` to pass a function that given a string returns an object that has the same methods as RegExp object; this option would default to `(s) => opts.unicodeRegExp ? new RegExp(s, "u") : new RegExp(s)`. It would also simplify the code in usePattern a bit - `opts.unicodeRegExp` would only be checked only if `regExp` option is not passed and only at instance construction.

Some other comments in code

lib/vocabularies/code.ts
```diff
@@ -90,8 +92,23 @@ export function callValidateCode(
   return context !== nil ? _`${func}.call(${context}, ${args})` : _`${func}(${args})`
 }
 
-export function usePattern({gen, it: {opts}}: KeywordCxt, pattern: string): Name {
+export function usePattern({gen, it: {self, opts}}: KeywordCxt, pattern: string): Name {
   const u = opts.unicodeRegExp ? "u" : ""
+  const {useRE2} = opts
+  if (u === "u" && useRE2) {
+    try {
+      const engine = new RE2(pattern)
```
epoberezkin: it's probably incorrect to call it "engine", RE2 is an engine, this one is probably just "regExp" or something...

lib/vocabularies/code.ts
```diff
@@ -90,8 +92,23 @@ export function callValidateCode(
...
+      return gen.scopeValue("pattern", {
+        key: pattern,
+        ref: engine,
+        code: _`new ${useFunc(gen, re2)}(${pattern})`,
+      })
+    } catch (e) {
+      self.logger.log(
+        "Warning: One of the regexes in the schema is not supported by RE2. Falling back to the native regex engine"
```
epoberezkin: this fallback would be just handled in the passed function

epoberezkin:
@efebarlas thanks - getting there! Possibly, having re2.ts file might create the expectation that re2 doesn't have to be imported separately... I guess it can be all covered in the docs.

docs/options.md
```diff
@@ -361,6 +362,11 @@ type CodeOptions = {
   // Code snippet created with `_` tagged template literal that contains all format definitions,
   // it can be the code of actual definitions or `require` call:
   // _`require("./my-formats")`
+  regExp: RegExpEngine
+  // Developers looking for a ReDoS mitigation may wish to use a DFA regex engine,
+  // such as node-re2. During validation of a schema, code.regExp will be 
+  // used to match strings against regexes. The supplied object must support
```
epoberezkin:
```suggestion
  // used to match strings against regexes. The supplied function must support 
```

lib/core.ts
```diff
@@ -134,16 +134,24 @@ export interface CurrentOptions {
...
+}
+
 export interface CodeOptions {
   es5?: boolean
   lines?: boolean
   optimize?: boolean | number
   formats?: Code // code to require (or construct) map of available formats - for standalone code
   source?: boolean
   process?: (code: string, schema?: SchemaEnv) => string
+  regExp?: RegExpEngine & {code: string} // add type
```
epoberezkin, line 150: I think code: string could be added to the type itself and to the docs. Possibly as optional, if typescript doesn’t not complain - it’s only needed for standalone code generation

epoberezkin (commented):
Cool - looks good - I can finalise

epoberezkin:
merged

marcospassos:
@epoberezkin and @efebarlas, passing a custom engine will also affect the validation of `format: 'regex'`, right? I couldn't find any information about this in the docs.

efebarlas:
After some experimenting, I don't think the validation of format: 'regex' is affected by changing the engine, because it seems that regexes with features RE2 doesn't support, such as (?=.*@.*), pass validation when RE2 is used. I think with formats, we just use the default regex engine.

I might work on a patch for this at some point, but I'm unfortunately busy at the moment.

epoberezkin:
Thank you!

reed-lawrence:
What is the `RegExtEngine.code` field intended to be used with? The following code is currently throwing an error:

```typescript
import Ajv from 'ajv';
import RE2 from 're2';

const ajv = new Ajv({ code: { regExp: RE2 } });
```

`Property 'code' is missing in type 'RE2Constructor' but required in type 'RegExpEngine'.`

davisjam:
@reed-lawrence , can you clarify whether you are asking "What is the purpose of this feature?" or something else?

reed-lawrence:
@davisjam Thanks for the reply. 

I guess I am wondering if the declared interface is correct or how to reconcile that the code sample in the [documentation for ReDoS attack mitigation](https://ajv.js.org/security.html#redos-attack) does not build. What would be the proper integration of RE2 given the error posted in my prior [comment](https://github.com/ajv-validator/ajv/pull/1684#issuecomment-1734113685)?

Is the `.code` field required or optional? Given that the `re2` library does not implement it. If the `.code` field is required, what is the proper implementation of an external regex engine such as `re2`

PR #1699, opened 2021-07-18
Update JSONSchemaType

erikbrinkman:

Various inconsistencies / undesired behavior with `JSONSchemaType`. When looking at other issues I was pouring over this class I notices these and combined them into this commit. Some aspects of this change behavior, but not in any way that I think is backwards incompatible, it only makes the type more permissive. That said, it's also reasonable to wait for a breaking release tp update these.

- Fixed the handling of `type: [...]` style unions. It wouldn't work for readonly array (e.g. const interfaces)
- Fixed a bug due to taking the intersection not the union of all the sub interfaces. There was a test for that that passed, and I'm not sure why, but it could have been related to the other bug where `unknown` clobbers the intersection, e.g. `unknown | { maximum: number}` greedily becomes `unknown`, `{} | {maximum: number}` is preserved, which then becomes intersected, and then optimized to the proper `{ maximum: number }`.
- Fixed where you couldn't omit `type: "boolean"` in boolean partial schemas
- Fixed where type schema wouldn't validate readonly tuples as tuples
- Allow type of `null` to be set property. Previously the type would allow an interface like `_: JSONSchemaType<null> = { nullable: true }`, but that's not a valid schema, it needs a type. It also allows you to specify null as a const. This was explicitly forbidden, and I agree it's mostly useless, but it still okay to accept.
- Changed some `{[Key in string]: ...}` to `Record<string, ...>` which has slightly different semantics, but is roughly the same. The one key difference is the former would have accepted `{ key: undefined }` whereas the latter doesn't (a breaking change, but one no one should be using) [second note, this behavior is possible to change with TS4.4's [Exaction Optional Property Types](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-beta/#exact-optional-property-types)]
- Simplified `Known` definition

It's not too complicated, so another set of eyes is always good. You may especially disagree with the null angle. Also a call on whether this is "breaking enough" for when to merge in.


epoberezkin:
I think it's ok to release without major version change. Thank you!

epoberezkin (approved):

Issue #1664, opened 2021-06-30
JSONSchemaType incorrectly requires optional properties to be nullable


swisener:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**

Ajv 8.6.0, Typescript 4.3.4

**Your typescript code**

```typescript
import { JSONSchemaType } from 'ajv/dist/2020';

interface Example {
  foo: string;
  bar?: string;
}

const schema: JSONSchemaType<Example> = {
  additionalProperties: false,
  type: 'object',
  properties: {
    foo: {
      type: 'string'
    },
    bar: {
      type: 'string'
    }
  },
  required: ['foo']
};
```

**Typescript compiler error messages**

```
TS2322: Type '{ additionalProperties: false; type: "object"; properties: { foo: { type: "string"; }; bar: { type: "string"; }; }; required: "foo"[]; }' is not assignable to type 'UncheckedJSONSchemaType<Example, false>'.
  Type '{ additionalProperties: false; type: "object"; properties: { foo: { type: "string"; }; bar: { type: "string"; }; }; required: "foo"[]; }' is not assignable to type '{ type: "object"; additionalProperties?: boolean | UncheckedJSONSchemaType<unknown, false> | undefined; unevaluatedProperties?: boolean | UncheckedJSONSchemaType<unknown, false> | undefined; ... 7 more ...; maxProperties?: number | undefined; } & { ...; } & { ...; } & { ...; }'.
    Type '{ additionalProperties: false; type: "object"; properties: { foo: { type: "string"; }; bar: { type: "string"; }; }; required: "foo"[]; }' is not assignable to type '{ type: "object"; additionalProperties?: boolean | UncheckedJSONSchemaType<unknown, false> | undefined; unevaluatedProperties?: boolean | UncheckedJSONSchemaType<unknown, false> | undefined; ... 7 more ...; maxProperties?: number | undefined; }'.
      The types of 'properties.bar' are incompatible between these types.
        Type '{ type: "string"; }' is not assignable to type '{ $ref: string; } | (UncheckedJSONSchemaType<string | undefined, false> & { nullable: true; const?: undefined; enum?: readonly (string | null | undefined)[] | undefined; default?: string | ... 1 more ... | undefined; })'.
[...]

PR #1701, opened 2021-07-18
JSONSchemaType change nullable to be for nulls not undefined

erikbrinkman:
#### What issue does this pull request resolve?

#1664 

#### What changes did you make?

This removes null from `Known`, changes `Nullable` to check for `null` instead of `undefined` in a type definition, and hacks `Nullable` to check if we're passing in `Known`, in which case it does a union instead of a conditional.

##### Detailed notes
They way that `Nullable<T>` handles types in `JSONSchemaType` makes it
very difficult to create `SomeJSONSchema` that any `JSONSchemaType` can
be assigned to. After many efforts to create this from scratch it
became clear that the only thing that would do this is a copy of
`JSONSchemaType` with `Nullable` fixed. Doing so revealed a lot of
surprising "bugs" in `SomeJSONSchema`, the most notable being some
places accept numbers where it should only accept string indices. This
seemed almost like a typescript bug, but I didn't delve in.

My conclusion from all of this is that it may be better to fully
redesign `JSONSchemaType` due to the number of inconsistencies. In light
of that, I opted for the least breaking change, which was actually to
hack `Nullable` to behave appropriately for `SomeJSONSchema`,
essentially, we check if that's what's being called, and if so, "do the
right thing". If someone were to try and type something identical to `Known` they
would similarly get a slightly more permissive type than they intended.
We could work around this by making the check explicit (akin to the
`IsPartial` keyword) but I think the times when this might happen aren't
bad enough to be worth it.

#### Is there anything that requires more attention while reviewing?

This is stacked on another PR, so only look at the second commit to see what this is changing. General thoughts on if you think another approach is more reasonable.


epoberezkin:
sorry, should I have merged this instead of #1699? Now there are conflicts...

erikbrinkman:
This is rebased so this one should be good. I'm currently trying to figure out how to appropriately stack PRs so that the second one can be reviewed independently.

epoberezkin:
this one probably should be a breaking change, am I right?

If so I should probably kick off v9 branch and merge it there...

erikbrinkman:
yeah, that makes sense, I can rebase / retarget onto that branch when it exists.

JamesJansson:
When is the release of v9 likely?

PR #1702, opened 2021-07-18
fix eslint warnings

erikbrinkman:

There were three eslint warnings, all about an unnecessary undefined check, this fixes them.

They were each slightly different:
1. One had a string index, which could be undefined, followed by a function call that would throw an exception if it was. Given that this wasn't causing crashes, I'm assuming that this branch was never hit, and can be safely remove for the time being.
2. Another has a valid check for an undefined string access, but typescript doesn't make accessing undefined by default. As of TS4.1 we could enable `"noUncheckedIndexedAccess": true` which will make index access also return undefined. That will fix the warning, but it introduces 29 typescript errors. The simplest mitigation would probably be to just ignore those with `!`, but I instead just ignored the warning so that when someone does add that check, they're more inclined to decide if we want to skip the check or enforce it at runtime.
3. Finally, one was a somewhat strange `any` bug. schemas have an index signature, potentially only containing other schemas? The access could be undefined, but schemas index value is any. The code is roughly
   ```
   schema: AnySchema
   schema = schema[key]
   ````
   if `schema[key]` returned undefined, typescript would error here, because you can't elide the undefined value when assigning to schema. typescript types hold for their whole scope, so the assignment doesn't widen the type. In this case, we need to first assign to a placeholder value, check that it's defined, and if so, then assign. This does do one extra variable assignment potentially slowing this down, although it doesn't look like a critical path, and this solution is future proofed for when the Schema index signature isn't any-typed.

This is pretty straightforward.

### DIFF
--- a/lib/compile/index.ts
+++ b/lib/compile/index.ts
@@ -300,11 +300,12 @@ function getJsonPointer(
 ): SchemaEnv | undefined {
   if (parsedRef.fragment?.[0] !== "/") return
   for (const part of parsedRef.fragment.slice(1).split("/")) {
-    if (typeof schema == "boolean") return
-    schema = schema[unescapeFragment(part)]
-    if (schema === undefined) return
+    if (typeof schema === "boolean") return
+    const partSchema = schema[unescapeFragment(part)]
+    if (partSchema === undefined) return
+    schema = partSchema
     // TODO PREVENT_SCOPE_CHANGE could be defined in keyword def?
-    const schId = typeof schema == "object" && schema[this.opts.schemaId]
+    const schId = typeof schema === "object" && schema[this.opts.schemaId]
     if (!PREVENT_SCOPE_CHANGE.has(part) && schId) {
       baseId = resolveUrl(baseId, schId)
     }
--- a/lib/runtime/parseJson.ts
+++ b/lib/runtime/parseJson.ts
@@ -128,14 +128,16 @@ export function parseJsonString(s: string, pos: number): string | undefined {
         let code = 0
         while (count--) {
           code <<= 4
-          c = s[pos].toLowerCase()
+          c = s[pos]
+          if (c === undefined) {
+            errorMessage("unexpected end")
+            return undefined
+          }
+          c = c.toLowerCase()
           if (c >= "a" && c <= "f") {
             code += c.charCodeAt(0) - CODE_A + 10
           } else if (c >= "0" && c <= "9") {
             code += c.charCodeAt(0) - CODE_0
-          } else if (c === undefined) {
-            errorMessage("unexpected end")
-            return undefined
           } else {
             errorMessage(`unexpected token ${c}`)
             return undefined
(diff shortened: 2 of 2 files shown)


lib/compile/index.ts
```diff
@@ -300,11 +300,12 @@ function getJsonPointer(
...
-    if (typeof schema == "boolean") return
-    schema = schema[unescapeFragment(part)]
-    if (schema === undefined) return
+    if (typeof schema === "boolean") return
+    const partSchema = schema[unescapeFragment(part)]
+    if (partSchema === undefined) return
+    schema = partSchema
     // TODO PREVENT_SCOPE_CHANGE could be defined in keyword def?
-    const schId = typeof schema == "object" && schema[this.opts.schemaId]
+    const schId = typeof schema === "object" && schema[this.opts.schemaId]
```
epoberezkin, line 308: eslint is fine with `==` in this particular context, it's like that everywhere.

lib/runtime/parseJson.ts
```diff
@@ -133,9 +133,6 @@ export function parseJsonString(s: string, pos: number): string | undefined {
             code += c.charCodeAt(0) - CODE_A + 10
           } else if (c >= "0" && c <= "9") {
             code += c.charCodeAt(0) - CODE_0
-          } else if (c === undefined) {
-            errorMessage("unexpected end")
-            return undefined
```
epoberezkin, line 138: I don't think it can be just removed, `c` can be undefined.

epoberezkin (changes requested):

lib/compile/index.ts
erikbrinkman: yeah this wasn't an eslint fix, just style fix. Is there a reason to use coercive equality here?

lib/runtime/parseJson.ts
erikbrinkman: yes and no. `s[pos]` can be undefined, but `s[pos]..toLowerCase()` can't because an undefined value will throw, and `toLowerCase` doesn't return undefined for any input.

epoberezkin: right, so this is a potential bug then - I don't remember if it is actually possible for pos to go out of range or was it simply defensive just in case - I think it can go out of range, in which case it needs to be fixed...

lib/compile/index.ts
epoberezkin: nope, just nitpick. It's like that everywhere (and eslint is happy with it because typeof always returns the string, so it's never coercive... nvm :)

erikbrinkman: I'm unsure how to interpret the nvm. Do you want me to change it back?

lib/runtime/parseJson.ts
erikbrinkman: Yeah, I was being lazy and not addressing the root cause, but figured that if it was possible people would be getting random TypeErrors and would have reported it? I can make the check accurate though.

lib/compile/index.ts
epoberezkin: nvm = "never mind" :)

epoberezkin: no need to revert

PR #1719, opened 2021-07-25
Ref fixes

### LINKED ISSUE #1718: Assigning defaults for nested schema doesn't work
I tried this but it didn't work:
```

const ajv = new Ajv({ useDefaults: true })
  const schema = {
    type: 'object',
    properties: {
      foo: { type: 'number' },
      bar: {
        type: 'object',
        properties: {
          zoo: { type: 'string', default: 'deffff' },
        },
      },
    },
    required: ['foo', 'bar'],
  }

  const data = {}
  const valid = validate(data)
  console.log(data)
```

erikbrinkman:

fixes #1718 1652

pulled `{ $ref: string }` from being defined on properties, to being a valid "top-level" type. Also pulled `Nullable` along with it. This is somewhat caused by the earlier diff that made nullable mean nullable not undefinable.

This is another commit in the stack, so only look at the last commit. The two tests that were added fail on the previous commit.

### DIFF
--- a/lib/types/json-schema.ts
+++ b/lib/types/json-schema.ts
@@ -34,12 +34,15 @@ interface StringKeywords {
 }
 
 type UncheckedJSONSchemaType<T, IsPartial extends boolean> = (
-  | // these two unions allow arbitrary unions of types
+  | // this allows empty $refs without needing to be in properties or items
   {
-      anyOf: readonly UncheckedJSONSchemaType<T, IsPartial>[]
+      readonly $ref: string
+    } // these two unions allow arbitrary unions of types
+  | {
+      readonly anyOf: readonly UncheckedJSONSchemaType<T, IsPartial>[]
     }
   | {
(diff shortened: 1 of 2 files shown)


epoberezkin:
also conflicts after merging #1699...

erikbrinkman:
I think this might be a problem if / when the other is merged. I think to do stacked PRs I need to have the branches be local for some reason, so I'll try that for the next commits.

epoberezkin:
this, unlike #1701, looks like non-breaking, correct?

erikbrinkman:
I think the ref change itself is non-breaking, but the nullable movement is conditioned on the previous commit, so breaking by dependence. Since those changes are somewhat linked due to where refs were positioned, it's easier just to keep this entirely marked as breaking, rather than trying to pull just the ref part out to commit on master.

epoberezkin:
@erikbrinkman I merged #1701 into v9 branch, and changed the base for this one to v9 - it now has the conflicts, sorry... Could you please have a look when you have a moment?

PR #1725, opened 2021-08-01
Pass `baseId` to `getSchemaRefs` to fix local `$ref` resolution in a schema without an id.

### LINKED ISSUE #1711: A `$ref` to a fragment defined with `$id` in the same document isn't found.
Version: 8.6.2

**Ajv options object**

```javascript
new Ajv({
  allErrors: true
  schemas: { "yaml.json": schema } // see below
});
```

**JSON Schema**

```json
{
  "definitions": {
    "SeeAlso": { "$id": "#SeeAlso", "type": "number" },
    "Engine": {
      "$id": "#Engine",
      "type": "object",
      "properties": {
        "see_also": { "$ref": "#SeeAlso" }
      }
    }
  }
}
```

**Sample data**

```json
{
  "see_also": 1
}
```

**Your code**

```javascript
var Ajv = require('ajv');
var schema = {
  "definitions": {
    "SeeAlso": { "$id": "#SeeAlso", "type": "number" },
    "Engine": {
      "$id": "#Engine",
      "type": "object",
      "properties": {
        "see_also": { "$ref": "#SeeAlso" }
      }
    }
  }
};

var data = {
  see_also: 1
};

ajv = new Ajv({
    allErrors: true,
    schemas: {
      "yaml.json": schema
    }
});

var result = ajv.validate("yaml.json#/definitions/Engine", data);
console.log(result);
```

[View on runkit](https://runkit.com/rbuckton/60f9d3906738f8001aab8ca5)

```
MissingRefError: can't resolve reference #SeeAlso from id yaml.json#Engine
```

As far as I can tell, it should be able to resolve fragment `#SeeAlso`, per https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01#section-8.2.4 and https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01#section-8.3.2.

If I have some time I can put up a PR. If anyone else wants to take this on, my findings are below based on debugging ajv:

When the `$ref` keyword is evaluated for `{ "$ref": "#SeeAlso" }`, I end up here in _ref.js_:

image

At this point, `baseId` is `yaml.json#Engine` and `$ref` is `#SeeAlso`.

When I step into `resolveRef`, the call to `resolveUrl` replaces `#SeeAlso` with `yaml.json#SeeAlso`:

image

Since this isn't a root ref, I step through to line 136, where you might expect to resolve `#SeeAlso` against `root.localRefs`, however that fails because the fragment is qualified:

image

As you can see in the watch window, `root.localRefs` contains a definition for `#SeeAlso`, however the local ref can't be resolved because `ref` was changed from `"#SeeAlso"` to `"yaml.json#SeeAlso"`.

If this is *supposed* to work, then I expect the solution would be to check just the fragment (`#SeeAlso`) against `root.localRefs` (after ensuring that `ref` and `baseId` differ only by URL fragment).

I've run into this issue on and off for several years and have always worked around it by using `#/definitions/Foo`, but from what I can tell from reading the spec, this kind of `$ref` *should* work:

image

rbuckton:

Fixes #1711

Pass `baseId` to `getSchemaRefs` so that `$ref` can resolve local references when the parent schema has no `$id`.

No, though running `npm run test-spec` results in 156 test failures *without* my changes, and doesn't seem to currently be wired up to `npm test` in any way.


lib/compile/resolve.ts
```diff
@@ -89,10 +89,10 @@ export function resolveUrl(baseId: string, id: string): string {
...
 const ANCHOR = /^[a-z_][-a-z0-9._]*$/i
 
-export function getSchemaRefs(this: Ajv, schema: AnySchema): LocalRefs {
+export function getSchemaRefs(this: Ajv, schema: AnySchema, baseId: string): LocalRefs {
   if (typeof schema == "boolean") return {}
   const {schemaId} = this.opts
-  const schId = normalizeId(schema[schemaId])
+  const schId = normalizeId(schema[schemaId] || baseId)
   const baseIds: {[JsonPtr in string]?: string} = {"": schId}
   const pathPrefix = getFullPath(schId, false)
```
rbuckton, line 97: Honestly, not even sure `schId` is necessary since `baseId` is already computed by `Ajv.prototype._addSchema`.
```suggestion
  const baseIds: {[JsonPtr in string]?: string} = {"": baseId}
  const pathPrefix = getFullPath(baseId, false)
```

epoberezkin:
thank you - in 8.6.3 - it was an annoying one, I'm surprised the fix was so simple ;)

PR #1733, opened 2021-08-14
v9

epoberezkin:


epoberezkin:
@erikbrinkman not sure what I've done here - will look later - this has some conflicts with master I didn't expect...

erikbrinkman:
yeah, I noticed. I think keeping flat histories (e.g. avoiding merge commits) helps reduce this. Either way, the proper resolution is just taking the version from this branch. Removing null from Know, and adding it to tuple respectively.

SrdjanCosicPrica:
Hi
Are there any plans for releasing V9?

Issue #1116, opened 2019-10-31
Should 'integer' type validate bigint ?


MaximeGagnebin:
This is more of a question than a proposal.

**What problem do you want to solve?**
I want to validate objects containing a bigint.

**What do you think is the correct solution to problem?**
Bigint instance could pass the validation against 'integer' type. This may sound weird as "regular" integer would also pass this validation, and in this way the "type" isn't really validated. But the JSON specifications do not impose a limit on the range of integers.

The other solution would be to extend the types to include BigInt.
Thanks for your answer.

epoberezkin:
it's ok with option "bigint"

MaximeGagnebin:
Could you clarify ? option is not a keyword in ajv ? Currently there is no "type" that will validate against a BigInt.

epoberezkin:
I mean that I am ok with the change to support it with a new option “bigint” that would treat them as “integer” from validation point of view. I recently did the similar change in json-source-maps to support bigint.

You can also use “typeof” custom keyword defined in ajv-keywords to specifically establish that the type (from JS point of view) is bigint rather than number.

Adding bigint to the list of types (from JSON Schema point of view) is definitely not ok - bigint type is purely JS thing and JSON Schema is supposed to be cross platform. I draw the line at allowing custom keywords - custom types are off limits :)

epoberezkin:
Treating bigint as integer seems reasonable to me, as “integer” is not a type in JS and indeed JSON has no limit on integer size. But as this is a non-backwards compatible change it requires an option, it cannot be the default behaviour (although this can become default behaviour in the next major version)

MaximeGagnebin:
Thanks for the detailled reply ! You make good points. Especially concerning custom types.

I think I will try to work with the "typeof" keyword for now. 
I will also look at how much work would be needed to accept bigint as "integer". This would be nice as it would allow to make the schema less JS specific. In that regard, it is helpful that expressions like "3 > 2n" returns true in JS. This make validation for "maximum" (and similar) a bit nicer.

Anyways, thanks again for the reply! :)

max-preuschen:
I'm fairly new to AJV, so I'm a bit stuck at the moment.
I have a given JSON-schema with 
```
"type": "integer",
"format": "int64"
```
Currently the schema doesn't pass validation, so I want to add `int64` as a custom format to the validator.

Unfortunately my tests are failing, since `type: integer` is converted into a JS-number instead of bigint.
[...]

PR #1739, opened 2021-08-23
feat: add bigint as primitive type

samuelscheit:
This PR adds ``bigint`` as a primitive type

- added ``bigint`` to ``JTDType``
- added ``bigint`` to the coercible types
- added ``bigint`` to the parser
- added ``bigint`` to the rules
- added ``bigint`` to the simpleTypes of ``jtd-schema.json``

I didn't implement tests for it, someone needs to help with that.
But I tested it manually with my schema.

### DIFF
--- a/lib/compile/jtd/parse.ts
+++ b/lib/compile/jtd/parse.ts
@@ -8,7 +8,7 @@ import N from "../names"
 import {hasPropFunc} from "../../vocabularies/code"
 import {hasRef} from "../../vocabularies/jtd/ref"
 import {intRange, IntType} from "../../vocabularies/jtd/type"
-import {parseJson, parseJsonNumber, parseJsonString} from "../../runtime/parseJson"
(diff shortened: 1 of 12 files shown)


epoberezkin:
Sorry, JTD is an RFC that we shouldn't really extend in an ad-hoc way.

There are also reason to NOT support integers over 2^32 in APIs and use strings instead (JTD spec covers it).

samuelscheit:
bigints are transcoded as strings and are coerced to bigint in the parser

samuelscheit:
what is the state of #1116?

epoberezkin:
#1116 can be implemented, but the interest seemed limited… It would be an option though, rather than a schema language change, and it would be JSON Schema only. In JTD schemas it could have been some custom keyword that would be used inside metadata member, that would effectively make strings of digits parse as bigint. Although just using strings seems cleaner to me…

PR #1747, opened 2021-08-31
Remove optional type from the validation context parameter

adrianolek:

8.6.2, happens in the latest version

**Your typescript code**

```typescript
import { DataValidateFunction } from "ajv/dist/types";

ajv.addKeyword({
  keyword: "test",
  schema: false,
  validate: function (_data: any, dataCxt) {
    dataCxt.rootData;
    return true;
  } as DataValidateFunction
});
```

**Typescript compiler error messages**

    (7,22): TS2532: Object is possibly `undefined`.

It means that the `dataCxt` object is possibly 'undefined'.

**Describe the change that should be made to address the issue?**

Remove the optional `?` operator from the `dataCxt` parameter in `ValidateFunction` and `SchemaValidateFunction` interface types.

I've checked the generated code, and the validation context object is always passed to the validation function. Also the codegen [validateFunction](https://github.com/ajv-validator/ajv/blob/8ffe5faca9c4b49fc538b35399a1b4febba0bc41/lib/compile/validate/index.ts#L49) seems to always include the validation context (although I'm not familiar with codegen itself, so I may be wrong here).

Yes. The PR removes the optional operator from `dataCxt` and renames `DataValidationCxt` to `ValidationCxt` & `dataCxt` to `valCxt` to match the codegen function.


epoberezkin:
I think it is most likely  a breaking change, and I do think it has to be optional... But in any case, please try to make the tests pass - I've fixed the typescript issue that was breaking the compilation.

PR #1756, opened 2021-09-09
Adds comments to properties on Options

chanced:
Improves user experience by inlining documentation of options in code editors like VSCode through intellisense.

Screen Shot 2021-09-09 at 1 17 50 PM

Adds comments to `Options` and `CodeGenOptions`.

Options missing comments:

-  `next`
-  `unevaluated`
- `dynamicRef`
- `schemaId`
- `defaultMeta`

### DIFF
--- a/lib/core.ts
+++ b/lib/core.ts
@@ -544,7 +544,7 @@ export interface CodeOptions {
    */
   formats?: Code
   /**
-   * add `source` property (see Source below) to validating function.
+   * add `source` property (seesSource) to validating function.
    */
   source?: boolean
   /**
--- a/lib/core.ts
+++ b/lib/core.ts
@@ -83,63 +83,473 @@ export type Options = CurrentOptions & DeprecatedOptions
 
 export interface CurrentOptions {
   // strict mode options (NEW)
+
+  /**
+   * ### strict
+   *
+   * By default Ajv executes in strict mode, that is designed to prevent any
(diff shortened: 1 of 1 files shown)


epoberezkin:
Sorry, I am not supportive of the idea to optimise comments for a particular editor, even though I am using VS code myself.

It has marginal utility to the users - all the docs are available on the website - but it makes it more difficult to maintain and to contribute, as now the docs have to be maintained in two places and the code is much more difficult to read.

The compromise could be to include the links to the website sections, so each option still takes a single line, but I am not sure if VSCode would understand inline comments and whether these links would be clickable.

chanced:
I doubt VS Code is the only editor that does this. In Go, comments of this nature are considered very important so I suspect that WebStorm has it built in, given JetBrains' success with GoLand. I'd be incredibly surprised if Vi / emacs didn't have plugins to support for it either.

>  but it makes it more difficult to maintain and to contribute, as now the docs have to be maintained in two places and the code is much more difficult to read.

Unless the docs are generated from the comments. I understand that's probably not desirable of a refactor, especially if you think it makes the code less readable. It does, however, make it far more apparent when there are options lacking documentation. 

I'm in no way, shape or form, a fan of JSDoc so I totally get why folks in the js ecosystem shy away from the idea of documenting in this fashion. I think it went way overboard.  Go got this right though aided by the fact that it's a statically typed language.

chanced:
Also, sorry I missed your previous comment until now. Take care.

PR #1815, opened 2021-11-12
resolve references before validating the discriminator

### LINKED ISSUE #1554: Discriminator logic will not dereference schemas before validating preconditions

`8.0.5`

Defer the evaluation of precondition checks on `discriminator`s after dereferencing `$ref`s. Currently, a schema will fail validation if the list of `oneOf` children contains a referenced schema instead of an embedded schema.

See: https://github.com/ajv-validator/ajv/blob/baf1475cede6d7606bb1009bbdd1a942f76cec6c/lib/vocabularies/discriminator/index.ts#L64-L72

The preconditions for `discriminator` support should be deferred.

I don't think so. 😬

dfeufel:
fixes #1554


lib/vocabularies/discriminator/index.ts
```diff
@@ -62,8 +63,12 @@ const def: CodeKeywordDefinition = {
       const topRequired = hasRequired(parentSchema)
       let tagRequired = true
       for (let i = 0; i < oneOf.length; i++) {
-        const sch = oneOf[i]
-        const propSch = sch.properties?.[tagName]
+        let sch = oneOf[i]
+        let propSch = sch.properties?.[tagName]
+        if (typeof propSch == "undefined" && sch?.["$ref"]) {
```
epoberezkin: This would only be a valid approach if $ref is the only keyword in the schema and there are no other keywords

epoberezkin: And the error message below needs to be changed

epoberezkin:
Great - added some comments, it’s quite a common request.

please also update the docs to mention that single $refs (that is, without siblings) are now supported as well.

lib/vocabularies/discriminator/index.ts
dfeufel: I tried to update the PR according to your comments. But I still cannot figure out why the CI builds are failing. Locally "npm run test" runs fine. Am I missing something?

epoberezkin:
CI runs a bit more combinations of options than run locally by default, it uses:

AJV_FULL_TEST=true npm test

Maybe that’s the difference?

See the test output in CI.

epoberezkin:
So you can try locally with the same env variable

lib/vocabularies/discriminator/index.ts
```diff
@@ -62,10 +63,24 @@ const def: CodeKeywordDefinition = {
       const topRequired = hasRequired(parentSchema)
       let tagRequired = true
       for (let i = 0; i < oneOf.length; i++) {
-        const sch = oneOf[i]
-        const propSch = sch.properties?.[tagName]
-        if (typeof propSch != "object") {
-          throw new Error(`discriminator: oneOf schemas must have "properties/${tagName}"`)
+        let sch = oneOf[i]
+        let propSch
+        if (sch?.["$ref"] && Object.keys(sch).length === 1) {
```
epoberezkin: There is some function that tests whether schema has validation keywords other than ref (ignoring e.g. title) - I will update it later.

lib/vocabularies/discriminator/index.ts
```diff
@@ -62,10 +63,24 @@ const def: CodeKeywordDefinition = {
...
+          if (typeof propSch != "object") {
+            throw new Error(
+              `discriminator: referenced oneOf schemas must have "properties/${tagName}"`
+            )
+          }
+        } else {
+          propSch = sch.properties?.[tagName]
+          if (typeof propSch != "object") {
+            throw new Error(`discriminator: oneOf schemas must have "properties/${tagName}"`)
+          }
```
epoberezkin: I would take them out of if/else to simplify - the only way they differ is by the error message, which can be unified to be ``` `discriminator: oneOf subschema (or referenced schema) must have "properties/${tagName}"` ```

epoberezkin:
Thanks for the change, it definitely improves the usability of discriminator - only supporting inlined schema was an annoying restriction...

There couple of comments I wanted to amend - let me know if it's ok I do it?

lib/vocabularies/discriminator/index.ts
epoberezkin: ah - this is `schemaHasRulesButRef(sch, self.RULES)` see in validate/index

dfeufel:
Sure, feel free.

docs/json-schema.md
```diff
@@ -992,7 +992,7 @@ There are following requirements and limitations of using `discriminator` keywor
 - `mapping` in discriminator object is not supported.
 - [oneOf](#oneof) keyword must be present in the same schema.
 - discriminator property should be [requried](#required) either on the top level, as in the example, or in all `oneOf` subschemas.
-- each `oneOf` subschema must have [properties](#properties) keyword with discriminator property.
+- each `oneOf` subschema must have [properties](#properties) keyword with discriminator property. Here, the subschemas can also be included by reference (`$ref` keyword).
```
epoberezkin:
```suggestion
- each `oneOf` subschema must have [properties](#properties) keyword with discriminator property. The subschemas should be either inlined or included as direct references (only `$ref` keyword without any extra keywords is allowed). 
```

lib/vocabularies/discriminator/index.ts
epoberezkin:
```suggestion
        if (sch?.$ref && !schemaHasRulesButRef(sch, it.self.RULES)) {
```

lib/vocabularies/discriminator/index.ts
epoberezkin:
```suggestion
          sch = resolveRef.call(it.self, it.schemaEnv, it.baseId, sch?.$ref)
          if (sch instanceof SchemaEnv) sch = sch.schema
        }
        propSch = sch?.properties?.[tagName]
        if (typeof propSch != "object") {
          throw new Error(
            `discriminator: oneOf subschema (or referenced schema) must have "properties/${tagName}"`
          )
```

lib/vocabularies/discriminator/index.ts
```diff
@@ -2,6 +2,7 @@ import type {CodeKeywordDefinition, AnySchemaObject, KeywordErrorDefinition} fro
 import type {KeywordCxt} from "../../compile/validate"
 import {_, getProperty, Name} from "../../compile/codegen"
 import {DiscrError, DiscrErrorObj} from "../discriminator/types"
+import {resolveRef, SchemaEnv} from "../../compile"
```
epoberezkin, line 5:
```suggestion
import {resolveRef, SchemaEnv} from "../../compile"
import {schemaHasRulesButRef} from "./util"
```

lib/vocabularies/discriminator/index.ts
```diff
@@ -2,6 +2,8 @@ import type {CodeKeywordDefinition, AnySchemaObject, KeywordErrorDefinition} fro
...
+import {schemaHasRulesButRef} from "./util"
```
epoberezkin:
```suggestion
import {schemaHasRulesButRef} from "../../compile/util"
```

lib/vocabularies/discriminator/index.ts
```diff
@@ -62,10 +64,17 @@ const def: CodeKeywordDefinition = {
       const topRequired = hasRequired(parentSchema)
       let tagRequired = true
       for (let i = 0; i < oneOf.length; i++) {
-        const sch = oneOf[i]
-        const propSch = sch.properties?.[tagName]
+        let sch = oneOf[i]
+        let propSch
```
epoberezkin:
```suggestion
```

lib/vocabularies/discriminator/index.ts
```diff
@@ -62,10 +64,17 @@ const def: CodeKeywordDefinition = {
...
+        if (sch?.$ref && !schemaHasRulesButRef(sch, it.self.RULES)) {
+          sch = resolveRef.call(it.self, it.schemaEnv, it.baseId, sch?.$ref)
+          if (sch instanceof SchemaEnv) sch = sch.schema
+        }
+        propSch = sch?.properties?.[tagName]
```
epoberezkin:
```suggestion
        const propSch = sch?.properties?.[tagName]
```

lib/vocabularies/discriminator/index.ts
```diff
@@ -62,10 +64,16 @@ const def: CodeKeywordDefinition = {
...
+        let sch = oneOf[i]
+        if (sch?.$ref && !schemaHasRulesButRef(sch, it.self.RULES)) {
+          sch = resolveRef.call(it.self, it.schemaEnv, it.baseId, sch?.$ref)
+          if (sch instanceof SchemaEnv) sch = sch.schema
+        }
+        const propSch = sch?.properties?.[tagName]
         if (typeof propSch != "object") {
-          throw new Error(`discriminator: oneOf schemas must have "properties/${tagName}"`)
+          throw new Error(
+            `discriminator: oneOf subschema (or referenced schema) must have "properties/${tagName}"`
```
epoberezkin:
```suggestion
            `discriminator: oneOf subschemas (or referenced schemas) must have "properties/${tagName}"`
```

Issue #1667, opened 2021-07-01
Ajv 8.6.0 makes VSCode IntelliSense extremely slow


aarbmx6s:
After adding to a project  **Ajv 8.6.0** and importing it in the project makes [InteliSense](https://code.visualstudio.com/docs/editor/intellisense) very slow.

Typing some text and getting suggestions for code completions takes ~5-15 sec. 

[Screen recording of the issue](https://user-images.githubusercontent.com/22832351/124131680-b448a800-da88-11eb-8652-3c3f1ab861b4.mp4)

[Example of project ](https://github.com/aarbmx6s/vscode-ajv-issue)

With **Ajv 7.2.4** no problem

epoberezkin:
Ajv v8 uses somewhat more complex JSONSchemaType type definition than v7, it enables union support and better error reporting.

The compilation time in 5-15 seconds for such a simple example is still difficult to explain - it might be related to how much memory, what CPU you have, so it's ultimately the problem of either TS or VSCode performance. The file with the type definition is only 180 lines of code - it is a complex type, but it should still be compiled in under 1 sec time.

Ultimately, the resource requirements for "compiling as you type" would be higher than just for editing and you need to either free-up memory and/or CPU, re-configure (or change) the editor so it is impacting the UI less.

I do not have such a noticeable performance degradation by the way, with substantially larger schemas than in your video.

Out of curiosity, what is the compilation time when you just compile with tsc? What's the system configuration (memory/CPU/OS)?

In any case, this is not a problem that can or should be addressed in Ajv - the compilation performance is something to be addressed in TS and or VSCode.

epoberezkin:
Also, why do you require ajv TS code? You absolutely should require compiled ajv code with `import Ajv from "ajv"`, not the TS source - that would absolutely make performance worse.

aarbmx6s:
> Out of curiosity, what is the compilation time when you just compile with tsc? What's the system configuration (memory/CPU/OS)?

I have a large project and compilation time takes ~10 sec
My system: MacBook Pro 13, macOS Big Sur, i5, 16GB RAM.

> In any case, this is not a problem that can or should be addressed in Ajv - the compilation performance is something to be addressed in TS and or VSCode.

I tried PhpStorm IDE and the same behaviour. Seems like some problem with TS.

> Also, why do you require ajv TS code? You absolutely should require compiled ajv code with import Ajv from "ajv", not the TS source - that would absolutely make performance worse.

[...]

PR #1833, opened 2021-12-05
split union types

erikbrinkman:

https://github.com/microsoft/TypeScript/pull/30639 introduced regressions for typescript checking of JSONSchemaType. It was suggested that https://gist.github.com/armanio123/95a33f9ad3286a7d78590eb11af293c4 might improve things. However, that diff removed the usage of `UnionToIntersection` which may be causing most of the regression.

After implementing this and cloning the [example](https://github.com/aarbmx6s/vscode-ajv-issue) I saw identical performance for running `tsc`, a little less than 2s on my machine. I'm posting so that the original so that the OP can try this branch out to see if it fixes anything for them.

More discussion [here](https://github.com/microsoft/TypeScript/issues/44851)

Meant to address #1667

Simply split out the types as in the gist.

You could double check I didn't make a mistake, but the tests pass, so this should be fine.

### DIFF
--- a/lib/types/json-schema.ts
+++ b/lib/types/json-schema.ts
@@ -33,115 +33,138 @@ interface StringKeywords {
   format?: string
 }
 
-type UncheckedJSONSchemaType<T, IsPartial extends boolean> = (
-  | // these two unions allow arbitrary unions of types
-  {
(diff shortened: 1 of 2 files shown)


erikbrinkman:
This didn't improve performance, and the problem was fixed in a recent release of typescript

Issue #1523, opened 2021-03-30
Option to generate ESM for standalone validators


realityking:
**What version of Ajv you are you using?**

8.0.1

**What problem do you want to solve?**

Importing CommonJS modules creates non-ideal code when bundling with Rollup or Webpack and let them not participate in tree shaking. Code will be smaller and more efficient when ESM is used for bundling.
This will also gain in importance as npm packages will soon ship more and more ESM code.

**What do you think is the correct solution to problem?**

There are two parts to be done here:

1. Add an `esm` option to `ajv compile`/`standaloneCode`
2. Ship ESM versions of the runtime files and have an `exports` map for Bundles and Node.js to pick them up correctly.

The first change is the more important one but both are necessary for the full effect.

**Will you be able to implement it?**

I'm not sure. I started looking through the code a bit and I see how to get started but I expect to get stuck at some point. I'd definitely need help down the line.

epoberezkin:
This does not seem in scope, there are other tools that do it. This is not the same argument as exists for supporting ES5 code - it is needed for run-time schema compilation in ES5 environments.

For build time compilation you can compile to CommonJS and then use some other tool to transpile to ESM that would then be passed to any tool that works better with ESM than with CommonJS. If anything, arguably, it is definitely more in scope of rollup or webpack to have better support for CommonJS that it is in scope of Ajv to generate ESM, but in any case another tool can be used.

realityking:
> If anything, arguably, it is definitely more in scope of rollup or webpack to have better support for CommonJS that it is in scope of Ajv to generate ESM, but in any case another tool can be used.

Both Webpack and Rollup have invested a lot of time into this and it works from interoperability standpoint. It does however generate less optimal code (everything will be wrapped in a closure) and in the case of Webpack the module won't participate in tree shaking.

ESM is the strictest of all the module formats as every export and import has to be static. This makes it easiest to transpile into any other module format. With both browsers and Node.js converging on it as the shared standard, I believe it's where the future is heading. It'd be great to gain of some the benefits with standalone validators generated by AJV.

epoberezkin:
I understand - ESM is less flexible than CommonJS, so there is no general way to effectively transpile from all CommonJS to ESM. On another hand, majority of CommonJS modules have effectively static exports, so it could be possible to detect it during code analysis and when it is the case to avoid closures... Not 100% sure about it.

[...]

PR #1861, opened 2021-12-19
Add option to generate ESM exports instead of CJS (#1523)

rehanvdm:
Adds an option so that the generated code can do both CJS and ESM exports. https://github.com/ajv-validator/ajv/issues/1523

Adde an extra flag  `code.exportEsm` to generate code with ESM exports instead of CJS.

I wanted to use the `code.es5` flag and then make the assumption that ES5 code uses CJS and ES6 uses ESM. But this would have been a breaking change because by default the code is exporting as ES6. So I thought it best to add an additional flag called `code.exportEsm`. It defaults to false so it introduces no breaking changes.

CJS (default) creates exports as `module.exports = ` for single exports and `exports.` or `exports[` for multiple exports. ESM creates exports as `export const` and  `export default` for single exports, where the `export const` has been given a fixed name of `validate`. This is just to cater for both import methods. Then for the multiple exports, the `export default` option is not available(or has not been implemented) and must be done with named exports as in `export const`.

I tested both scenarios for when the code has a single and multiple exports. Note that ESM can not export with special characters in the name, CJS can because it then uses the array syntax like `exports[https://example.com/string.json]`. I didn't just want it to fail the ESM version when the CJS passed so instead I used some regex to replace special characters and then the ESM export would look like `export const https___example_com_string_json`

I also needed to add a new devDependencies called `module-from-string` to be able to test the generated code. It is similar to `require-from-string`. The [package](https://www.npmjs.com/package/module-from-string) is almost a year old with not a lot of stars or downloads but looks stable and was good enough for the tests. I am not sure how else to test ESM modules, open to suggestions if you do not like this method.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -65,7 +65,7 @@ const defaultOptions = {
   code: {
     // NEW
     es5: false,
-    exportEsm: false,
+    esm: false,
     lines: false,
     source: false,
     process: undefined, // (code: string) => string
@@ -348,7 +348,7 @@ Code generation options:
 ```typescript
 type CodeOptions = {
   es5?: boolean // to generate es5 code - by default code is es6, with "for-of" loops, "let" and "const"
-  exportEsm?: boolean // how functions should be exported - by default CJS is used, so the validate function(s) 
+  esm?: boolean // how functions should be exported - by default CJS is used, so the validate function(s) 
   // file can be `required`. Set this value to true to export the validate function(s) as ES Modules, enabling 
   // bunlers to do their job.
   lines?: boolean // add line-breaks to code - to simplify debugging of generated functions
--- a/docs/options.md
+++ b/docs/options.md
@@ -65,6 +65,7 @@ const defaultOptions = {
   code: {
     // NEW
     es5: false,
+    exportEsm: false,
     lines: false,
     source: false,
     process: undefined, // (code: string) => string
@@ -347,6 +348,9 @@ Code generation options:
 ```typescript
 type CodeOptions = {
   es5?: boolean // to generate es5 code - by default code is es6, with "for-of" loops, "let" and "const"
+  exportEsm?: boolean // how functions should be exported - by default CJS is used, so the validate function(s) 
+  // file can be `required`. Set this value to true to export the validate function(s) as ES Modules, enabling 
+  // bunlers to do their job.
   lines?: boolean // add line-breaks to code - to simplify debugging of generated functions
   source?: boolean // add `source` property (see Source below) to validating function.
   process?: (code: string, schema?: SchemaEnv) => string // an optional function to process generated code
--- a/lib/compile/codegen/code.ts
+++ b/lib/compile/codegen/code.ts
@@ -155,6 +155,19 @@ export function getProperty(key: Code | string | number): Code {
   return typeof key == "string" && IDENTIFIER.test(key) ? new _Code(`.${key}`) : _`[${key}]`
 }
 
+//Does best effort to format the name properly
+export function getEsmExportName(key: Code | string | number): Code {
+  if (typeof key == "string" && IDENTIFIER.test(key)) {
+    return new _Code(`${key}`)
+  }
+  const name = key.toString()
+  let newName = name.replace(/[^\w$_]|[\s]/gi, "_")
+  if (/^[0-9]/.test(name)) {
+    newName = "_" + newName
+  }
+  return new _Code(`${newName}`)
+}
+
 export function regexpCode(rx: RegExp): Code {
   return new _Code(rx.toString())
 }
--- a/lib/compile/codegen/code.ts
+++ b/lib/compile/codegen/code.ts
@@ -160,12 +160,7 @@ export function getEsmExportName(key: Code | string | number): Code {
   if (typeof key == "string" && IDENTIFIER.test(key)) {
     return new _Code(`${key}`)
   }
-  const name = key.toString()
-  let newName = name.replace(/[^\w$_]|[\s]/gi, "_")
-  if (/^[0-9]/.test(name)) {
-    newName = "_" + newName
-  }
-  return new _Code(`${newName}`)
+  throw new Error(`CodeGen: invalid export name: ${key}, use explicit $id name mapping`)
 }
 
 export function regexpCode(rx: RegExp): Code {
--- a/lib/core.ts
+++ b/lib/core.ts
@@ -140,7 +140,7 @@ export interface CurrentOptions {
 
 export interface CodeOptions {
   es5?: boolean
-  exportEsm?: boolean
+  esm?: boolean
   lines?: boolean
   optimize?: boolean | number
   formats?: Code // code to require (or construct) map of available formats - for standalone code
--- a/lib/core.ts
+++ b/lib/core.ts
@@ -140,6 +140,7 @@ export interface CurrentOptions {
 
 export interface CodeOptions {
   es5?: boolean
+  exportEsm?: boolean
   lines?: boolean
   optimize?: boolean | number
   formats?: Code // code to require (or construct) map of available formats - for standalone code
--- a/lib/standalone/index.ts
+++ b/lib/standalone/index.ts
@@ -30,7 +30,7 @@ function standaloneCode(
     const usedValues: UsedScopeValues = {}
     const n = source?.validateName
     const vCode = validateCode(usedValues, source)
-    if (ajv.opts.code.exportEsm) {
+    if (ajv.opts.code.esm) {
       // Always do named export as `validate` rather than the variable `n` which is `validateXX` for known export value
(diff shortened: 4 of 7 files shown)


docs/options.md
```diff
@@ -65,6 +65,7 @@ const defaultOptions = {
   code: {
     // NEW
     es5: false,
+    exportEsm: false,
```
epoberezkin: maybe just esm?

also I was wondering, it’s not a problem if both es5 and esm are combined, esm export can be limited to es5 syntax too, so they are indeed orthogonal…

lib/compile/codegen/code.ts
```diff
@@ -155,6 +155,19 @@ export function getProperty(key: Code | string | number): Code {
   return typeof key == "string" && IDENTIFIER.test(key) ? new _Code(`.${key}`) : _`[${key}]`
 }
 
+//Does best effort to format the name properly
+export function getEsmExportName(key: Code | string | number): Code {
```
epoberezkin, line 159: To be honest, I think it’s better to simply fail if user attempts to use a non-identifier as schema key, then try to adjust based on some rules that a user does not know producing unexpected names.

lib/standalone/index.ts
```diff
@@ -43,7 +47,15 @@ function standaloneCode(
       const v = getValidateFunc(schemas[name] as T)
       if (v) {
         const vCode = validateCode(usedValues, v.source)
-        code = _`${code}${_n}exports${getProperty(name)} = ${v.source?.validateName};${_n}${vCode}`
+        if (ajv.opts.code.exportEsm) {
```
epoberezkin: I am tempted to refactor common parts out

epoberezkin:
This is great, thank you! Please see comments - let me know what you think

docs/options.md
rehanvdm: >maybe just esm? 

That makes sense, follows the naming convention of the es5 option

>also I was wondering, it’s not a problem if both es5 and esm are combined, esm export can be limited to es5 syntax too, so they are indeed orthogonal…

Correct me if I am wrong, but at default es5 is false, so it uses es6. If we assume that many are using the default, it would be a breaking change if we make the rule that es6 exports in ESM. That is what happened while I was developing, many of the tests failed because they didn't specify es5 true, meaning es6 + ESM and then the tests failed because the moduleCode was still using `requireFromString` which did not work because the code is now a module so I had to use `importFromStringSync`. That is why I don't think they can be combined?

lib/compile/codegen/code.ts
rehanvdm: Makes sense, just though I do it this way as I saw ES Build do it like this. Will change it to throw an error. Any idea for a good error message? I saw in the tests that it is possible to alias $ids 

```js
     const moduleCode = standaloneCode(ajv, {
          validateNumber: "https://example.com/number.json",
          validateString: "https://example.com/string.json",
        })
```

Maybe mention the solution like that in the error?

lib/standalone/index.ts
rehanvdm: Okay will do for this IF block or did you have something bigger in mind?

rehanvdm:
> This is great, thank you! Please see comments - let me know what you think

Okay cool, this will actually be my first Open Source PR. I have done a few things here or there, but this one feels much bigger and certainly the one that took the most time 😅. Great repo by the way, good tests, linting and safeguards. I did struggle a little bit to get it going Windows, but managed with WSL (used to many projects only supporting Linux development)

docs/options.md
epoberezkin: Sorry, by “combined” I meant that it will be ok if a user uses both options. These should be separate options in ajv, you are right.

lib/compile/codegen/code.ts
epoberezkin: Something like: invalid export name ${name}. You can use an explicit name mapping.

Do the docs not have this name mapping, is it only in tests?

lib/standalone/index.ts
epoberezkin: No, only this if

docs/options.md
epoberezkin:
```suggestion
    esm: false,
```

docs/options.md
```diff
@@ -347,6 +348,9 @@ Code generation options:
 ```typescript
 type CodeOptions = {
   es5?: boolean // to generate es5 code - by default code is es6, with "for-of" loops, "let" and "const"
+  exportEsm?: boolean // how functions should be exported - by default CJS is used, so the validate function(s)
```
epoberezkin:
```suggestion
  esm?: boolean // how functions should be exported - by default CJS is used, so the validate function(s) 
```

lib/core.ts
```diff
@@ -140,6 +140,7 @@ export interface CurrentOptions {
 
 export interface CodeOptions {
   es5?: boolean
+  exportEsm?: boolean
```
epoberezkin:
```suggestion
  esm?: boolean
```

lib/standalone/index.ts
epoberezkin:
```suggestion
        if (ajv.opts.code.esm) {
```

spec/standalone.spec.ts
```diff
@@ -21,31 +42,61 @@ describe("standalone code generation", () => {
...
+        })
+        testExportTypeCjs(moduleCode, false)
+        const m = requireFromString(moduleCode)
+        assert.strictEqual(Object.keys(m).length, 2)
+        testExports(m)
       })
 
-      it("should generate module code with named exports", () => {
+      it("should generate module code with named export - ESM", () => {
+        ajv = new _Ajv({code: {source: true, exportEsm: true}})
```
epoberezkin:
```suggestion
        ajv = new _Ajv({code: {source: true, esm: true}})
```

spec/standalone.spec.ts
```diff
@@ -21,31 +42,61 @@ describe("standalone code generation", () => {
...
         const m = requireFromString(moduleCode)
         assert.strictEqual(Object.keys(m).length, 2)
         testExports({
           validateNumber: m["https://example.com/number.json"],
           validateString: m["https://example.com/string.json"],
         })
       })
+
+      it("should generate module code with all exports - ESM", () => {
+        ajv = new _Ajv({code: {source: true, exportEsm: true}})
```
epoberezkin:
```suggestion
        ajv = new _Ajv({code: {source: true, esm: true}})
```

spec/standalone.spec.ts
```diff
@@ -242,6 +294,26 @@ describe("standalone code generation", () => {
     }
   })
 
+  it("should generate module code with a single export - ESM", () => {
+    const ajv = new _Ajv({code: {source: true, exportEsm: true}})
```
epoberezkin:
```suggestion
    const ajv = new _Ajv({code: {source: true, esm: true}})
```

docs/options.md
rehanvdm: Okay cool

lib/compile/codegen/code.ts
rehanvdm: I was about to say no it doesn't, but after about 15minutes of reading the docs I see that it is mentioned only in code comments here: https://ajv.js.org/standalone.html#usage-from-code. When I initially read it last week when I started to really dig into AJV, I just didn't notice it and did not understand why it is needed so thought it isn't important. 

It might be good to somehow highlight that if `code.esm` is true then your ids must be valid/exportable names and that you can remap them like that? Otherwise you will get that error.

rehanvdm:
Changes applied, let me know what you think

lib/compile/codegen/code.ts
epoberezkin: yes, worth adding to the docs, but can be a separate PR

lib/standalone/index.ts
```diff
@@ -43,7 +47,10 @@ function standaloneCode(
       const v = getValidateFunc(schemas[name] as T)
       if (v) {
         const vCode = validateCode(usedValues, v.source)
-        code = _`${code}${_n}exports${getProperty(name)} = ${v.source?.validateName};${_n}${vCode}`
+        const exportSyntax = ajv.opts.code.esm
```
epoberezkin, line 50: cool

epoberezkin:
Thank you, this is a very solid piece of work - merged, release is coming.
I would only add named exports to the docs, as you said, and the test when "all exports" actually work with ESM - e.g. when schema IDs are just identifiers, as the spec allows - does it make sense?

rehanvdm:
Pleasure. On the docs and additional PR, I am busy writing/rewriting this https://ajv.js.org/standalone.html doc to have that concrete example I mentioned. I am also including knowledge/the options about ES5/ES6 and CJS/ESM exports + examples. 

Might be able to get it done tomorrow night, otherwise only sometime next week (mid-week) - Just for if you want to hold off on the release for this doc change.

epoberezkin:
Releases just happen, no problem if it’s after that.

PR #1862, opened 2021-12-20
feat: add uriresolver option

### LINKED ISSUE #1844: Replace uri-js by fast-uri
fast-uri is a fork of uri-js which is rewritten in javascript to focus on performances issues.

Ajv 8.8.2

Speed improvement

As discussed with @epoberezkin , an option can be added to inject `fast-uri` to replace `uri-js` to not have a breaking change in the current version

Yes

Ref: https://github.com/fastify/fast-uri

zekth:

fix: https://github.com/ajv-validator/ajv/issues/1844

Integrate `fast-uri` as an optional uri resolver.

Not sure about where to add the tests. Right now i only used the `resolve` test suite. Is there any other spot where we should consider adding `fast-uri` in the test suite?

### DIFF
--- a/lib/compile/index.ts
+++ b/lib/compile/index.ts
@@ -257,9 +257,14 @@ export function resolveSchema(
   root: SchemaEnv, // root object with properties schema, refs TODO below SchemaEnv is assigned to it
   ref: string // reference to resolve
 ): SchemaEnv | undefined {
-  const p = URI.parse(ref)
-  const refPath = _getFullPath(p)
-  let baseId = getFullPath(root.baseId)
+  let p: URI.URIComponents
+  if (this.opts.uriResolver) {
+    p = this.opts.uriResolver.parse(ref)
+  } else {
+    p = URI.parse(ref)
+  }
+  const refPath = _getFullPath(p, this.opts.uriResolver)
+  let baseId = getFullPath(root.baseId, undefined, this.opts.uriResolver)
   // TODO `Object.keys(root.schema).length > 0` should not be needed - but removing breaks 2 tests
   if (Object.keys(root.schema).length > 0 && refPath === baseId) {
     return getJsonPointer.call(this, p, root)
--- a/lib/compile/index.ts
+++ b/lib/compile/index.ts
@@ -112,7 +112,7 @@ export function compileSchema(this: Ajv, sch: SchemaEnv): SchemaEnv {
   // TODO refactor - remove compilations
   const _sch = getCompilingSchema.call(this, sch)
   if (_sch) return _sch
-  const rootId = getFullPath(sch.root.baseId, this.opts.uriResolver) // TODO if getFullPath removed 1 tests fails
+  const rootId = getFullPath(this.opts.uriResolver, sch.root.baseId) // TODO if getFullPath removed 1 tests fails
   const {es5, lines} = this.opts.code
   const {ownProperties} = this.opts
   const gen = new CodeGen(this.scope, {es5, lines, ownProperties})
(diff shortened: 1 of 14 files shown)


mcollina (commented):
Good work!

zekth:
Any news on this?

epoberezkin:
@zekth - thanks for doing it.

I think using a condition at all call sites is a brittle approach that may lead to missing some call site, particularly in the future changes.

What could be done instead is removing URI import from resolve.ts entirely (and whatever other file that uses it now, if any) and instead putting it into options, similarly to how it was done for RegExp. Then all call sites would be updated to unconditionally use the resolver from the options, whether it's the default one or the one passed by the user.

This would be set up during instance options initialisation, there are two different types for user supplied options (Options) and runtime options (InstanceOptions), in the latter the uriResolver should be a required member, so you won't have to check in each call site if it is present, and just use it directly.

Also it has to be supported in standalone code, same as it was done for RegExp (I think).

I am referring to this PR that added RE2 support - https://github.com/ajv-validator/ajv/pull/1684 (it's marked as closed but it was just accidentally merged via another branch, you can see in the master branch how it is done)...

zekth:
@epoberezkin i addressed the changes needed. Also discovered that i forgot some other changes. If the shape of this PR is ok for you let me know and i'll rebase it properly.

epoberezkin:
looks good - thank you!

Couple questions/comments:

1. Why does MissingRefError need a reference to resolver? I don't see it used anywhere...
2. getFullPath accepts resolver as the first parameter and resolveUrl as the last - could you please make it the first in both cases? Another option would be just to call them with Ajv context so resolver is there... (e.g. `getFullPath.call(this, ...)`)

zekth:
1- It's needed because of this: https://github.com/ajv-validator/ajv/pull/1862/files#diff-14042115de8ebf12385d7afecb8ceaafd6cd54b8ae726e5ef5a68509c965755dR10-R11
2 - Sure i'll rework the shape.

epoberezkin:
I see - thank you!

zekth:
@epoberezkin i reordered the parameters. Is it ok this way or you'd prefer to pass the complete `Ajv` context within the functions?

mcollina:
I think it's better to bump fast-uri to 0.1.0 or even 1.0.0 to simplify maintenance.

epoberezkin:
@zekth - this is great, sorry for being so slow - going to release it now!

epoberezkin:
alright, I already pushed the release out, and then it occurred to me that we didn't add any docs for the feature.

@zekth - please add to the options page if you can :)

Thank you!

zekth:
Will do ASAP!

Maribel400 (commented):
😏

Tuneabq (commented):
You like

Tuneabq (commented):
Bump it

epoberezkin:
Bump what? :)

Maribel400:
Vaya usted a chingar a su madre

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

Maribel400:
Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows
sakese

chevett:
It appears that this is a breaking change?  i'm seeing https://github.com/ajv-validator/ajv-merge-patch/blob/master/keywords/add_keyword.js#L22 fail with ajv@8.10+

fetobasic:
> **What issue does this pull request resolve?**
> 
> fix: #1844
> 
> **What changes did you make?**
> 
> Integrate `fast-uri` as an optional uri resolver.
> 
> **Is there anything that requires more attention while reviewing?**
> 
> Not sure about where to add the tests. Right now i only used the `resolve` test suite. Is there any other spot where we should consider adding `fast-uri` in the test suite?

@zekth @mcollina @epoberezkin Correct me if I'm wrong, but I assume even if you go with the `fast-uri` option vulnerability scanners like Orca will still pick up the CVE-2021-44906 vulnerability in uri-js dependency (transitive) since there's still an import `import type {URIComponents} from "uri-js"` on line 3 of resolve.ts??

zekth:
@fetobasic this is a `dev-dependency` so it's not applicable as vulnerability

fetobasic:
> @fetobasic this is a `dev-dependency` so it's not applicable as vulnerability

@zekth Thanks for the quick response, but I meant `uri-js` not fast-uri. Uri-js is still a regular dependency in Ajv and uri-js has a transitive dependency on Minimist which contains the vulnerability. 

I think I answered my own question the fact that uri-js is still a regular dependency for Ajv, scanners will pick it up, even if you avoid using the library by utilizing the optional uriresolver like fast-uri?

zekth:
Sorry i misunderstood. Yes you're right unfortunately.

PR #1918, opened 2022-02-28
Update limitProperties.ts

TIBCOeddie:
in absence of condition pluralization of message, at least have item(s)

### DIFF
--- a/lib/vocabularies/validation/limitProperties.ts
+++ b/lib/vocabularies/validation/limitProperties.ts
@@ -5,7 +5,7 @@ import {_, str, operators} from "../../compile/codegen"
 const error: KeywordErrorDefinition = {
   message({keyword, schemaCode}) {
     const comp = keyword === "maxProperties" ? "more" : "fewer"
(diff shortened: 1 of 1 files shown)


lib/vocabularies/validation/limitProperties.ts
```diff
@@ -5,7 +5,7 @@ import {_, str, operators} from "../../compile/codegen"
 const error: KeywordErrorDefinition = {
   message({keyword, schemaCode}) {
     const comp = keyword === "maxProperties" ? "more" : "fewer"
-    return str`must NOT have ${comp} than ${schemaCode} items`
+    return str`must NOT have ${comp} than ${schemaCode} item(s)`
```
epoberezkin:
```suggestion
    return str`must NOT have ${comp} than ${schemaCode} properties`
```

TIBCOeddie:
You lost the purpose of the initial request... to be ready for "0 properties" "1 property" "2+ properties". I mean, I'd suggest you use a library or util that can conditionalize correctly, even a simple ${schemaCode} == 1 ? "property" : "properties" would do...

Issue #1790, opened 2021-10-15
ValidationError for missing property even though it's not required


Linksku:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**

8.6.3

**Ajv options object**

None

**JSON Schema**

```javascript
{
  type: 'object',
  required: [ 'userId', 'metaKey', 'metaValue' ],
  properties: {
    id: { type: 'integer', minimum: 1 },
    userId: { type: 'integer', minimum: 1 },
    metaKey: { type: 'string', minLength: 1, maxLength: 30 },
    metaValue: { type: 'string' }
  },
  additionalProperties: false
}
```

**Sample data**

```javascript
{
  userId: 3840,
  metaKey: 'LAST_ACTIVITY_TIME',
  metaValue: '2021-10-15 12:19:36'
}
```

**Your code**

UserMeta is an Objection object.

```javascript
UserMeta.query()
      .insert({
        userId: currentUserId,
        metaKey: LAST_ACTIVITY_TIME,
        metaValue: toMysqlDateTime(new Date()),
      })
```

**Validation result, data AFTER validation, error messages**

```
ValidationError: id: should be integer
```

**What results did you expect?**

No error

**Additional info**

This was working fine until recently. I didn't update any packages and I checked all my recent code carefully. Below is part of the compiled validation function produced by Ajv. It's not allowing `id` to be `undefined`:

```
var data1 = data.id;
var errs_1 = errors;
if ((typeof data1 !== "number" || (data1 % 1) || data1 !== data1)) {
    var err = {
        keyword: 'type',
        dataPath: (dataPath || '') + '.id',
        schemaPath: '#/properties/id/type',
        params: {
            type: 'integer'
        },
        message: 'should be integer'
    };
    if (vErrors === null)
        vErrors = [err];
    else
        vErrors.push(err);
    errors++;
}
if ((typeof data1 === "number")) {
    if (data1 < 1 || data1 !== data1) {
        var err = {
            keyword: 'minimum',
            dataPath: (dataPath || '') + '.id',
            schemaPath: '#/properties/id/minimum',
            params: {
                comparison: '>=',
                limit: 1,
                exclusive: false
            },
            message: 'should be >= 1'
        };
        if (vErrors === null)
            vErrors = [err];
        else
            vErrors.push(err);
        errors++;
    }
}
```

Linksku:
I realized this is because Objection uses AJV 6, it works after I manually replace it with AJV 8.

PR #1955, opened 2022-04-08
docs: fix link for useDefaults

bradjones1:

I ran into what is essentially https://github.com/ajv-validator/ajv/issues/1790, where I was confused that validation failed but the `errors` property on the validation function was not set. I have tested this and it appears this is due to a missing, but not required, property in the validated object.

Fixes a broken link to the option and adds a clarifying note regarding the `errors` property in this case.

Ensure I'm correct about this behavior :-)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -329,7 +329,7 @@ Options can have properties `separator` (string used to separate errors, ", " by
 
 In case of validation failure, Ajv assigns the array of errors to `errors` property of validation function (or to `errors` property of Ajv instance when `validate` or `validateSchema` methods were called). In case of [asynchronous validation](./guide/async-validation.md), the returned promise is rejected with exception `Ajv.ValidationError` that has `errors` property.
 
-Note, the `errors` property may not be set if validation fails due to a missing property, even if it is not required in the schema. In these cases, consider using [option `useDefaults`](./options.md#options).
+Note, the `errors` value may not be set if validation fails due to a missing property, even if it is not required in the schema. In these cases, consider using [option `useDefaults`](./options.md#options).
 
 ### Error objects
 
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -329,6 +329,8 @@ Options can have properties `separator` (string used to separate errors, ", " by
 
 In case of validation failure, Ajv assigns the array of errors to `errors` property of validation function (or to `errors` property of Ajv instance when `validate` or `validateSchema` methods were called). In case of [asynchronous validation](./guide/async-validation.md), the returned promise is rejected with exception `Ajv.ValidationError` that has `errors` property.
 
+Note, the `errors` property may not be set if validation fails due to a missing property, even if it is not required in the schema. In these cases, consider using [option `useDefaults`](./options.md#options).
```
epoberezkin: I think this comment is misleading - the errors property is always set, but it may not include the error...

bradjones1: Updated `s/property/value`

epoberezkin:
I don't think the correction is correct, #1790 was related to ajv v6.

docs/api.md
```diff
@@ -329,6 +329,8 @@ Options can have properties `separator` (string used to separate errors, ", " by
 
 In case of validation failure, Ajv assigns the array of errors to `errors` property of validation function (or to `errors` property of Ajv instance when `validate` or `validateSchema` methods were called). In case of [asynchronous validation](./guide/async-validation.md), the returned promise is rejected with exception `Ajv.ValidationError` that has `errors` property.
 
+Note, the `errors` value may not be set if validation fails due to a missing property, even if it is not required in the schema. In these cases, consider using [option `useDefaults`](./options.md#options).
+
```
epoberezkin:
```suggestion
```

Issue #2046, opened 2022-07-26
Add support of the anyOf keyword for discriminator


AntonKozachuk:
**What version of Ajv you are you using?**
**8.6.0**

**What problem do you want to solve?**
According to the OpenAPI specification, the `discriminator` object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.

But now the user with valid definition will see an error if `anyOf` or `allOf` was used.

**What do you think is the correct solution to the problem?**
It would be great to have support for at least `oneOf` and `anyOf` in the project.

**Will you be able to implement it?**
Yes, I'm planning to create a PR in the nearest time.

epoberezkin:
allOf keyword implies that all branches must be valid. anyOf that more than one can be valid. Neither makes sense for discriminated union that implies that there will be exactly one schema for each tag value (and the compilation process validates that it is the case).

What is the use case to use discriminator with anyOf? Could you provide an example schema? In the PR test example anyOf can be replaced with oneOf, as subschemas are mutually exclusive, they cannot be valid at the same time.

RomanHotsiy:
@epoberezkin you're right that `anyOf` doesn't make sense here.
But this it about following the spec.

According to the spec:

> The discriminator object is legal only when using one of the composite keywords oneOf, anyOf, allOf.

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#discriminator-object

Some people are using `anyOf` and expect this to work because the spec allows it.
We can open an issue in the spec but I believe that the implementation must follow the spec in the meantime.

Ivo-Evans:
I was using `oneOf` with a discriminator because the discriminator made it easier to interpret the errors, but I've had to switch to `anyOf` for performance (based on [this comment](https://github.com/ajv-validator/ajv/issues/1705#issuecomment-898863044)). But now I have to deal with the errors. For instance in this example, if I switch from `oneOf` with `discriminator` to `anyOf`, I can't see any way to tell from the errors that `invalidFirst` is an invalid `genusOne` and `invalidSecond` is an invalid `genusTwo`:

```ts
import Ajv from "ajv";

const ajv = new Ajv({
  allErrors: true,
  discriminator: true,
});

const genusOne = {
  type: "object",
  properties: {
    genus: { type: "string", enum: ["a", "b"] },
    species: { type: "string", enum: ["1", "2"] },
  },
};

const genusTwo = {
  type: "object",
  properties: {
[...]

PR #2048, opened 2022-07-27
Add support of the anyOf keyword for discriminator

AntonKozachuk:
#2046

Added support of the `anyOf` keyword for `discriminator` validation.

Don't think so. **


epoberezkin:
see comment in the issue

PR #2079, opened 2022-09-02
ensure that nullable is not set for required parameters

### LINKED ISSUE #2030: JSONSchemaType allows schemas with optional attributes for Types where they are required
Using 8.11.0 (latest)

**Ajv options object**

No options are set
```javascript
export const ajv = new Ajv();
```

**JSON Schema**

```javascript
interface MyExample {
  foo: number;
}
const myExampleSchema: JSONSchemaType<MyExample> = {
  type: 'object',
  properties: {
    foo: {
      type: 'number',
      nullable: true,
    },
  },
  required: [],
}
```

No static analysis issues come out of TypeScript -- it's considered a valid schema for that type.

I expected a TypeScript error saying something along the lines of `"foo?" is incompatible with type "foo"`

I don't think I know how to, but happy to make an effort if this is indeed a valid bug.

erikbrinkman:
nullable was enforced for optional parameters, but not forbidden for
required parameters. This tests and enforces the latter case.

fixes #2030

at least partially

add `nullable?: false` and tests

nope


erikbrinkman:
I should note that typescript 4.8 causes errors with the JTD schema type, so I restricted the typescript version too.

epoberezkin:
Fixed the error in 4.8 now, so made it >= 4.8

PR #2090, opened 2022-09-12
Doc: update the documentation of addKeyword args

TommyDew42:
Update the documentation of the [addKeyword](https://github.com/TommyDew42/ajv/blob/99e884dc4bbb828cf47771b7bbdb14f23193b0b1/lib/core.ts#L585) method. 
The argument should be `string | KeywordDefinition`.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -297,7 +297,7 @@ interface KeywordDefinition {
 }
 ```
 
-If only property `keyword` is provided in the definition object, one can also pass the string value alone to the method as the argument.
+If only the property `keyword` is provided in the definition object, one can also pass the string value alone to the method as the argument.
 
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -245,7 +245,7 @@ Formats can be also added via `formats` option.
 
 <a name="api-addkeyword"></a>
 
-### ajv.addKeyword(definition: object): Ajv
+### ajv.addKeyword(definition: string | KeywordDefinition): Ajv
```
epoberezkin:
```suggestion
### ajv.addKeyword(definition: string | object): Ajv
```

docs/api.md
```diff
@@ -297,6 +297,8 @@ interface KeywordDefinition {
 }
 ```
 
+If only the property `keyword` is provided in the definition object, one can also pass the string value alone to the method as the argument.
```
epoberezkin:
```suggestion
If only the property `keyword` is provided in the definition object, you can also pass the keyword name as the argument.
```

PR #2135, opened 2022-10-17
fix: use WeakMap to cache compiled schemas

stepanho:
Memory leakage of `compile` function cache.

Add new Cache class, which uses WeakMap for object-typed keys.

There is no tests, because I cannot start tests in my workspace.

### DIFF
--- a/lib/core.ts
+++ b/lib/core.ts
@@ -272,6 +272,31 @@ export interface Logger {
   error(...args: unknown[]): unknown
 }
 
+class Cache<K = unknown, V = unknown> {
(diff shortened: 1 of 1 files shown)


stepanho:
Issue: https://github.com/ajv-validator/ajv/issues/2134

sempasha (commented):
Using WeakMap insted of Map looks suitable for caching purpose. Usage of plain Map to cache validation functions for schemas expressed by primitive values (like `true`) is also good point.

sempasha (approved):

PR #2158, opened 2022-11-14
special case empty object for jtd

### LINKED ISSUE #2123: Can't create JTDSchemaType from tagged union with no additional properties
**What I'm trying to do**
I want to use the JTDSchemaType to create a schema for messages I will pass through a websocket. The messages have a `type` field used with a discriminator in the schema. Some of the messages require no additional information, such as a "ping" message, but I am getting an error when creating the `JTDSchemaType` unless there are additional properties.

v8.11

**Ajv options object**

```javascript
{}
```

**Your code**

```typescript
import Ajv, { JTDSchemaType } from "ajv/dist/jtd";

const ajv = new Ajv();

type Message = { type: "ping" } // causes error unless an additional property is added
            |  { type: "error"; message: string };

const schema: JTDSchemaType<Message> = {
  discriminator: "type",
  mapping: {
    ping: {
      properties: {}, // compile error because ping is type 'never'
    },
    error: {
      properties: {
        message: { type: "string" },
      },
    },
  },
};
```

I expect to be able to create a schema for a tagged union that has no properties except the tag. I can't compile this because typescript expects the mapping of `ping` to be type `never`.

I'm not sure how, but I'm happy to help.

erikbrinkman:

fixes #2123

In short, this special cases empty objects since they have special meaning due to the way typescripts type system works. I'll leave more detail in the issue.

I did some various "clean up" to JTD Schema as well, but the substantiative change is special casing that type and adding tests. I think the test coverage is pretty solid (although clearly not perfect) so I wouldn't be too worries.

### DIFF
--- a/lib/types/jtd-schema.ts
+++ b/lib/types/jtd-schema.ts
@@ -74,7 +74,7 @@ type EnumString<T> = [T] extends [never]
   : null
 
 /** true if type is a union of string literals */
-type IsEnum<T> = null extends EnumString<Exclude<T, null>> ? false : true
+type IsEnum<T> = null extends EnumString<T> ? false : true
 
 /** true only if all types are array types (not tuples) */
 // NOTE relies on the fact that tuples don't have an index at 0.5, but arrays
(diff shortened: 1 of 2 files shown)


epoberezkin:
Great, thank you!

epoberezkin (approved):

epoberezkin:
It failed when merging master in - will check later what happened…

erikbrinkman:
The build failure is
```
Error: node_modules/re2/re2.d.ts(13,13): error TS24[30](https://github.com/ajv-validator/ajv/actions/runs/3507783326/jobs/5875718510#step:4:31): Interface 'RE2' incorrectly extends interface 'RegExp'.
  The types returned by 'exec(...)' are incompatible between these types.
    Type 'RE2ExecArray<string> | null' is not assignable to type 'RegExpExecArray | null'.
      Property '0' is missing in type 'RE2ExecArray<string>' but required in type 'RegExpExecArray'.
```

which seems unrelated to this merge. Seems like a versioning issue, but probably worth investigating :/

Issue #2047, opened 2022-07-27
Syntax for Ajv import?


ajvincent:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**

Version 8.11.0, installed today

**Ajv options object**
None.

**Your code**

```typescript
// ProjectJSON.mts

// this works, https://github.com/microsoft/TypeScript/issues/24847}
import { default as Ajv } from "ajv";

// this doesn't work, but is the currently recommended import line
//import Ajv from "ajv";
/*
_02_passthrough_types/source/ProjectJSON.mts(98,17): error TS2351: This expression is not constructable.
  Type 'typeof import("/home/ajvincent/code-generation/cross-stitch/node_modules/ajv/dist/ajv")' has no construct signatures.
*/
const ajv = new Ajv();
void(ajv);
```

I was bitten on this exact error with https://github.com/dsherret/code-block-writer/issues/42 a few weeks ago.

**Are you going to resolve the issue?**

I can, by adding a little bit of documentation.  I just need to know where people want it, if they don't want to quickly fix this themselves.

epoberezkin:
difficult to say what is incorrect without seeing your TS configuration.

`import Ajv from "ajv"` works with the TS config that ajv uses - please check it.

ajvincent:
> difficult to say what is incorrect without seeing your TS configuration.
> 
> `import Ajv from "ajv"` works with the TS config that ajv uses - please check it.

Here's what I have:
```json
{
  "compilerOptions": {
    "lib": ["es2022"],
    "module": "es2022",
    "target": "es2022",
    "moduleResolution": "node",
    "sourceMap": true,
    "declaration": true,

    "experimentalDecorators": true
  },

  "exclude": [
    "spec/**"
  ],

  "extends": "@tsconfig/node16/tsconfig.json"
}

```

The file is https://github.com/ajvincent/cross-stitch/blob/main/_02_passthrough_types/source/ProjectJSON.mts

XC-Zhang:
This configuration works for me:
```json
{
    "compilerOptions": {
        "esModuleInterop": true,
        "module": "node16",
        "moduleResolution": "node",
        "sourceMap": true,
        "strict": true
    }
}
```

Maybe both `esModuleInterop` and `moduleResolution` are required?

kalvenschraut:
```ts
 import { default as Ajv } from 'ajv'
```
[...]

PR #2164, opened 2022-11-14
Export Ajv as a named export

jhford-scout24:
This is solving an issue where it's impossible for me to import Ajv in a typescript project.  This solves #2047 for me

import { Ajv } from 'ajv';

#2047 

Added a single line exporting `Ajv` in a slightly different way.  I'm not sure if there are any tests on the exported interface.

I don't think so -- this should be a no-op for those who don't use it, and the test suite (`yarn test`) passes.

### DIFF
--- a/lib/ajv.ts
+++ b/lib/ajv.ts
@@ -36,6 +36,8 @@ Object.defineProperty(exports, "__esModule", {value: true})
 
 export default Ajv
 
+export {Ajv}
+
 export {
   Format,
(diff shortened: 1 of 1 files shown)


ajvincent:
This feels wrong, adding another export.

I think you might want to consult with @dsherret and ask him how he fixed https://github.com/dsherret/code-block-writer/issues/42 - both the wrong way and the right way.

PR #2166, opened 2022-11-15
chore: provide an esm distribution

me4502:

https://github.com/ajv-validator/ajv/issues/2015

I setup the build script to create a second distribution folder containing esm. This is done by running tsc with two flags passed to it, that alter the output directory, and the module format. The rest uses the existing tsconfig.

I passed the `module` entrypoint of the package json to this distribution, which allows anything that supports ESM to consume a native ESM version. This allows better tree shaking and scope hoisting for supported environments.

This makes no alterations to existing distributions, and the bundles are left unchanged. This only affects the package when consumed via the package.json file in an environment supporting ESM.

I'm a bit confused about why files in the runtime folder assign a `code` property that contains a require statement in a string. Currently this continues to point to the CJS variants of the files in case this is required

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ node_modules
 
 # compiled typescript
 dist/
+dist-esm/
 
 # browser bundles
 bundle/
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,6 +3,7 @@ spec/json-typedef-spec
 .browser
 coverage
 dist
+dist-esm
 bundle
 .nyc_output
 spec/_json
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "8.11.2",
   "description": "Another JSON Schema Validator",
   "main": "dist/ajv.js",
+  "module": "dist-esm/ajv.js",
   "types": "dist/ajv.d.ts",
   "files": [
     "lib/",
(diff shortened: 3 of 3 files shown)


cyberwombat:
I tried building that but had errors with running it. First the `files` array in package.json needs `dist-esm` but after doing so this is what I got trying to build with it - I think some fundamental refactoring is needed before esm ready:
```
Calling "traverse" will crash at run-time because it's an import namespace object, not a function [call-import-namespace]

    node_modules/ajv/dist-esm/compile/resolve.js:95:4:
      95 │     traverse(schema, { allKeys: true }, (sch, jsonPtr, _, parentJsonPtr) => {
         ╵     ~~~~~~~~

  Consider changing "traverse" to a default import instead:

    node_modules/ajv/dist-esm/compile/resolve.js:3:7:
      3 │ import * as traverse from "json-schema-traverse";
        │        ~~~~~~~~~~~~~
        ╵        traverse

▲ [WARNING] Calling "equal" will crash at run-time because it's an import namespace object, not a function [call-import-namespace]

    node_modules/ajv/dist-esm/compile/resolve.js:139:35:
      139 │         if (sch2 !== undefined && !equal(sch1, sch2))
          ╵                                    ~~~~~

  Consider changing "equal" to a default import instead:

    node_modules/ajv/dist-esm/compile/resolve.js:2:7:
      2 │ import * as equal from "fast-deep-equal";
        │        ~~~~~~~~~~
        ╵        equal

✘ [ERROR] Cannot assign to import "code"

    node_modules/ajv/dist-esm/runtime/uri.js:2:4:
      2 │ uri.code = 'require("ajv/dist/runtime/uri").default';
        ╵     ~~~~

  Imports are immutable in JavaScript. To modify the value of this import, you must export a setter
  function in the imported file (e.g. "setCode") and then import and call that function here
  instead.

✘ [ERROR] Cannot assign to import "code"

    node_modules/ajv/dist-esm/runtime/equal.js:3:6:
      3 │ equal.code = 'require("ajv/dist/runtime/equal").default';
        ╵       ~~~~

  Imports are immutable in JavaScript. To modify the value of this import, you must export a setter
  function in the imported file (e.g. "setCode") and then import and call that function here
  instead.
```

cyberwombat:
`dist-esm` needs to be added to the `files` entry of `package.json`. I had some issues importing it though - my code is all esm but still pulling from `dist`. Not sure what I am missing though. I played around with adding `exports` field without success.

Zamiell:
@epoberezkin Can you please review this pull request? I am trying to update my JavaScript project to use ESM and I use Ajv, so this is a show-stopper for me, and I'm sure many others as well.

dasa:
This looks like a great addition 👍🏻 Has anyone found an alternative solution?

phoenix-oes:
Any updates on this? This is a showstopper for us as well.

PR #2206, opened 2023-01-25
`satisfies` does not properly enforce enum on `SomeJTDSchemaType`

### LINKED ISSUE #2205: Typescript 4.9 `satisfies` does not properly enforce enum on `SomeJTDSchemaType`
Ajv version `8.12.0`

Typescript 4.9 allows us to use the `satisfies` operator, which can be used to enforce the correctness of a schema through `SomeJTDSchemaType` before sending it to `JTDDataType`:

```typescript
const schema1 = { type: 'string' } as const satisfies SomeJTDSchemaType;

type Schema1 = JTDDataType<typeof schema1>;

// No error, Schema1 = string as expected


const schema2 = { type: 'something_else' } as const satisfies SomeJTDSchemaType;

type Schema2 = JTDDataType<typeof schema2>;

// Error on the `satisfies SomeJTDSchemaType`, as expected:
//   Type '{ readonly type: "something_else"; }' does not satisfy the expected type 'SomeJTDSchemaType'.
//   Types of property 'type' are incompatible.
//   Type '"something_else"' is not assignable to type '"boolean" | NumberType | StringType | undefined'.
```

This feature does not work properly with `enum` types:

```typescript
const schema3 = { enum: ['apple', 'banana'] } as const satisfies SomeJTDSchemaType;

type Schema3 = JTDDataType<typeof schema3>;

// Although this should be fine, we have an unexpected error on `satisfies SomeJTDSchemaType`:
//   Type '{ readonly enum: readonly ["apple", "banana"]; }' does not satisfy the expected type 'SomeJTDSchemaType'.
//   Types of property 'enum' are incompatible.
//   The type 'readonly ["apple", "banana"]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.
```

In fact, in `types/jtd-schema.ts:14`, we see that `SomeJTDSchemaType` contains `{enum: string[]}`, to which `readonly [...]` cannot be assigned to.

This seems fixable by simply changing the enum type in this file to `{enum: readonly string[]}`.

Can I open a PR for that?

sinderw:

Closes #2205 

Made the type of string enums readonly in `SomeJTDSchemaType`

Also documented the use of `JTDDataType` to extract types from schema, and the use of `SomeJTDSchemaType` to ensure a schema is properly formed before sending it to `JTDDataType`

This only impacts the typescript typings used in a way that was not documented before ; no impact on the result of the unit tests ; so I don't think so

### DIFF
--- a/docs/json-type-definition.md
+++ b/docs/json-type-definition.md
@@ -379,6 +379,52 @@ const schema: JTDSchemaType<LinkedList, {node: LinkedList}> = {
 `JTDSchemaType` currently validates that if the schema compiles it will verify an accurate type, but there are a few places with potentially unexpected behavior.
 `JTDSchemaType` doesn't verify the schema is correct. It won't reject schemas that definitions anywhere by the root, and it won't reject discriminator schemas that still define the descriminator in mapping properties. It also won't verify that enum schemas have every enum member as this isn't generally feasible in typescript yet.
 
+## JTDDataType
+
+The type `JTDDataType` can be used to transform a written schema to a type you can expect to be valid. This type is strict such that if typescript compiles, you should require no further type guards. The downside of this is that the types that `JTDDataType` can verify are limited to the types that JTD can verify. If a type doesn't verify, `JTDDataType` should resolve to `never`, throwing an error when you try to assign to it. This means that types like `1 | 2 | 3`, or general untagged unions (outside of unions of string literals) cannot be used with `JTDDataType`.
+
+### Most Schemas
+
+Most straightforward schemas should work with `JTDDataType`, e.g.
+
+```typescript
+const schema = {
+  properties: {
+    num: {type: "float64"},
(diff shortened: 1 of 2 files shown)


erikbrinkman:
@epoberezkin this seems good to me. It's so easy to forget `readonly`... Thanks for the fix!

docs/json-type-definition.md
```diff
@@ -379,6 +379,52 @@ const schema: JTDSchemaType<LinkedList, {node: LinkedList}> = {
...
+const schema = {
+  properties: {
+    num: {type: "float64"},
+    nullableEnum: {enum: ["v1.0", "v1.2"], nullable: true},
+    values: {values: {type: "int32"}},
+  },
+  optionalProperties: {
+    optionalStr: {type: "string"},
+  },
+} as const satisfies SomeJTDSchemaType
```
loucadufault, line 421:
```suggestion
} satisfies SomeJTDSchemaType
```

PR #2262, opened 2023-05-01
Support for discriminator tag lookup in referenced schemas

z6n9n:
https://github.com/ajv-validator/ajv/issues/2261

Support for discriminator tag lookup in referenced schemas

no

### DIFF
--- a/lib/vocabularies/discriminator/index.ts
+++ b/lib/vocabularies/discriminator/index.ts
@@ -1,7 +1,7 @@
 import type {CodeKeywordDefinition, AnySchemaObject, KeywordErrorDefinition} from "../../types"
 import type {KeywordCxt} from "../../compile/validate"
 import {_, getProperty, Name} from "../../compile/codegen"
-import {DiscrError, DiscrErrorObj} from "../discriminator/types"
+import {DiscrError, DiscrErrorObj} from "./types"
 import {resolveRef, SchemaEnv} from "../../compile"
 import {schemaHasRulesButRef} from "../../compile/util"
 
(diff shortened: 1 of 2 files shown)


mefellows:
Is this PR defunct? If you'd consider merging it with changes, I'd be open to contributing the changes back in. We would love to have it support `discriminator` in https://github.com/pactflow/swagger-mock-validator/issues/51.

lib/vocabularies/discriminator/index.ts
```diff
@@ -69,18 +69,46 @@ const def: CodeKeywordDefinition = {
           sch = resolveRef.call(it.self, it.schemaEnv.root, it.baseId, sch?.$ref)
           if (sch instanceof SchemaEnv) sch = sch.schema
         }
-        const propSch = sch?.properties?.[tagName]
-        if (typeof propSch != "object") {
+        let propSch = sch?.properties?.[tagName]
+        let hasSubSchRequired = false
+        if (!propSch && sch?.allOf) {
+          const {hasRequired, propertyObject} = mapDiscriminatorFromAllOf(propSch, sch)
```
zoidyzoidzoid, line 75: Re: the GitHubActions warnings, maybe we can rename this key or avoid the destructuring and just use a constant for the returned object and reference each key in the following lines?

E.g.

```js
const discriminator = mapDiscriminatorFromAllOf(propSch, sch)
hasSubSchRequired = discriminator.hasRequired
propSch = discriminator.propertyObject
```

zoidyzoidzoid:
@mefellows What changes would be necessary to get this merged? I'm happy to send another patch with those included.

mefellows:
I'm not a maintainer here, I'm just an interested party and are open helping out with direction from the project.

zoidyzoidzoid:
Whoops, my bad. I confused contributor of the pactflow repo with this one!

@epoberezkin would you be able to answer that question?

lib/vocabularies/discriminator/index.ts
```diff
@@ -69,18 +69,46 @@ const def: CodeKeywordDefinition = {
...
+        for (const subSch of sch.allOf) {
+          if (subSch?.properties) {
+            propSch = subSch.properties[tagName]
+            subSchObj = subSch
+          } else if (subSch?.$ref) {
+            subSchObj = resolveRef.call(it.self, it.schemaEnv.root, it.baseId, subSch.$ref)
+            if (subSchObj instanceof SchemaEnv) subSchObj = subSchObj.schema
+            propSch = subSchObj?.properties?.[tagName]
+          }
+          if (propSch) {
```
VolodymyrMachekhin, line 104: is it possible to move the check out of `for`? The reason is that currently the first match is applied and it is better to return the last match. Therefore the last override has priority

lifeiscontent:
@josdejong any chance we could get this merged / looked at? it's been over a year and for openapi specs that use discriminator, `mapping` is almost required in most cases due to how ref can potentially be rewritten it's not as simple as plugging in the values to a static JSON Schema itself.

lifeiscontent:
@epoberezkin  any chance we could get this merged / looked at? it's been over a year and for openapi specs that use `discriminator`, `mapping` is required in a lot of cases due to how ref can potentially be rewritten by Ajv it's not as simple as plugging in the values to a static JSON Schema itself.

kand:
@epoberezkin checking in as well, since we'd like to use `swagger-mock-validator` but `mapping` is an OpenAPI feature we use quite a bit. Could this be merged?

https://github.com/pactflow/swagger-mock-validator/issues/51

PR #2263, opened 2023-05-01
Support OpenAPI 3.0 discriminator.mapping

### LINKED ISSUE #2003: Support OpenAPI 3.0 `discriminator.mapping`

8.11.0

I'd like to validate JSON schema extracted from OpenAPI 3.0 that includes `discriminator.mapping`. Currently having it will make AJV throw a unavoidable `discriminator: mapping is not supported` error. Example code that throws:

```js
import Ajv from "ajv";

const schema = {
  type: "object",
  properties: {
    type: {
      type: "string"
    },
    value: {
      oneOf: [
        {type: "string"},
        {type: "number"},
      ],
      discriminator: {
        propertyName: "type",
        mapping: {
          string: {type: "string"},
          number: {type: "number"},
        },
      },
    },
  },
};

const ajv = new Ajv({discriminator: true});
const validate = ajv.compile(schema); // Error: discriminator: mapping is not supported
console.info(validate({type: "string", value: "foo"}));
```

Either implement `discriminator.mapping` [as described here](https://swagger.io/specification/#discriminator-object) or alternatively gracefully ignore the presence of it so user code can implement this feature itself.

I could probably make it gracefully ignore the property.

z6n9n:
resolves https://github.com/ajv-validator/ajv/issues/2003

option to ignore additional mappings for discriminator. Mappings are not needed in AJV, but they are passed from OAS and should be ignored.

The provided documentation


zekth:
@epoberezkin could you check this PR?

jeffmay:
@epoberezkin Any update on this one? I'm about to implement a work-around using extensions and `jq` to transform the spec (since [`oapi-codegen`](https://github.com/deepmap/oapi-codegen) requires the `mappings` object, but it would be nice to be able to produce a spec that doesn't require these extensions.

afmhenry:
I've been running into this issue as well--is this being reviewed/considered for approval?

VolodymyrMachekhin:
I have to use patch as a workaround while it is not approved

samkim130:
@epoberezkin  bump on this! Seems like a feasible solution

Typiqally:
Hi @jeffmay or @VolodymyrMachekhin, could you explain how you applied the work around?

schimini:
> Hi @jeffmay or @VolodymyrMachekhin, could you explain how you applied the work around?

probably using this: https://www.npmjs.com/package/patch-package

matthias-praxipal:
Came across this as Pydantic generates discriminator.mapping for all union types, would be great to have this merged!

PR #2297, opened 2023-06-27
fix: coerce whitespace strings correctly

vwong:

Number/Integer strings with only whitespace gets coerced to `0`, when it should be `undefined`

The code already skips coercion on empty zero-length strings, I refined this to skip coercion on strings that are empty *after* `.trim()`.

No


vwong:
The CI is broken, but I'm not sure what I would've done to cause it to break. Help?

PR #2308, opened 2023-07-20
docs: Clarify keyword name constraints for `addKeyword`

### LINKED ISSUE #2217: Docs: addKeyword argument
8.11.2

**What is the problem**

[The docs](https://ajv.js.org/api.html#ajv-addkeyword-definition-string-object-ajv) are missing, that `":"` also is an allowed non-leading character in custom keywords added via `addKeyword()`. It probably also is worth mentioning, that *letter* means an ASCII letter.

RegEx for ref: `/^[a-z_$][a-z0-9_$:-]*$/i` at [Line](https://github.com/ajv-validator/ajv/blob/b3e0cb17d0e095b5c883042b2306571be5ec86b7/lib/core.ts#L819)

NitantP:
Closes #2217.

Based on the [`KEYWORD_NAME` regex](https://github.com/ajv-validator/ajv/blob/b3e0cb17d0e095b5c883042b2306571be5ec86b7/lib/core.ts#L819):
- Mentioned `:` as a valid character
- Specified that "letters" refers to the ASCII character set

No.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -251,7 +251,7 @@ Add validation keyword to Ajv instance.
 
 Keyword should be different from all standard JSON Schema keywords and different from previously defined keywords. There is no way to redefine keywords or to remove keyword definition from the instance.
 
-Keyword must start with a letter, `_` or `$`, and may continue with letters, numbers, `_`, `$`, `-`, or `:`.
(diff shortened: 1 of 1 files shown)


NitantP:
Not sure why the CI is running into issues here, but it seems to be a broader issue with other recent PRs as well. Seems like this is the relevant issue from the logs for the failing Node.js 14.x stage:

```
Error: Cannot find module './build/Release/re2.node'
```

PR #2316, opened 2023-07-29
add refs variable for jtd validation

### LINKED ISSUE #2167: Type of sibling property of a ref property is unknown when validated via JTDSchemaType

8.11.2 (latest as of this writing)

**Ajv options object**

```ts
import Ajv, { JTDSchemaType } from "ajv/dist/jtd";
const ajv = new Ajv();
```

**JTD Schema**

```ts
type MyType = {
  referenced: number;
  stringProp: string;
};

const schema: JTDSchemaType<MyType, { num: number }> = {
  definitions: {
    num: { type: "int32" }
  },
  properties: {
    referenced: { ref: "num" },
    stringProp: { type: "string" }
  }
} as const;

```

**Sample data**

```ts
const data = {} as any;
```

**Your code**

```ts
const validate = ajv.compile(schema);

if (validate(data)) {
  const revealType: never = data;
  const stringProp: string = data.stringProp;
  console.log(stringProp);
}
```

https://replit.com/@ento/IntrepidIndolentOutput#index.ts

^`strictNullChecks` is enabled in `tsconfig.json`

```json
		"strictNullChecks": true, 
```

TypeScript's typecheck errors:

```
// const revealType
Type '{ referenced: number; stringProp: unknown; } & {}' is not assignable to type 'never'.

// const stringProp
Type 'unknown' is not assignable to type 'string'.
```

The type of `stringProp` to be `string` after being narrowed by `validate(data)` 

My current workaround is to hold the definition in a const and use it to compose the schema object. (The actual code is a bit more complex than this.)

```ts
const numSchema = { type: "int32" } as const;

const schema: JTDSchemaType<MyType> = {
  properties: {
    referenced: numSchema,
    stringProp: { type: "string" }
  }
} as const;
```

erikbrinkman:
There seems to be some weird behavior with typescript when refs was left to default. By specifying it manually the behavior was fixed.

fixes #2167

- Add `R extends Record<string, unknown>` to compile and validate for JTD schemas that resolves the referenced issue.
- Add tests that failed prior to the change
- Added `--cache` to prettier invocation. This is not related to the diff, but a nice quality of life change. I'll remove if requested

1. If you're cool with adding `--cache` to prettier
2. `compile` and `validate` have really complicated types. This change only resolved the existing tests, but we know coverage is not 100%, so if anything comes to mind that this might hurt.

with or without cache, prettier was tweaking files that broke typescript. I reverted the changes here to get the pull request in, but it's worth calling out

### DIFF
--- a/lib/core.ts
+++ b/lib/core.ts
@@ -340,7 +340,10 @@ export default class Ajv {
   validate<T>(schema: Schema | JSONSchemaType<T> | string, data: unknown): data is T
   // Separated for type inference to work
   // eslint-disable-next-line @typescript-eslint/unified-signatures
-  validate<T>(schema: JTDSchemaType<T>, data: unknown): data is T
+  validate<T, R extends Record<string, unknown>>(
+    schema: JTDSchemaType<T, R>,
+    data: unknown
+  ): data is T
   // This overload is only intended for typescript inference, the first
   // argument prevents manual type annotation from matching this overload
   validate<N extends never, T extends SomeJTDSchemaType>(
@@ -371,7 +374,10 @@ export default class Ajv {
   compile<T = unknown>(schema: Schema | JSONSchemaType<T>, _meta?: boolean): ValidateFunction<T>
   // Separated for type inference to work
   // eslint-disable-next-line @typescript-eslint/unified-signatures
-  compile<T = unknown>(schema: JTDSchemaType<T>, _meta?: boolean): ValidateFunction<T>
+  compile<T = unknown, R extends Record<string, unknown> = Record<string, unknown>>(
+    schema: JTDSchemaType<T, R>,
+    _meta?: boolean
+  ): ValidateFunction<T>
   // This overload is only intended for typescript inference, the first
   // argument prevents manual type annotation from matching this overload
   compile<N extends never, T extends SomeJTDSchemaType>(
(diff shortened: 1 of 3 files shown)


erikbrinkman:
@epoberezkin something seems to have changed recently, and tests are failing silently at the build stage with:

```
Trying https://github.com/uhop/node-re2/releases/download/1.20.1/linux-x64-83.br ...
Trying https://github.com/uhop/node-re2/releases/download/1.20.1/linux-x64-83.gz ...
Building locally ...

> re2@1.20.1 rebuild /home/runner/work/ajv/ajv/node_modules/re2
> node-gyp rebuild

make: Entering directory '/home/runner/work/ajv/ajv/node_modules/re2/build'
  CXX(target) Release/obj.target/re2/lib/addon.o
  CXX(target) Release/obj.target/re2/lib/new.o
../lib/new.cc: In static member function ‘static Nan::NAN_METHOD_RETURN_TYPE WrappedRE2::New(Nan::NAN_METHOD_ARGS_TYPE)’:
../lib/new.cc:313:55: error: ‘kHasIndices’ is not a member of ‘v8::RegExp’
  313 |                 hasIndices = bool(flags & v8::RegExp::kHasIndices);
      |                                                       ^~~~~~~~~~~
```

In general this seems like a version incompatibility between the local version of v8 and re2, but looking at successful builds, it looks like they were able to fetch the appropriate binary, so maybe the fix is getting that request to go through. The location seems to exist, so it must be some firewalling on githubs side.

Otherwise, `nor run test-ci` works locally for me.

PR #2325, opened 2023-08-31
docs: modify the link of `Error logging` to the right address

LDT-dream:

A link in the dossier doesn't point to the right address. 

I modified the link to point to the correct address.

N/A


jasoniangreen:
Thanks for your PR but this has been fixed now.

PR #2356, opened 2023-12-20
fix: Update parseJson() for v8 11.7.72 and newer

### LINKED ISSUE #2355: JSON parsing fails for v8 versions 11.7.72 and newer
v8.12.0

**Ajv options object**
not used

**JSON Schema**

```json
{
  "properties": {
    "test": { "type": "uint8" }
  },
  "additionalProperties": true
}
```

**Sample data**

```json
{
      "test": 1,
      "error": "now"
}
```

**Your code**

```javascript
const Ajv = require("ajv/dist/jtd");

const JTDschema = {
  properties: {
    test: { type: "uint8" },
  },
  additionalProperties: true,
};

const payload = `{
        "test": 1,
        "error": "now"
  }`;

const ajv = new Ajv();

const parser = ajv.compileParser(JTDschema);

const output = parser(payload);

if (!output) {
  console.log(parser.position);
  console.log(parser.message);
}
```

**Console output**

```
47
unexpected end
```

Validates successfully, no console output

Yes, patch is already written, PR incoming

**Additional context**
v8 11.7.72 changed the message templates for JSON parsing related error messages (https://chromium-review.googlesource.com/c/v8/v8/+/4652014)  
This breaks `parseJson()` and leads to parsing errors if utilizing `additionalProperties` in a JSONTypeDefinition (other issues not specifically tested).  
`npm test` also produces several errors with the mentioned v8 version.  
This can be tested with NodeJS v21.0.0 or newer and deno v1.36.1 or newer since those versions updated their respective v8 versions past the affected version

egfx-notifications:
Fixes #2355

Update regex to extract end position from error in `parseJson()` - the updated regex works both for the previous and the new v8 error messages

https://chromium-review.googlesource.com/c/v8/v8/+/3513684 also added some changes to the messages with some of the new message templates not mentioning a position at all. I'm not sure if this would affect ajv, but since `npm test` completes successfully with my patch I assume it does not.

### DIFF
--- a/lib/runtime/parseJson.ts
+++ b/lib/runtime/parseJson.ts
@@ -1,4 +1,4 @@
-const rxParseJson = /position\s(\d+)$/
+const rxParseJson = /position\s(\d+)(?: \(line \d+ column \d+\))?$/
 
 export function parseJson(s: string, pos: number): unknown {
   let endPos: number | undefined


jasoniangreen:
I have verified this on https://regex101.com/, looks good. I will look to get this fix in after we get the builds working again.

jasoniangreen (approved):

PR #2365, opened 2024-01-08
Fix types for NodeNext

### LINKED ISSUE #2132: Typescipt and ESM yields "Ajv This expression is not constructable"
I have a TS project in ESM mode (type:module) and getting VSCode errors using certain modules such as Ajv..

Given this in my tsconfig.json:
```
{  
  "compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "nodenext",
    "esModuleInterop": true
    // I had a lot more stuff but the top two lines are the ones causing issue (Node16 is same)
  }
}
```

package.json
```
{
  "name": "test",
  "type": "module",
  "dependencies": {
    "ajv": "^8.11.0"
  },

  "devDependencies": {
    "@tsconfig/node16": "^1.0.3",
    "@types/node": "^18.11.0",
    "typescript": "^4.8.3"
  }
}
```
and finally code:
```
import Ajv from "ajv"
export const getClient = ():Ajv => { 
  return  new Ajv()
}
```
I get the error as attached
Screen Shot 2022-10-15 at 9 58 18 PM

Note that this code compiles correctly.

If I do:
```
import  Ajv from "ajv"
export const getClient = ():Ajv.default => { 
  return  new Ajv.default()
}
```
Then error goes away but it does not run.

I think this is something to do with CJS modules needing to add something for esm export but not sure exactly what. 

Here's a [sandbox](https://codesandbox.io/p/github/cyberwombat/ts-esm-test/draft/kind-sea?file=%2Findex.ts&workspace=%257B%2522activeFileId%2522%253A%2522cl9bh5ug20002lrhb6woealfc%2522%252C%2522openFiles%2522%253A%255B%255D%252C%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522gitSidebarPanel%2522%253A%2522COMMIT%2522%252C%2522sidekickItems%2522%253A%255B%257B%2522type%2522%253A%2522TERMINAL%2522%252C%2522shellId%2522%253A%2522cl9bh6bqv000blrhb74jv6h3z%2522%252C%2522key%2522%253A%2522cl9bh6b2o00523b6fhlt7a3mv%2522%252C%2522isMinimized%2522%253Afalse%257D%255D%257D)

benasher44:

Closes #2132

Changed the way types are exported to improve support for users `module: node16/nodenext` users.


lib/ajv.ts
```diff
@@ -31,39 +34,65 @@ class Ajv extends AjvCore {
   }
 }
 
-module.exports = exports = Ajv
+export = Ajv
 Object.defineProperty(exports, "__esModule", {value: true})
```
andrewbranch:
```suggestion
Object.defineProperty(Ajv, "__esModule", {value: true})
```

The `export =` line causes a `module.exports = Ajv` to be written at the end of the file, so these mutations of the free variable `exports` aren’t going anywhere anymore. I missed this in my original suggestion.

benasher44:
@epoberezkin I think this is ready for review now, and I'm now just hitting the node 14 build issue plaguing the main branch. Maybe we can disable now that node 14 is EOL?

benasher44:
Okay yep this is ready. You can see the build passing on the other builds in https://github.com/ajv-validator/ajv/pull/2365/commits/b7b5fd0861baa89d702d7d36e01c0a76eb4b50c3, where the only change I did was temporarily remove 14 from the matrix.

lib/ajv.ts
```diff
@@ -31,39 +31,75 @@ class Ajv extends AjvCore {
...
+import * as compile from "./compile"
+import * as compileValidate from "./compile/validate"
+import * as vocabulariesErrors from "./vocabularies/errors"
+import * as compileRules from "./compile/rules"
+import * as typesJsonSchema from "./types/json-schema"
+import * as compileCodegen from "./compile/codegen"
+import * as runtimeValidationError from "./runtime/validation_error"
+import * as compileRefError from "./compile/ref_error"
+
+// eslint-disable-next-line @typescript-eslint/no-namespace, no-redeclare
```
benasher44, line 55: seems typescript-eslint isn't very aware of the `export import` feature in namespaces https://github.com/typescript-eslint/typescript-eslint/issues/4129

benasher44:
Pushed more updates + confirmed the build pass without node 14

benasher44:
Thanks @jasoniangreen! Let me know if you have feedback on this PR — would love to move this forward

jasoniangreen:
> Thanks @jasoniangreen! Let me know if you have feedback on this PR — would love to move this forward

Let me update your branch with master and have a look.

jasoniangreen:
Something that hasn't been considered with this change is the browser bundles we generate. With this PR the rollup command, which is used to generate browser bundles, doesn't work. 
Also there are a lot of changes in this PR, and after speaking to epoberezkin there is not a lot appetite for such a large change to resolve this, at least not at this time. He has asked me to explore some alternatives, possible based on [this work around](https://github.com/ajv-validator/ajv/issues/2132#issuecomment-1290409907).

Either way we have to be wary of the various exports and ways that AJV is used and ensure we don't make life difficult for a lot of people by changing things unnecessarily.

benasher44:
That's totally fair! Would you consider just adding named exports for the classes alongside the default exports? The main pain point is the default exports. If people have the named exports as an alternative, I think that would satisfy most folks, even if it means the default exports are left clunky.

benasher44:
Opened #2389, which is much smaller. NodeNext users can switch to using the named export instead, side-stepping the default import/export issues.

benasher44:
Closing in favor of #2389

PR #2373, opened 2024-01-28
Change import package

jarqvi:
In fact, some codes that were not needed much have been removed.

The recent removal of the line ‍‍‍`const {default: Ajv} = require('ajv')` was made possible by eliminating the export statement `module.exports.default = AjvClass`. As a result, we no longer require the named import of the Ajv module.

And since we mentioned the following code sample in the documentation, I don't think there is a problem:
`const Ajv = require("ajv")`

require packages in CJS


jarqvi:
Or at least it can be as follows:
`const {Ajv} = require('ajv')`

The following also works:
`const Ajv = require('ajv')`

I feel the code is cleaner with this method instead of using the code below:
`const {default: Ajv} = require('ajv')`

jarqvi:
@jasoniangreen what happened?!

jasoniangreen:
> @jasoniangreen what happened?!

I just wanted to pull master into your branch so I can see if it passes and do a proper review in the up to date context.

jarqvi:
Thanks.

jasoniangreen:
Hi @jarqvi, I don't really understand why should make this change?

jarqvi:
Hi @jasoniangreen, 
I felt that it has more readability when importing package.

jasoniangreen:
> Hi @jasoniangreen, I felt that it has more readability when importing package.

Ah ok, but it would be a potentially breaking change, correct? Depending on how people are currently importing it?

jarqvi:
> > Hi @jasoniangreen, I felt that it has more readability when importing package.
> 
> Ah ok, but it would be a potentially breaking change, correct? Depending on how people are currently importing it?

Well, we can make these changes by keeping the previous method and be backward compatible

jarqvi:
> > > Hi @jasoniangreen, I felt that it has more readability when importing package.
> > 
> > 
> > Ah ok, but it would be a potentially breaking change, correct? Depending on how people are currently importing it?
> 
> Well, we can make these changes by keeping the previous method and be backward compatible

@jasoniangreen Done.

PR #2389, opened 2024-03-01
Add named exports for main classes to improve types for NodeNext

benasher44:

Users of TypeScript + ESM + NodeNext have difficulty using the default exported classes, which is the only way to consume them (#2132).

This additionally adds a named export, which side-steps this issues around default imports in that environment. This is a very lightweight alternative to #2365 to improve things for these users, even if it won't make this package pass [attw](https://arethetypeswrong.github.io/).

I don't think so.

### DIFF
--- a/lib/2019.ts
+++ b/lib/2019.ts
@@ -10,7 +10,7 @@ import addMetaSchema2019 from "./refs/json-schema-2019-09"
 
 const META_SCHEMA_ID = "https://json-schema.org/draft/2019-09/schema"
 
-class Ajv2019 extends AjvCore {
+export class Ajv2019 extends AjvCore {
   constructor(opts: Options = {}) {
     super({
       ...opts,
@@ -44,6 +44,7 @@ class Ajv2019 extends AjvCore {
 }
 
 module.exports = exports = Ajv2019
+module.exports.Ajv2019 = Ajv2019
 Object.defineProperty(exports, "__esModule", {value: true})
 
 export default Ajv2019
--- a/lib/2020.ts
+++ b/lib/2020.ts
@@ -7,7 +7,7 @@ import addMetaSchema2020 from "./refs/json-schema-2020-12"
 
 const META_SCHEMA_ID = "https://json-schema.org/draft/2020-12/schema"
 
-class Ajv2020 extends AjvCore {
+export class Ajv2020 extends AjvCore {
   constructor(opts: Options = {}) {
     super({
       ...opts,
@@ -38,6 +38,7 @@ class Ajv2020 extends AjvCore {
 }
 
 module.exports = exports = Ajv2020
+module.exports.Ajv2020 = Ajv2020
 Object.defineProperty(exports, "__esModule", {value: true})
 
 export default Ajv2020
--- a/lib/ajv.ts
+++ b/lib/ajv.ts
@@ -8,7 +8,7 @@ const META_SUPPORT_DATA = ["/properties"]
 
 const META_SCHEMA_ID = "http://json-schema.org/draft-07/schema"
 
-class Ajv extends AjvCore {
+export class Ajv extends AjvCore {
   _addVocabularies(): void {
     super._addVocabularies()
     draft7Vocabularies.forEach((v) => this.addVocabulary(v))
@@ -32,6 +32,7 @@ class Ajv extends AjvCore {
 }
 
 module.exports = exports = Ajv
+module.exports.Ajv = Ajv
 Object.defineProperty(exports, "__esModule", {value: true})
 
 export default Ajv
--- a/lib/jtd.ts
+++ b/lib/jtd.ts
@@ -35,7 +35,7 @@ type JTDOptions = CurrentOptions & {
   multipleOfPrecision?: never
 }
 
-class Ajv extends AjvCore {
+export class Ajv extends AjvCore {
   constructor(opts: JTDOptions = {}) {
     super({
       ...opts,
@@ -93,6 +93,7 @@ class Ajv extends AjvCore {
 }
 
 module.exports = exports = Ajv
+module.exports.Ajv = Ajv
 Object.defineProperty(exports, "__esModule", {value: true})
 
 export default Ajv


benasher44:
The idea is that once this ships, users having trouble with the default export can switch to the name export instead.

jasoniangreen:
Thanks for coming up with a simpler solution, but just so I understand, why do you need to both add `export` to `export class Ajv2019 extends AjvCore` as well as adding the `.Ajv2019` to exports? And will adding the export keyword conflict with `module.exports = exports = Ajv2019`?

benasher44:
Good questions!

For Ajv2019 (and Ajv2020), that was just for consistency. I don't use them, but without much context, it felt weird to not do it for the others. I'm happy undo those!

As for the module.exports bit, let's look at the final output (dist/ajv.js), for example. Here's the relevant snippet:

```
exports.Ajv = Ajv;
module.exports = exports = Ajv;
module.exports.Ajv = Ajv;
```

First, `exports.Ajv = Ajv` is set and appears to be emitted by tsc. The problem is that the next line `module.exports = exports = Ajv` immediately overwrites that. Without any changes, this would cause a named import of Ajv to crash at runtime, since the Ajv class itself does not come with a static prop (or otherwise) to support the named export. This `module.exports = exports = Ajv` line comes from ajv.ts directly, so tsc is forced to emit it (afaict), despite (and probably unbeknownst to tsc) that is overwrites the named export. To counteract that, I added `module.exports.Ajv = Ajv` on the following line, which effectively adds an `Ajv` prop to the class, re-enabling the named export.

jasoniangreen:
Thanks for the explanation, I will discuss with EP and get back to you. It might take me a couple of days as I have another priority at the moment, but I will definitely get back to you soon.

jasoniangreen:
I haven't forgotten you. Just trying to get a release together with updated deps (as it has been a while), and then I will follow up on this with EP.

benasher44:
No problem. Thanks!

jomaa-daniel:
Waiting on this update as well. 🙏

jasoniangreen:
@epoberezkin FYI

jasoniangreen:
I spoke to EP, I will be preparing a release with this and general dependency updates over the coming days. He has asked me to do some very specific due diligence and tests to reduce the chance of any issues. We have to be super cautious given the huge user base that rely on AJV and the wide variety of ways it is used, imported and built into projects.

jasoniangreen (approved):
I will include this in the next release. There are some manual steps and some extra testing that EP has asked me to conduct, so no timeline yet.

flowluap:
FYI: 

I used those two commits, to patch the package for us, pre release. I am still getting:
```
TS2709: Cannot use namespace  Ajv  as a type.
``` 

This is how I import it...

```
import Ajv from 'ajv';
```  

Could you include a named export for Ajv, usable as type?

benasher44:
The expectation is you would change the import to:

`import { Ajv } from 'ajv';`

flowluap:
@benasher44 thanks for the fast reply. I installed the latest PR regarding the topic:

````
 "ajv": "github:ajv-validator/ajv#pull/2365/head",

```

And that one seems to be working perfectly fine with

```

 "lib": ["es2023"],
 "module": "node16",
 "moduleResolution": "nodenext",
```

Looking forward to the new release, thanks for your work!

benasher44:
👍 can you try with this PR and the named import I suggested? I don't expect that #2365 will ship, at this point.

flowluap:
With the PR installed, using the named import like that:

```
import { Ajv } from 'ajv';

/// TS2595:  Ajv  can only be imported by using a default import.
```

Working fine is the default:

```
import  Ajv  from 'ajv';
```

benasher44:
@flowluap I'm not sure what you're doing, but if I pull master, run `npm i`, then `npm run build`, then `npm run pack`, I get a usable package. From there, I installed the locally built package. With that installed, I can `import { Ajv } from 'ajv'` without issue.

ehaynes99:
For anyone who needs a workaround until this is released, this works and is typesafe. There are similar problems with `ajv-errors` and `ajv-formats`, and those haven't been published in forever.

```typescript
import AjvLib from 'ajv'
import AjvErrorsLib from 'ajv-errors'
import AjvFormats from 'ajv-formats'

const { default: Ajv } = AjvLib
type Ajv = InstanceType<typeof Ajv>

const { default: ajvErrors } = AjvErrorsLib
const { default: addFormats } = AjvFormats

const ajv: Ajv = new Ajv({ allErrors: true, strict: false })
ajvErrors(ajv, { singleError: true })
addFormats(ajv)
```

emmenko:
Thanks for fixing this! ❤️

Is there a plan when this might be released? 🙏

shumphrey:
Now that this has been released, I'm having issues getting ajv-formats working with this change, does a similar change need to be made to ajv-formats?

benasher44:
What's the issue?

shumphrey:
> What's the issue?

> Argument of type 'import("blank/node_modules/ajv/dist/ajv").Ajv' is not assignable to parameter of type 'import("blank/node_modules/ajv-formats/node_modules/ajv/dist/core").default

This is with various different versions of `import addFormats from "ajv-formats";`

Seems like the type has changed and is no longer compatible?

shumphrey:
The old typescript code was:
```ts
import Ajv from "ajv";
import addFormats from "ajv-formats";

const ajv = new Ajv.default({ allowUnionTypes: true });
addFormats.default(ajv);
```

With the new release of ajv:
```ts
import { Ajv } from "ajv";
import addFormats from "ajv-formats";
 
const ajv = new Ajv({ allowUnionTypes: true });
addFormats.default(ajv); // this throws a typescript error
```

I've tried various combos of how we import ajv-formats, I think the typing of the two don't work together any more

benasher44:
Can you please post the typescript error?

benasher44:
This works fine for me with ajv-formats 3.01 and ajv 8.13.0. Can you please post a reproduction?

```typescript
import { Ajv } from "ajv";
import addFormats from "ajv-formats";
addFormats.default(new Ajv());
```

shumphrey:
```
npm ls ajv
│ ├─┬ ajv-formats@3.0.1
│ │ └── ajv@8.12.0
│ └── ajv@8.13.0
```

I believe ajv-formats was bringing in the old ajv, resetting the package-lock resolved it.

Zebiano:
For those wanting to use "JSON Type Definition" with `import`, here's a way:
```ts
import { Ajv } from 'ajv/dist/jtd.js';
```
I'm using the following `tsconfig` options:
```json
"compilerOptions": {
  "strict": true,
  "target": "es2017",
  "module": "Node16",
  "moduleResolution": "Node16",
  "esModuleInterop": true
}
```

I'm obviously not a big fan of having the `.js` import on a `.ts` file, but I'm at the point of "I'll accept anything that works" right now. 😄 

If anyone has suggestions on how I can change my `import` approach, or what can be done to change the `package.json` info of this project (ajv) to make the process of importing `jtd` easier, please let me/us know!

---

Also, it could be worth updating the "[Getting started](https://ajv.js.org/guide/getting-started.html)" section to also cover examples with `import` instead of `require()`!

PR #2409, opened 2024-04-06
fix `tsconfig.json#extends` file reference

rotu:


rotu:
I think this relies on the undocumented behavior here of appending a file name when extends is a path or package: https://github.com/microsoft/TypeScript/issues/51782

PR #2410, opened 2024-04-06
Update workflows

rotu:

Silence warnings about outdated actions e.g. https://github.com/ajv-validator/ajv/actions/runs/8490491525
Switch to `${{}}` syntax in `if` statement for quoting expressions (necessary when condition starts with `!`)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,9 +15,9 @@ jobs:
         node-version: [16.x, 18.x, 20.x, 21.x]
 
     steps:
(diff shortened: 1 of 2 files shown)


jasoniangreen:
This one looks good to me @epoberezkin, it passes the build with the updates to the build script so should be good.

jasoniangreen:
@epoberezkin I am going to merge this as it is something I was planning to do anyway and in the worse case that it breaks, it can only break workflows which will only impact me.

jasoniangreen (approved):

Issue #2350, opened 2023-12-02
Deprecation Warning: punycode Dependency via uri-js


camiloux:
Version 8.12.0, Node 21.2.0, using npm.

Hi there,
I wanted to raise an issue regarding `uri-js` dependency, which uses `punycode` under the hood, which has been deprecated.

While working with ajv, I've noticed deprecation warnings related to the usage of the `punycode` module through its dependency on uri-js. It seems uri-js hasn't been updated for three years, causing these warnings.

├─┬ ajv@8.12.0
│ └─┬ uri-js@4.4.1
│   └── punycode@2.3.0

Just thought I'd bring this up for consideration. Thank you for your work on ajv!

domdomegg:
I've opened a PR to fix this in uri-js, the package ajv depends on here: https://github.com/garycourt/uri-js/pull/95

ThatOneCalculator:
Since `uri-js` seems to be unmaintained, maybe `ajv` should use a fork of `uri-js` (i.e. the repo maintained by domdomegg)

kibertoad:
@ThatOneCalculator I've created a fork that addresses this problem: https://github.com/kibertoad/toad-uri-js

Happy to invite ajv folks to co-maintain if there is interest.

ttodua:
any update on this? 
@kibertoad did you have any response from ajv devs?

andreinwald:
Hi there,
I created a replacement for library uri-js as temporal solution.
- Based on Node.js and browser URL api
- 99% compatible with original URI.js library
- Solves "The punycode module is deprecated" warning in Node
- Tested with libraries: ESLint, Webpack, Ajv (runned all Ajv tests)

https://github.com/andreinwald/uri-js-replace

You can add to package.json of your project:
```json
"overrides": {
    "uri-js": "npm:uri-js-replace"
 }

Niek:
@jasoniangreen Any chance to switch over the dependency to `uri-js-replace`?

jasoniangreen:
Hi @Niek thanks for your suggestion but I think [this](https://github.com/ajv-validator/ajv/pull/2415) will be our most likely move. We already have fast-uri as a second option in the codebase/docs so it makes sense to switch it to be the default.

jasoniangreen:
Closed as we are moving to fast-uri

kibertoad:
@jasoniangreen fast-uri was reverted. What is the plan now?

PR #2377, opened 2024-02-09
Fix punycode deprecation warning

kibertoad:
fixes #2350 #2343

**What issue does this pull request resolve?**

punycode warnings. See linked issues.

**What changes did you make?**

Switched dependency to a fork that addresses the issue, and which is intended to be maintained for a long time.

**Is there anything that requires more attention while reviewing?**

No.

Issue #2343, opened 2023-10-30
Drop dependency on uri-js/punycode


gajus:
**What version of Ajv you are you using?**

v6.12.6

**What problem do you want to solve?**

Logs being flooded with deprecation warnings:

```
contra-api:test:vitest: (Use `node --trace-deprecation ...` to show where the warning was created)
contra-api:test:vitest: (node:747235) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
contra-api:test:vitest: (Use `node --trace-deprecation ...` to show where the warning was created)
contra-api:test:vitest: (node:747235) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
contra-api:test:vitest: (Use `node --trace-deprecation ...` to show where the warning was created)
contra-api:test:vitest: (node:747235) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
contra-api:test:vitest: (Use `node --trace-deprecation ...` to show where the warning was created)
contra-api:test:vitest: (node:747235) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
contra-api:test:vitest: (Use `node --trace-deprecation ...` to show where the warning was created)
contra-api:test:vitest: (node:747235) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
contra-api:test:vitest: (Use `node --trace-deprecation ...` to show where the warning was created)
contra-api:test:vitest: (node:747235) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
contra-api:test:vitest: (Use `node --trace-deprecation ...` to show where the warning was created)
```

**What do you think is the correct solution to problem?**

* Drop Node.js support outside of LTE
* Drop uri-js (which depends on punycode)

**Will you be able to implement it?**

Yes

Related issues:

* https://github.com/garycourt/uri-js/issues/94

Note:

* My understanding is that the only reason for uri-js is backwards compatibility for older Node.js versions.

mikicymulate:
Also happens in newer version 
"ajv": "^8.12.0"
https://github.com/ajv-validator/ajv/issues/2350

domdomegg:
I've opened a PR to fix this in uri-js, the package ajv depends on here: https://github.com/garycourt/uri-js/pull/95

[...]

PR #2415, opened 2024-04-13
Replace `uri-js` with `fast-uri`

vixalien:

This PR fixes issues #2350 and #2343. It removes the (deep) dependency on `punycode`, a deprecated module.

I replaced the default uriResolver to [fast-uri](https://github.com/fastify/fast-uri) instead of [uri-js](https://github.com/garycourt/uri-js).

This PR supersedes #2377.

It is also worth considering using the web API `URL` which is defined in [URL Living Standard](https://url.spec.whatwg.org/) instead of `URI` which is defined in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) and is less prevalent and can resolve the issue raised by @jasoniangreen in the above linked PR:

> ...I have been speaking to @epoberezkin and he wants me to explore removing uri-js. [..] there is just an open question around browser support. I will update you when I have more.

We can use the native `URL` API as it is supported with node 10+ and all major browsers and it has [97.68% caniuse score](https://caniuse.com/url) while `URI` is not standardized afaict.

Thanks!

TODO:

- [ ] rebase
- [ ] fix [spec/resolve.spec.ts](https://github.com/ajv-validator/ajv/pull/2415/files#diff-8eaec48b707b9d12a6c79d05a2bc004c400d48d307c05f3bc86b752f9848a37c)

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -344,7 +344,7 @@ Include human-readable messages in errors. `true` by default. `false` can be pas
 
 ### uriResolver
 
-By default `uriResolver` is undefined and relies on the embedded uriResolver [uri-js](https://github.com/garycourt/uri-js). Pass an object that satisfies the interface [UriResolver](https://github.com/ajv-validator/ajv/blob/master/lib/types/index.ts) to be used in replacement. One alternative is [fast-uri](https://github.com/fastify/fast-uri).
+By default `uriResolver` is undefined and relies on the embedded uriResolver [fast-uri](https://github.com/fastify/fast-uri). Pass an object that satisfies the interface [UriResolver](https://github.com/ajv-validator/ajv/blob/master/lib/types/index.ts) to be used in replacement. One alternative is [uri-js](https://github.com/garycourt/uri-js).
 
 ### code <Badge text="v7" />
 
--- a/lib/compile/index.ts
+++ b/lib/compile/index.ts
@@ -14,7 +14,7 @@ import N from "./names"
 import {LocalRefs, getFullPath, _getFullPath, inlineRef, normalizeId, resolveUrl} from "./resolve"
 import {schemaHasRulesButRef, unescapeFragment} from "./util"
 import {validateFunctionCode} from "./validate"
-import * as URI from "uri-js"
+import * as URI from "fast-uri"
 import {JSONType} from "./rules"
 
 export type SchemaRefs = {
--- a/lib/compile/resolve.ts
+++ b/lib/compile/resolve.ts
@@ -1,6 +1,6 @@
 import type {AnySchema, AnySchemaObject, UriResolver} from "../types"
 import type Ajv from "../ajv"
-import type {URIComponents} from "uri-js"
+import type {URIComponents} from "fast-uri"
 import {eachItem} from "./util"
 import * as equal from "fast-deep-equal"
 import * as traverse from "json-schema-traverse"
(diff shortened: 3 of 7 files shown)


mbtools:
`ajv` is a major contributor to the `[DEP0040] DeprecationWarning: The `punycode` module is deprecated` issue that is popping up everywhere (since it's used in `eslint`). Please fix this. 

Less dependencies. Better performance. This is the way!

vixalien:
I would be very happy if there were an initial review or at least a comment from the AJV team.

tianyingchun:
how progress of this?

tianyingchun:
@epoberezkin

tianyingchun:
how progress of this?

vixalien:
No feedback from maintainers, can't move forward :(

tianyingchun:
I see that there is still a version of this library released recently,

vixalien:
Yes, other pull requests are getting merged and I believe this one is just ignored.

Some feedback - any feedback - from the maintainers would be appreciated

jasoniangreen:
Hi all, I will try and get some guidance from @epoberezkin on this one. Sorry I missed your activity here.

tianyingchun:
Haha, merge and release a version as soon as possible 😁

vixalien:
Ready

jasoniangreen:
Update: After discussing with @epoberezkin, we want to get one more minor release out with some bug fixes and then will release this change as 9.0.0.

jasoniangreen:
> Update: After discussing with @epoberezkin, we want to get one more minor release out with some bug fixes and then will release this change as 9.0.0.

Actually after discussing with EP and understanding the swap, we've decided this can go out as a minor release. I'm going to give it all one last check and do that next.

jasoniangreen (approved):
Looks good to me. I have verified: 
- that uri-js and fast-uri are both built to the same standard [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986)
- that there are test for URNs which was a concern as we have had issues with them in the past
- that both fast-uri and uri-js are still both being run in tests, we have just swapped what is the default

jasoniangreen:
Hmm, something broke the build to master, I think it's related to a bug in the get_contributors script. Investigating.

jasoniangreen:
Well I did just publish 8.15.0 with fast-uri but it's failing to build browser bundles because for some reason fast-uri has this line `require('node:url')`? I have no idea why it's not just `require('url')` but I think I'll have to roll forward with a revert of this change. Will discuss with @epoberezkin.

yuki-js:
FYI
`fast-uri` is not officially said browser-incompatible module itself, so it's no wonder it uses `node:` prefix.

Webpack error message

```
Module build failed: UnhandledSchemeError: Reading from "node:url" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```

voxpelli:
Related: https://github.com/fastify/fast-uri/issues/21

silverwind:
👍 `fast-uri` depends on `node:url` which is no-go for brower compatibilty.

fcma-bbientz:
We are having the same issue with 8.15.0.  Since pulling this package our builds have been failing with the following error:

Module build failed: UnhandledSchemeError: Reading from "node:url"

fcma-bbientz:
In package.json, we had "^18.12.0" and builds worked yesterday, failed today.  Finally tracked it down to this update from 8.14.0 to 8.15.0

silverwind:
See https://github.com/ajv-validator/ajv/issues/2446

PR #2429, opened 2024-05-10
fix #2427 compileAsync a schema with discriminator and $ref

### LINKED ISSUE #2427: compileAsync does not loadSchema a schema that is $ref’d by a oneOf with a discriminator

latest (8.13.0)

**Ajv options object**

```javascript
const options = {
  discriminator: true,
  async loadSchema(url) {
    console.log('loadSchema', url)
    const name = url.substring("scheme://".length)
    return schemas[name]
  },
};
```

**JSON Schema**

```js
const schemas = {
  main: {
    type: "object",
    discriminator: {propertyName: "foo"},
    required: ["foo"],
    oneOf: [
      {
        $ref: "scheme://schema1",
      },
      {
        $ref: "scheme://schema2",
      },
    ],
  },
  schema1: {
    type: "object",
    properties: {
      foo: {const: "x"},
    },
  },
  schema2: {
    type: "object",
    properties: {
      foo: {enum: ["y", "z"]},
    },
  },
};
```

**Sample data**

```json
{"foo": "x"}
```

**Your code**

```javascript
const Ajv = require("ajv");
const ajv = new Ajv(options);
// If you uncomment this, then there is no error
// Object.entries(schemas).forEach(([id, schema]) => {
//   ajv.addSchema(schema, 'scheme://' + id)
// })
async function main() {
  const validate = await ajv.compileAsync(schemas.main)
  validate(data)
}
main().catch((err) => console.log(err))
```

```
Error: discriminator: oneOf subschemas (or referenced schemas) must have "properties/foo"
```

No error

Maybe if it is easy

Notes:
* If you call `ajv.addSchema` (i.e., uncomment the commented code), then the error goes away. Therefore, it is the resolving of the `$ref` that is the issue
* When `discriminator: true`, then `loadSchema` is **not** called.
* But if you set `discriminator: false` and remove the `discriminator` from `schema.main`, then `loadSchema` **is** called.

  ```
  loadSchema scheme://schema1
  loadSchema scheme://schema2
  ```

My guess is that lib/vocabularies/discriminator/index.ts’s `getMapping()` needs to `throw new MissingRefError` so that _compileAsync can catch the error and load the missing schemas.

yonran:
Make the discriminator code generation throw MissingRefError when the $ref does not synchronously resolve so that compileAsync can loadSchema and retry.

#2427 

Make the discriminator code generation throw MissingRefError when the $ref does not synchronously resolve so that compileAsync can loadSchema and retry.

The existing tests create separate test `Ajv` instances by calling `withStandalone(getAjvInstances(AjvClass, …))` (https://github.com/ajv-validator/ajv/blob/v8.13.0/spec/discriminator.spec.ts#L22). I could not directly reuse those `Ajv` instances because I need to add the `loadSchema` option and I can’t call compileAsync on an `Ajv` after `withStandalone`. For now I only called `new _Ajv`. To what extent should I try creating separate ajv, ajv2019, standalone instances in the test?

### DIFF
--- a/lib/vocabularies/discriminator/index.ts
+++ b/lib/vocabularies/discriminator/index.ts
@@ -3,6 +3,7 @@ import type {KeywordCxt} from "../../compile/validate"
 import {_, getProperty, Name} from "../../compile/codegen"
 import {DiscrError, DiscrErrorObj} from "../discriminator/types"
 import {resolveRef, SchemaEnv} from "../../compile"
+import MissingRefError from "../../compile/ref_error"
 import {schemaHasRulesButRef} from "../../compile/util"
 
 export type DiscriminatorError = DiscrErrorObj<DiscrError.Tag> | DiscrErrorObj<DiscrError.Mapping>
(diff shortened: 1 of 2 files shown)


jasoniangreen:
I've made some small changes just to get it passing linting and see how the tests run.

jasoniangreen:
Not sure why coveralls is erroring. Will try a clean branch and see if I get the same issue.

jasoniangreen:
This has been moved to #2433 due to coveralls issues.

PR #2473, opened 2024-07-10
perf: replace fast-deep-equal with dequal

### LINKED ISSUE #2466: Replace fast-deep-equal with dequal
8.16.0

`dequal` is faster than `fast-deep-equal`: [benchmarks](https://github.com/lukeed/dequal?tab=readme-ov-file#benchmarks)

```diff
-import deepEqual from 'fast-deep-equal'
+import { dequal } from 'dequal/lite'
```

No

gurgunday:
lukeed's `dequal` is amazing, both smaller and [faster](https://github.com/lukeed/dequal?tab=readme-ov-file#benchmarks) than `fast-deep-equal` while being compatible

closes #2466

### DIFF
--- a/lib/compile/resolve.ts
+++ b/lib/compile/resolve.ts
@@ -2,7 +2,7 @@ import type {AnySchema, AnySchemaObject, UriResolver} from "../types"
 import type Ajv from "../ajv"
 import type {URIComponent} from "fast-uri"
 import {eachItem} from "./util"
-import { dequal as equal } from "dequal"
+import {dequal as equal} from "dequal"
 import * as traverse from "json-schema-traverse"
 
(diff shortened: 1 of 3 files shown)


gurgunday:
cc @jasoniangreen 🤠

jasoniangreen:
Do you know the bundle size difference between the two? I tried to look it up on [bundlephobio](https://bundlephobia.com/)  but it doesn't show the size of sub paths like `/lite`

gurgunday:
cc @jasoniangreen

This should be an easy swap, much easier than the fast-uri swap 🤣

I believe they are semantically 100% equivalent

gurgunday:
Closing per https://github.com/ajv-validator/ajv/issues/2466#issuecomment-2234217778

PR #2474, opened 2024-07-13
perf: stop `.split` early at `getFullPath`

gurgunday:
`String.prototype.split()` supports a [second parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#limit) that stops the search as soon as the number of elements in the resulting list are hit

In this case, we are always getting the first element after splitting with `#`, therefore it doesn't make sense to continue searching the string for more separators and end up with a larger-than-necessary array


gurgunday:
@epoberezkin can you take a look at this?

gurgunday:
@jasoniangreen can you take a look?

Issue #2477, opened 2024-07-19
`regexp` should be wrapped in try/catchs


RedHatter:
**What version of Ajv you are you using?**
8.17.1

**What problem do you want to solve?**
Tracking down invalid regex.

**What do you think is the correct solution to problem?**
An invalid regex in the schema (in my case `^[0-9]{2-4}`) throws an error without any indication where in the schema or what pattern failed. When you have a large schema this makes it difficult to track down exactly where the bad pattern is. There's a [todo in the code](https://github.com/ajv-validator/ajv/blob/9050ba1359fb87cd7c143f3c79513ea7624ea443/lib/vocabularies/validation/pattern.ts#L21) about this.

This can be somewhat mitigated by use the `regExp` option.

```
new Ajv({
  code: {
    regExp: (pattern: string, u: string) => {
      try {
        return new RegExp(pattern, u)
      } catch (e) {
        console.error('Bad RegExp: ', pattern, e)
      }
    },
  }
})
```

However this only displays the bad pattern not where in the schema it's used.

**Will you be able to implement it?**
No.

jasoniangreen:
Hi @RedHatter, you are right, it's not very helpful in this situation. Thanks for sharing the mitigating approach. I will try and raise it with the author sometime but there are other priorities at the moment, though I am curious why it wasn't just wrapped at the time. There might be some good reason why not.

I'll keep an eye on this issue too to see how requested a change in this area is.

jasoniangreen:
I'll try and find time to fit in a fix on this but have other priorities so I'm also happy to review a PR by someone else.

jasoniangreen:
Sorry @RedHatter but I want to check something. You said that the existing error doesn't show you the failing `pattern` but I get an error message like this:
```
SyntaxError: Invalid regular expression: /^[0-9]{2-4}/: Incomplete quantifier
```

RedHatter:
The exact error message I got was
```
Uncaught SyntaxError: incomplete quantifier in regular expression
```
Note that this was in firefox. I didn't check the error message in any other browsers at the time.

jasoniangreen:
Ah I see. It seems in node (at least version 18) you get a more useful message that includes the expression. I'll still keep it in mind because we could include the json schema path of the expression.

dragonfleas:
> **What version of Ajv you are you using?** 8.17.1
> 
> **What problem do you want to solve?** Tracking down invalid regex.
> 
[...]

PR #2485, opened 2024-08-04
feat: add pattern and path to regexp create errors #2477

jasoniangreen:

#2477 - Wrap regexp instantiation in try/catch to add pattern and schema path to the error message

Wrapped both the regular and codegen'd versions of the pattern keyword regexp creation

Not sure if there is a better way to do the codegen, I did try using gen.try etc but couldn't figure it out.

### DIFF
--- a/lib/vocabularies/validation/pattern.ts
+++ b/lib/vocabularies/validation/pattern.ts
@@ -24,7 +24,7 @@ const def: CodeKeywordDefinition = {
         try {
           return new RegExp(${schemaCode}, ${u})
         } catch (e) {
-          throw new Error('Invalid regular expression: ' + ${schemaCode} + ' at ' + ${it.errSchemaPath})
+          throw new Error(e.message + ' | pattern ' + ${schemaCode} + ' at ' + ${it.errSchemaPath})
         }
       })()`
       : usePattern(cxt, schema)
--- a/lib/vocabularies/validation/pattern.ts
+++ b/lib/vocabularies/validation/pattern.ts
@@ -18,9 +18,16 @@ const def: CodeKeywordDefinition = {
   error,
   code(cxt: KeywordCxt) {
     const {data, $data, schema, schemaCode, it} = cxt
-    // TODO regexp should be wrapped in try/catchs
     const u = it.opts.unicodeRegExp ? "u" : ""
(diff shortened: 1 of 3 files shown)


lib/vocabularies/validation/pattern.ts
```diff
@@ -18,9 +18,16 @@ const def: CodeKeywordDefinition = {
   error,
   code(cxt: KeywordCxt) {
     const {data, $data, schema, schemaCode, it} = cxt
-    // TODO regexp should be wrapped in try/catchs
     const u = it.opts.unicodeRegExp ? "u" : ""
-    const regExp = $data ? _`(new RegExp(${schemaCode}, ${u}))` : usePattern(cxt, schema)
+    const regExp = $data
+      ? _`(function() {
```
jasoniangreen, line 23: There is probably a way to build this with `gen.try` etc but maybe this is ok? I'm still getting my head around codegen stuff.

lib/vocabularies/validation/pattern.ts
```diff
@@ -18,9 +18,16 @@ const def: CodeKeywordDefinition = {
   error,
   code(cxt: KeywordCxt) {
     const {data, $data, schema, schemaCode, it} = cxt
-    // TODO regexp should be wrapped in try/catchs
     const u = it.opts.unicodeRegExp ? "u" : ""
-    const regExp = $data ? _`(new RegExp(${schemaCode}, ${u}))` : usePattern(cxt, schema)
```
jasoniangreen, line 23: This was never using `opts.code.regExp`? Is that intentional?

lib/vocabularies/validation/pattern.ts
```diff
@@ -18,9 +18,16 @@ const def: CodeKeywordDefinition = {
...
+    const regExp = $data
+      ? _`(function() {
+        try {
+          return new RegExp(${schemaCode}, ${u})
+        } catch (e) {
+          throw new Error(e.message + ' | pattern ' + ${schemaCode} + ' at ' + ${it.errSchemaPath})
+        }
+      })()`
+      : usePattern(cxt, schema)
```
jasoniangreen, line 30: There is probably a way to build this with gen.try etc but maybe this is ok? I'm still getting my head around codegen stuff.

Issue #2482, opened 2024-08-01
serializing Infinity creates invalid JSON


MatthiasMann:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**
8.17.1

**Ajv options object**
{}

Test case:
```javascript
interface LoudnessMeasurement {
  m: number | null,
  s: number | null,
  g: number | null,
  g_seconds: number
}
const schema: JTDSchemaType<LoudnessMeasurement> = {
  properties: {
    m: { type: "float64", nullable: true },
    s: { type: "float64", nullable: true },
    g: { type: "float64", nullable: true },
    g_seconds: { type: "float64" }
  }
};
const ajv = new Ajv();
let serializer = ajv.compileSerializer(schema);
console.log(serializer({m:-Infinity, s:-12.0, g:null, g_seconds: 420.69}));
```
Result:
```{"m":-Infinity,"s":-12,"g":null,"g_seconds":420.69}```

This JSON is invalid and should instead be:
```{"m":"-Infinity","s":-12,"g":null,"g_seconds":420.69}```

**Comment by MatthiasMann** [comment]
Setting the options object to ```{strictNumbers: false}``` did not help.

**Comment by jasoniangreen** [comment]
Hi @MatthiasMann thanks for raising this. It does indeed seem wrong to me that `compileSerializer` is able to produce a string that will throw when passed to `JSON.parse` but according to the JSON spec your proposal is also incorrect. In theory `Infinity` (and also `NaN`) should be be converted to `null` when serialised. You can try it yourself:

```
> JSON.stringify({"blah": Infinity})
> '{"blah":null}'
```

So the correct fix (though I would need to consult on this more to be sure) would be to convert to `null`. If you wanted `"-Infinity"` in your case, then that is something you would need to manage separately, for instance, by traversing your objects and swapping `Infinity` for `"Infinity"`.

MatthiasMann:
Hmm you are right - this is very annoying.

As I only have very few elements which need to handle -Infinity I just replaced these with my own function and template literals for this object:
```javascript
function peak_to_JSON(n: number): string {
  return (n > -190) ? n.toString() : "\"-Infinity\"";
}
function peak_to_JSON_nullable(n: number | null): string {
  return (n === null) ? "null" : peak_to_JSON(n);
}```

let serializer = (data:LoudnessMeasurement) => `{\"m\":${peak_to_JSON_nullable(data.momentary)},\
[...]

PR #2487, opened 2024-08-11
fix: #2482 Infinity and NaN serialise to null

jasoniangreen:
#2482 

Added an option to support serializing the special case numbers`Infinity`, `-Infinity` and `NaN`. See added docs for expected behavior but basically these options help to avoid invalid JSON when dealing with these values.

The tests and the new docs explain it all.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -204,6 +204,14 @@ Defines how date-time strings are parsed and validated. By default Ajv only allo
 This option makes JTD validation and parsing more permissive and non-standard. The date strings without time part will be accepted by Ajv, but will be rejected by other JTD validators.
 :::
 
+### safeNumbers <Badge text="JTD only" />
+
+Defines how special case numbers, Infinity, -Infinity and NaN are handled. Use `safeNumbers: "null"` to serialize them to `null` which is correct behavior according to the JSON spec. If you wish to handle these values, however, and not lose the data you can use `safeNumbers: "string"` which will serialize them to strings. If this option is not set the values will be included as the original literal values.
+
+::: warning The default behavior can produce invalid JSON
+If `safeNumbers` is left undefined, the serializer will produce invalid JSON when there are any special case numbers in the data. This is, however, the fastest mode and so should be used unless you expect to encounter special case numbers.
+:::
+
 ### int32range <Badge text="JTD only" />
 
 Can be used to disable range checking for `int32` and `uint32` types.
--- a/docs/options.md
+++ b/docs/options.md
@@ -206,10 +206,16 @@ This option makes JTD validation and parsing more permissive and non-standard. T
 
 ### specialNumbers <Badge text="JTD only" />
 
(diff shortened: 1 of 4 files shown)


lib/compile/jtd/serialize.ts
```diff
@@ -228,8 +228,17 @@ function serializeString({gen, data}: SerializeCxt): void {
...
+  const condition = _`${data} === Infinity || ${data} === -Infinity || Number.isNaN(${data})`
+  const addNumber = (): CodeGen => gen.add(N.json, _`"" + ${data}`)
+
+  if (self.opts.specialNumbers === "null") {
+    gen.if(condition, () => gen.add(N.json, _`null`), addNumber)
+  } else if (self.opts.specialNumbers === "string") {
+    gen.if(condition, () => gen.add(N.json, str`"${data}"`), addNumber)
+  } else {
+    addNumber()
+  }
```
epoberezkin, line 241:
```suggestion
  if (self.opts.specialNumbers === "null") {
    gen.if(
      _`${data} === Infinity || ${data} === -Infinity || Number.isNaN(${data})`,
      () => gen.add(N.json, _`null`),
      addNumber
    )
  } else {
    addNumber()
  }
  
  function addNumber() CodeGen {
    gen.add(N.json, _`"" + ${data}`)
  }
```

docs/options.md
```diff
@@ -204,6 +204,14 @@ Defines how date-time strings are parsed and validated. By default Ajv only allo
 This option makes JTD validation and parsing more permissive and non-standard. The date strings without time part will be accepted by Ajv, but will be rejected by other JTD validators.
 :::
 
+### specialNumbers <Badge text="JTD only" />
```
epoberezkin, line 207: add default to top and correct to reflect that "string" is default and is faster than "null"

lib/compile/jtd/serialize.ts
epoberezkin: x !== x avoids function call overhead, the only advantage of Number.isNaN is clarity, but we're optimizing for performance here.

epoberezkin, line 241:
```suggestion
  if (self.opts.specialNumbers === "null") {
    gen.if(condition, () => gen.add(N.json, _`null`), addNumber)
  } else if (self.opts.specialNumbers === "string") {
    gen.if(condition, () => gen.add(N.json, str`"${data}"`), addNumber)
  } else {
    addNumber()
  }
  
  function addNumber(): CodeGen {
    gen.add(N.json, _`"" + ${data}`)
  }
```

epoberezkin:
```suggestion
  if (self.opts.specialNumbers === undefined || self.opts.specialNumbers === "fast") {
    gen.add(N.json, _`"" + ${data}`)
  } else {
    gen.if(
      _`${data} === Infinity || ${data} === -Infinity || ${data} !== ${data}`,
      () => gen.add(N.json, _`null`),
      () => gen.add(N.json, _`"" + ${data}`)
    )
  }
```

jasoniangreen: We discussed in person, we will still have the 3 options but I got rid of the wrapping function `addNumber` and added in the default handling.

jasoniangreen: Done

docs/options.md
jasoniangreen: Done

PR #2575, opened 2026-01-17
implement parallel schema loading

Kosta-Github:

speed up the loading process of external `$refs` if a schema contains a lot of them

implement parallel schema loading

* introduce a new flag `enableParallelLoading`
* defaults to `false` to keep existing sequential behavior
* when enabled the full schema gets scanned first for all external `$refs` and all missing `$refs` are loaded in parallel with `Promise.allSettled()`
* this can speed up the loading process considerably for schemas that have a lot of external `$refs`
* the existing and the new behavior are both covered by corresponding unit tests
* documentation has been updated in various places as well

No


Kosta-Github:
The build is failing within the `npm install` step. But I haven't changed any of the dependencies in this PR and therefore the issue should be unrelated to this PR.

PR #2582, opened 2026-02-13
Fix: ignore undefined values in schema properties

### LINKED ISSUE #2578: undefined values in schemas are not ignored but throws TypeError
I was struggling for hours with an error only happening in my deployed function, but not locally.
Turns out, since I generate the schema programmatically, I had some optional property that was undefined in the schema. This causes AJV to throw `TypeError: Cannot convert undefined or null to object`.

I logged my schema and data, copy pasted it in a local test program and spent HOURS not understanding what was happening. Why was it working locally but not when running in my deployed function?
Turns out the schema had an `undefined` in it, which was of course trimmed away silently when logging to cloud watch, hence not ending up in my local test program.

is there any reason Ajv could not ignore undefineds? I assume it does something like iterate properties using `Object.keys` or `for const k in obj`, and just checking `if obj[k] === undefined {<skip this>}` seems harmless?

ajv 8.17.1 - latest

**Ajv options object**

```javascript
{strictSchema: false}

```

**JSON Schema**

```json
{
  type: "object", 
  properties: {
    foo: {type: "string"},
    bar: undefined
  }
}

```

**Sample data**

```json
{foo:  "FOO"}

```

**Your code**

```javascript
new Ajv({strictSchema: false}).validate(schema, data);
```

```
TypeError: Cannot convert undefined or null to object
```

I expected `undefined` values in schemas to be ignored silently and the object to match the schema.

I'm happy to attempt at a PR

veeceey:
## Summary

This PR fixes issue #2578 where schemas with undefined property values (commonly from destructuring) cause ajv to throw a TypeError instead of ignoring them.

**Before:**
```js
const ajv = new Ajv({strictSchema: false, validateSchema: false});
const schema = {
  type: 'object',
  properties: {
    foo: {type: 'string'},
    bar: undefined  // From destructuring or conditional logic
  }
};
ajv.validate(schema, {foo: 'test'}); // Throws: "Cannot convert undefined or null to object"
```

**After:**
```js
ajv.validate(schema, {foo: 'test'}); // Returns: true (undefined properties ignored)
```

## Changes

- Modified `allSchemaProperties()` in `lib/vocabularies/code.ts` to filter out undefined values
- Added comprehensive test coverage for various scenarios with undefined schema properties

## Test Coverage

The fix includes tests for:
- Basic undefined property handling
- Multiple undefined properties
- Validation still works correctly for defined properties
- Undefined values from destructuring scenarios

## Manual Testing

Verified the fix handles:
- Single undefined properties
- Multiple undefined properties mixed with valid ones
- Validation errors still work correctly for defined properties
- Pattern properties with undefined values

Closes #2578

### DIFF
--- a/lib/vocabularies/code.ts
+++ b/lib/vocabularies/code.ts
@@ -63,7 +63,9 @@ export function noPropertyInData(
 }
 
 export function allSchemaProperties(schemaMap?: SchemaMap): string[] {
-  return schemaMap ? Object.keys(schemaMap).filter((p) => p !== "__proto__" && schemaMap[p] !== undefined) : []
+  return schemaMap
+    ? Object.keys(schemaMap).filter((p) => p !== "__proto__" && schemaMap[p] !== undefined)
+    : []
 }
 
 export function schemaProperties(it: SchemaCxt, schemaMap: SchemaMap): string[] {
--- a/lib/vocabularies/code.ts
+++ b/lib/vocabularies/code.ts
@@ -63,7 +63,7 @@ export function noPropertyInData(
 }
 
 export function allSchemaProperties(schemaMap?: SchemaMap): string[] {
-  return schemaMap ? Object.keys(schemaMap).filter((p) => p !== "__proto__") : []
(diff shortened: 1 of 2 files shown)


veeceey:
gentle ping — would be great to get some eyes on this when possible

PR #2583, opened 2026-02-13
fix: reject date-time/time strings with trailing characters (#2572)

### LINKED ISSUE #2572: Issue: Invalid `date-time` and `time` Formats Accepted
I am encountering issues with the JSON Schema validation for both the `date-time` and `time` formats. The validator is incorrectly accepting invalid strings that should be rejected according to the ISO 8601 standard.

### 1. Invalid `date-time` Format:

The following date-time string is being accepted as valid, but it should be considered invalid:

- Valid date-time: `2024-06-01T12:34:56Z`
- Invalid date-time (but accepted): `2024-06-01T12:34:56ZAS`

The `ZAS` suffix is not part of the correct ISO 8601 format, yet the validator allows it.

### 2. Invalid `time` Format (with Offset):

The following time string with an offset is accepted as valid, which is correct:

- Valid time with offset: `12:34:56+12:44`

However, the following string is also accepted, which should **not** be valid:

- Invalid time with extraneous characters: `12:34:56+12:44AS`

The `AS` suffix is not part of the valid ISO 8601 time format with offset, but the validator incorrectly allows it.

### Expected Behavior:
1. For the `date-time` format, only valid ISO 8601 date-time strings (e.g., `2024-06-01T12:34:56Z`) should be accepted, and any additional suffixes (like `ZAS`) should be rejected.
2. For the `time` format with offset, only valid time strings with an offset (e.g., `12:34:56+12:44`) should be accepted. Any extra characters (like `AS` in `12:34:56+12:44AS`) should cause the string to fail validation.

### Steps to Reproduce:
1. Use a JSON Schema validator that supports the `date-time` and `time` formats (e.g., AJV, Joi, etc.).
2. Provide the following strings as input:
   - `2024-06-01T12:34:56ZAS` for `date-time`.
   - `12:34:56+12:44AS` for `time` with offset.
3. Both strings should be rejected, but the validator accepts them as valid.

### Version of the Validator:
- Version: 1.4.0

### Additional Notes:
- The issue arises when extra characters are appended to valid time strings, which should be rejected.
- This behavior could be problematic when ensuring strict ISO 8601 conformance for both `date-time` and `time` fields with offsets.

Thanks for investigating this!

veeceey:
Fixes #2572

The `validTimestamp` runtime in `lib/runtime/timestamp.ts` already correctly rejects date-time and time strings that have trailing characters after the timezone (e.g. `2024-06-01T12:34:56ZAS`, `12:34:56+12:44AS`). The `TIME` regex's `$` anchor ensures no extra characters are accepted.

This PR adds regression tests that explicitly cover the cases from the issue:

- Trailing characters after `Z` timezone indicator (`2024-06-01T12:34:56ZAS`)
- Trailing characters after offset (`2024-06-01T12:34:56+12:44AS`)  
- Leading whitespace or characters
- End-to-end JTD timestamp validation for these cases

Tests verified locally:

```
  Invalid date-time and time with trailing characters (issue #2572)
    validTimestamp runtime
      ✔ should reject date-time strings with trailing characters after Z
      ✔ should reject date-time strings with trailing characters after offset
      ✔ should reject date-time strings with leading whitespace or characters
      ✔ should accept valid date-time strings
      ✔ should accept valid date strings when allowDate is true
      ✔ should reject invalid date strings when allowDate is true
    JTD timestamp validation end-to-end
      ✔ should reject timestamps with trailing characters after Z
      ✔ should reject timestamps with trailing characters after offset

  8 passing (2ms)
```

### DIFF
--- a/spec/issues/2572_invalid_datetime_trailing_chars.spec.ts
+++ b/spec/issues/2572_invalid_datetime_trailing_chars.spec.ts
@@ -1,0 +1,67 @@
+import _AjvJTD from "../ajv_jtd"
+import validTimestamp from "../../dist/runtime/timestamp"
+
+describe("Invalid date-time and time with trailing characters (issue #2572)", () => {
+  describe("validTimestamp runtime", () => {
+    it("should reject date-time strings with trailing characters after Z", () => {
+      if (validTimestamp("2024-06-01T12:34:56Z", false) !== true) throw new Error("fail")
+      if (validTimestamp("2024-06-01T12:34:56ZAS", false) !== false) throw new Error("fail")
+      if (validTimestamp("2024-06-01T12:34:56Zx", false) !== false) throw new Error("fail")
+      if (validTimestamp("2024-06-01T12:34:56Z ", false) !== false) throw new Error("fail")
+      if (validTimestamp("2024-06-01T12:34:56Z\t", false) !== false) throw new Error("fail")
+      if (validTimestamp("2024-06-01T12:34:56Z\n", false) !== false) throw new Error("fail")
+    })
+
+    it("should reject date-time strings with trailing characters after offset", () => {
+      if (validTimestamp("2024-06-01T12:34:56+12:44", false) !== true) throw new Error("fail")
+      if (validTimestamp("2024-06-01T12:34:56+12:44AS", false) !== false) throw new Error("fail")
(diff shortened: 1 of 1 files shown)


veeceey:
hey, wondering if this could use a review — happy to rework anything

PR #2588, opened 2026-02-18
fix(security): backport CVE-2025-69873 - wrap $data pattern in try/catch

hdsenevi:
- CVE-2025-69873 (ReDoS in the pattern keyword when $data is enabled).
- With $data: true, the pattern value can come from validated data and is passed to new RegExp() without safeguards. That allows invalid regex syntax to throw and malicious patterns (e.g. ^(a|a)*$) to cause catastrophic backtracking (ReDoS).
- This PR backports the same class of fix already applied on the main branch (v8) in #2586 (commit 720a23fa).

- lib/dot/pattern.jst
   - When the pattern value comes from $data, the template now generates a try/catch around RegExp creation and .test().
   - If new RegExp(...) or .test() throws (e.g. invalid regex from user input), the catch sets the result to “invalid” so validation fails instead of throwing.
   - The non-$data path is unchanged (still uses it.usePattern($schema)).
- spec/issues/cve_2025_69873_redos_attack.spec.js
   - New spec (aligned with the approach in 720a23fa453ffae8340e92c9b0fe886c54cfe0d5) that:
      - Ensures the default engine handles the ReDoS-style payload without throwing and returns false (with a safe timeout).
      - Ensures invalid regex syntax in the $data pattern fails validation instead of throwing.
      - Ensures valid $data patterns still validate correctly.
      - Ensures the attack payload is processed without throwing and within a bounded time.
- After editing pattern.jst, npm run build was run so the compiled lib/dotjs/pattern.js is updated (if your workflow commits built output).

- Scope of the fix: The change only adds a try/catch on the $data path. It prevents throws from invalid or problematic patterns and makes validation fail safely. It does not remove ReDoS risk: a malicious pattern can still cause long runtimes via catastrophic backtracking; it just won’t throw. v8 also documents/supports a safe regex engine (e.g. RE2) for full ReDoS mitigation; v6 has no such option, so this backport is intentionally limited to the try/catch behavior.
- Tests: The new spec uses small repeat counts and generous time limits (e.g. &lt; 8–10s) so the suite doesn’t hang on slow CI while still asserting that the try/catch path is used and no throw occurs. Reviewers may want to run the new spec locally (`npx mocha spec/issues/cve_2025_69873_redos_attack.spec.js`) to confirm.

### DIFF
--- a/lib/dot/pattern.jst
+++ b/lib/dot/pattern.jst
@@ -3,12 +3,24 @@
 {{# def.setupKeyword }}
 {{# def.$data }}
 
+{{? $isData }}
 {{
-  var $regexp = $isData
-                ? '(new RegExp(' + $schemaValue + '))'
-                : it.usePattern($schema);
+  var $regexp = '(new RegExp(' + $schemaValue + '))';
 }}
-
+var patternValid;
+try {
+  patternValid = {{=$regexp}}.test({{=$data}});
+} catch (e) {
+  patternValid = false;
+}
+if ({{# def.$dataNotType:'string' }} !patternValid ) {
+  {{# def.error:'pattern' }}
+} {{? $breakOnError }} else { {{?}}
+{{??}}
+{{
+  var $regexp = it.usePattern($schema);
+}}
 if ({{# def.$dataNotType:'string' }} !{{=$regexp}}.test({{=$data}}) ) {
   {{# def.error:'pattern' }}
 } {{? $breakOnError }} else { {{?}}
+{{?}}
--- a/spec/issues/cve_2025_69873_redos_attack.spec.js
+++ b/spec/issues/cve_2025_69873_redos_attack.spec.js
@@ -1,0 +1,107 @@
+'use strict';
+
+var Ajv = require('../ajv');
+require('../chai').should();
+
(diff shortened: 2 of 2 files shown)


Chenky100 (approved):

subhadipmikmak (approved):

wayne530:
Is there an estimated timeline to get this merged and a new 6.x release cut or anything we can do to help get this moving along?

epoberezkin:
sorry, missed it, fix merged now. Thank you!

epoberezkin:
it's in [6.14.0](https://www.npmjs.com/package/ajv/v/6.14.0)

wayne530:
Thanks @epoberezkin !

wayne530:
Sorry, one last question @epoberezkin - the advisories don't know the fix has been backported to v6 so it still treats 6.14.0 as vulnerable since it is < 8.18.0. Do you happen to know how to update this?

mitar:
@wayne530 A pull request has to be made against: https://github.com/github/advisory-database

wayne530:
https://github.com/github/advisory-database/pull/6991

PR #2590, opened 2026-02-18
add regExp option to address $data exploit via a regular expression (CVE-2025-69873)

epoberezkin:
**What issue does this pull request resolve?**

**What changes did you make?**

**Is there anything that requires more attention while reviewing?**

PR #2591, opened 2026-02-18
fix(security): mitigate ReDoS in compile-dots.js by avoiding regex match on tainted code

hdsenevi:
- This PR fixes a Regular Expression Denial of Service (ReDoS) (CWE-400) reported by Snyk in scripts/compile-dots.js. The script compiles doT templates and its input (paths and template output) can be influenced by command-line arguments. That input was passed into code.match(regexp) and into a regex using a greedy [^;]+, which could lead to excessive backtracking on crafted or large input and cause the script to hang. The PR removes that vulnerable pattern and makes the remaining regex safe so the ReDoS finding is resolved.

- Replace regex-based counting with indexOf/split so tainted template output is never passed to .match(), addressing the ReDoS finding.
- Use lazy quantifier [^;]*? in the remove-unused-var regex to avoid catastrophic backtracking.
- Derive ERROR_KEYWORD_STR from ERROR_KEYWORD and use it in removeAlwaysFalsyInOr instead of the '$errorKeyword' literal.
- Remove the occurrences() helper; counting is now done inline without regex on user-influenced code.

- If the project runs Snyk (or similar) in CI, re-running the scan on this branch should show the ReDoS finding for scripts/compile-dots.js resolved.
- When paired with PR https://github.com/ajv-validator/ajv/pull/2588 this will resolve the ReDoS security vulnerability in v6

### DIFF
--- a/scripts/compile-dots.js
+++ b/scripts/compile-dots.js
@@ -28,6 +28,7 @@ var FUNCTION_NAME = /function\s+anonymous\s*\(it[^)]*\)\s*{/;
 var OUT_EMPTY_STRING = /out\s*\+=\s*'\s*';/g;
 var ISTANBUL = /'(istanbul[^']+)';/g;
 var ERROR_KEYWORD = /\$errorKeyword/g;
+var ERROR_KEYWORD_STR = ERROR_KEYWORD.source.replace(/^\\\\/, '');
 var ERROR_KEYWORD_OR = /\$errorKeyword\s+\|\|/g;
 var VARS = [
   '$errs', '$valid', '$lvl', '$data', '$dataLvl',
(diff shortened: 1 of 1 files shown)


paroczigergo:
We are really waiting for this to be released! Thank you @hdsenevi for the quick fix!

hdsenevi:
Hi @epoberezkin, I see PR #2588 was closed in favor of #2590. The ReDoS issue still appears to be present in [ajv@6.14.0](https://www.npmjs.com/package/ajv/v/6.14.0). 

This PR proposes a fix for that.

## Verification
I verified the vulnerability and the fix locally with the [Snyk CLI](https://docs.snyk.io/developer-tools/snyk-cli) locally:

Before: 
[before-snyk-vuls-high.log](https://github.com/user-attachments/files/25464857/before-snyk-vuls-high.log)
```
 ✗ [HIGH] Regular Expression Denial of Service (ReDoS)
   Path: scripts/compile-dots.js, line 71
   Info: Unsanitized user input from a command line argument flows into match, where it is used to build a regular expression. This may result in a Regular expression Denial of Service attack (reDOS).
```

After: 
[after-snyk-vuls-high.log](https://github.com/user-attachments/files/25464858/after-snyk-vuls-high.log)

Happy to adjust the approach or wording if you’d prefer a different direction for the v6 branch.

Issue #2595, opened 2026-03-09
JTD parser fails on empty array that has whitespace between brackets


nathan-duddles:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**
8.18.0 (latest at time of submission)

**Ajv options object**

```javascript
import Ajv, { JTDDataType } from "ajv/dist/jtd";
const ajv = new Ajv();
```

**JSON Type Definition**

```json
{"properties": { "data": { "elements": { "type": "string"} } }
```

**Sample data**

```json
{"data": [ ]}
```

**Your code**

```javascript
import Ajv, { JTDDataType } from "ajv/dist/jtd";

import schema from './schema.json'; // Contains JTD above

const ajv = new Ajv();

export type SchemaType = JTDDataType<type of schema>;
export const dataParser = ajv.compileParser<SchemaType>(schema);

const text = `{"data": [ ]}`;
const parsedData = dataParser(text);

console.log(parsedData);
console.log(dataParser.message);
console.log(dataParser.position);
```

**Validation result, data AFTER validation, error messages**

```
undefined
unexpected token ]
10
```

**What results did you expect?**
Parser should not fail on an empty array that contains white spaces.

**Are you going to resolve the issue?**
I will open a PR. It is a one line fix - we just need to add `skipWhitespace(cxt)` to `lib/compile/jtd/parse.ts` in `parseElements` method before calling `parseItems`.

PR #2596, opened 2026-03-09
Fix: skip whitespace in empty array brackets

nathan-duddles:
Fixes issue #2595 

Skip whitespace before parsing items within array, because the function that parses array, `parseItems`, assumes that if there is whitespace then there exists items to parse.

This fix might also be accomplished by adding the skip whitespaces call to `tryParseItems`, but `tryParseItems` is also used by `parseValues` function in the same file, and it is not clear if `parseValues` would require the same change (while testing parsing an empty `{ }` containing white space, the `parseJson` was used instead of `parseValues`, so I am not sure if this same bug appears in that path... I don't understand where `parseValues` is actually used).


epoberezkin:
Thank you! Please add the test failing without this change.

PR #2597, opened 2026-03-10
fix: recursively check for validity of discriminator

AmadeusK525:
The old implementation was only checking for a single level of validity of a schema using the discriminator extension, without accounting for "unions of unions". So, for example:

```json
{
  "discriminator": {
    "propertyName": "field"
  }
  "required": ["field"],
  "oneOf": [
    {
      "oneOf": [
        {
          "properties": {
            "field": { "const": "a" }
          },
          "required": ["field"]
        },
        {
          "properties": {
            "field": { "const": "b" }
          },
          "required": ["field"]
        }
      ],
      "required": ["field"]
    },
    {
      "properties": {
        "field": { "const": "c" }
      },
      "required": ["field"]
    }
  ]
}
```

This schema would fail because the compiler would complain about the first schema in `oneOf` not having the `field` property, which was defined as the `discriminator`. But it *does* have that property, just in a subschema inside another `oneOf` definition!

### DIFF
--- a/lib/vocabularies/discriminator/index.ts
+++ b/lib/vocabularies/discriminator/index.ts
@@ -63,29 +63,86 @@ const def: CodeKeywordDefinition = {
     function getMapping(): {[T in string]?: number} {
       const oneOfMapping: {[T in string]?: number} = {}
       const topRequired = hasRequired(parentSchema)
-      let tagRequired = true
-      for (let i = 0; i < oneOf.length; i++) {
-        let sch = oneOf[i]
-        if (sch?.$ref && !schemaHasRulesButRef(sch, it.self.RULES)) {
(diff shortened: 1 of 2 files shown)


epoberezkin:
It's debatable it's correct, given how discriminator is specified I think.

AmadeusK525:
> It's debatable it's correct, given how discriminator is specified I think.

@epoberezkin Sorry, but why would this be debatable? The spec may not mention recursion explicitly, but there's no reason to not cover this case, as it's in line with the reasoning behind the discriminator keyword (making sure that every object in the discriminated union has the discriminator property). I really don't see how an implementation that doesn't cover the recursion "edge" case would be more correct than one that doesn't

Issue #2601, opened 2026-04-07
fast-uri - missing export URIComponent


hollandjake:
There are a few instances of this import being defined - `fast-uri` does not have an export for URIComponent as it is a type

https://github.com/ajv-validator/ajv/blob/142ce84b807c4fe66e619c22480a28d0e4bd50fa/lib/types/index.ts#L1

```bash
 1 │ import { URIComponent } from "fast-uri";
   │          ──────┬─────  
   │                ╰─────── Missing export
   │ 
   │ Note: If you meant to import a type rather than a value, make sure to add the `type` modifier (e.g. `import { type Foo } from 'fast-uri'`).
```

PR #2602, opened 2026-04-07
fix: replace import for fast-uri with type import

hollandjake:
#2601 

Define export as type

N/A

### DIFF
--- a/lib/types/index.ts
+++ b/lib/types/index.ts
@@ -1,4 +1,4 @@
-import {URIComponent} from "fast-uri"
+import type {URIComponent} from "fast-uri"
 import type {CodeGen, Code, Name, ScopeValueSets, ValueScopeName} from "../compile/codegen"
 import type {SchemaEnv, SchemaCxt, SchemaObjCxt} from "../compile"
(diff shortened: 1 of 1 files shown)


epoberezkin:
why is it needed? It works without it, no?

hollandjake:
@epoberezkin some bundlers fail when the types are not defined (I've since forgotten which one i hit with it as I moved off it because of this issue) you can see a snippet in #2601

s-ol:
@epoberezkin tsdown won't let me build with ajv as a transitive (!) dependency because of this.

hollandjake:
@s-ol thats the one! Thank you

hollandjake:
@s-ol I ended up fixing it by mutating the source inside the build. here is the plugin

```typescript
{
    name: "exclude_node_modules_from_sourcemaps",
    transform(code, id) {
        if (!id.includes("/ajv/")) return null;
        return code.replace(
            /import\s*\{\s*URIComponent\s*\}\s*from\s*["']fast-uri["'];?/g,
            'import type {URIComponent} from "fast-uri";',
        );
    },
},
```

PR #2615, opened 2026-05-12
Fix dynamicRef static target resolution

### LINKED ISSUE #1745: Support $dynamicAnchor and $recursiveAnchor in any schema location (not only root)
8.6.2, which should be the latest

**Ajv options object**
```javascript
{
    allErrors: true, 
    validateFormats: false, 
    strictSchema: false, 
    strictTypes: false
}
```

See https://github.com/OAI/OpenAPI-Specification/pull/2489 for a discussion on the options.

**JSON Schema**

The OpenAPI 3.1 schema defined here: https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.json

I'm not proficient enough to make this a minimal reproduction example.

**Sample data**

the OpenApi v3.1 webhook example here: https://github.com/OAI/OpenAPI-Specification/blob/main/examples/v3.1/webhook-example.json

**Your code**

See: https://runkit.com/essential-randomness/612d87df07ae0a00092a3492.

This simply call ajv with the given options, passes the OpenApiv3.1 schema and tries to validate the OpenApi Webhook example:

```
Invalid: 
data/webhooks/newPet/post/requestBody/content/application~1json/schema must NOT have unevaluated properties, 
data/webhooks/newPet/post/requestBody must match "else" schema,
data/webhooks/newPet must match "else" schema, 
data/components/schemas/Pet must NOT have unevaluated properties, 
data/components/schemas/Pet must NOT have unevaluated properties
```

**What results did you expect?** I would expect the validation to pass.

I would like some help understanding whether the issue is in AJV or the OpenAPI specification. I was trying to add support for OpenAPIv3.1 to some JavaScript libraries that use AJV, but unfortunately this makes it impossible. 

There's many libraries that will (hopefully) want to make this switch at some point, but I couldn't find anyone who has already done it or has published an investigation of what's needed to get this working. If anyone is ever going to fix it, understanding what's going on would be the first step :)

nsteve-one:
Fixes #1745.

This is an alternative to #2573 against current `master`. It keeps the change scoped to `$dynamicRef` resolution and does not include workflow changes.

### Summary

- Resolve the `$dynamicRef` static target with the existing ref resolver before applying dynamic-anchor overrides.
- Use the resolved non-root `$dynamicAnchor` validator as the fallback target, instead of falling back to the current/root validator.
- Preserve the existing dynamic-scope “bookend” behavior for annotation/location-only dynamic anchors.
- Add regression coverage for the minimized issue schema and a draft-2020/OpenAPI-style schema using `unevaluatedProperties`.

### Tests

- `npm run build`
- `TS_NODE_PROJECT=spec/tsconfig.json ./node_modules/.bin/mocha -r ts-node/register spec/dynamic-ref.spec.ts -R dot`
- `TS_NODE_PROJECT=spec/tsconfig.json ./node_modules/.bin/mocha -r ts-node/register spec/json-schema.spec.ts --grep dynamicRef -R dot`
- Manual draft-2020 repro check with `./dist/2020`
- `npm run prettier:check`
- `npm run eslint`
- `git diff --check`
- `npm test` (`7614 passing`, `350 pending`)

### DIFF
--- a/lib/compile/index.ts
+++ b/lib/compile/index.ts
@@ -12,7 +12,7 @@ import {CodeGen, _, nil, stringify, Name, Code, ValueScopeName} from "./codegen"
 import ValidationError from "../runtime/validation_error"
 import N from "./names"
 import {LocalRefs, getFullPath, _getFullPath, inlineRef, normalizeId, resolveUrl} from "./resolve"
-import {schemaHasRulesButRef, unescapeFragment} from "./util"
+import {schemaHasResourceDynamicAnchors, schemaHasRulesButRef, unescapeFragment} from "./util"
 import {validateFunctionCode} from "./validate"
 import {URIComponent} from "fast-uri"
 import {JSONType} from "./rules"
@@ -311,7 +311,12 @@ function getJsonPointer(
     }
   }
   let env: SchemaEnv | undefined
-  if (typeof schema != "boolean" && schema.$ref && !schemaHasRulesButRef(schema, this.RULES)) {
+  if (
+    typeof schema != "boolean" &&
+    schema.$ref &&
+    !schemaHasRulesButRef(schema, this.RULES) &&
+    !schemaHasResourceDynamicAnchors(schema, this.opts.schemaId)
+  ) {
     const $ref = resolveUrl(this.opts.uriResolver, baseId, schema.$ref)
     env = resolveSchema.call(this, root, $ref)
   }
--- a/lib/compile/util.ts
+++ b/lib/compile/util.ts
@@ -43,6 +43,35 @@ export function schemaHasRulesButRef(schema: AnySchema, RULES: ValidationRules):
   return false
 }
 
+export function schemaResourceDynamicAnchors(schema: AnySchema, schemaId: "$id" | "id"): string[] {
+  const anchors: {[Ref in string]?: true} = {}
+  findDynamicAnchors(schema, true)
+  return Object.keys(anchors)
+
+  function findDynamicAnchors(sch: unknown, isRoot: boolean): void {
+    if (!sch || typeof sch != "object") return
+    if (Array.isArray(sch)) {
+      for (const item of sch) findDynamicAnchors(item, false)
+      return
+    }
+
+    const schObj = sch as {[Key in string]?: unknown}
+    if (!isRoot && typeof schObj[schemaId] == "string") return
+    const {$dynamicAnchor, $recursiveAnchor} = schObj
+    if (typeof $dynamicAnchor == "string") anchors[$dynamicAnchor] = true
+    if ($recursiveAnchor === true) anchors[""] = true
+
+    for (const key in schObj) findDynamicAnchors(schObj[key], false)
+  }
+}
+
+export function schemaHasResourceDynamicAnchors(
+  schema: AnySchema,
+  schemaId: "$id" | "id"
+): boolean {
+  return schemaResourceDynamicAnchors(schema, schemaId).length > 0
+}
+
 export function schemaRefOrVal(
   {topSchemaRef, schemaPath}: SchemaObjCxt,
   schema: unknown,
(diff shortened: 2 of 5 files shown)


aqeelat:
Thanks for this PR — I tested it against the `$dynamicRef` cases we're tracking in https://github.com/aqeelat/openapi-dynamicref-adoption-tracker.

This PR appears to fix the static-target case from #1745: `$dynamicRef` can now resolve a non-root `$dynamicAnchor` in `$defs`.

However, it does not yet cover the JSON Schema "generics" / type-binding pattern described in https://json-schema.org/blog/posts/dynamicref-and-generics and discussed in OAI/OpenAPI-Specification#3601.

## Minimal reproduction

```js
const Ajv2020 = require("ajv/dist/2020.js");
const ajv = new Ajv2020({ allErrors: true, strict: false, validateFormats: false });

const schema = {
  $id: "https://example.com/test",
  $defs: {
    User: {
      type: "object",
      required: ["id", "email"],
      properties: {
        id: { type: "string" },
        email: { type: "string" },
      },
    },
    PaginatedTemplate: {
      $id: "https://example.com/schemas/PaginatedTemplate",
      $defs: {
        itemType: {
          $dynamicAnchor: "itemType",
          not: {},
        },
      },
      type: "object",
      required: ["items", "total", "page", "pageSize"],
      properties: {
        items: {
          type: "array",
          items: { $dynamicRef: "#itemType" },
        },
        total: { type: "integer", minimum: 0 },
        page: { type: "integer", minimum: 1 },
        pageSize: { type: "integer", minimum: 1 },
      },
    },
    PaginatedUserResponse: {
      $id: "https://example.com/schemas/PaginatedUserResponse",
      $defs: {
        itemType: {
          $dynamicAnchor: "itemType",
          $ref: "https://example.com/test#/$defs/User",
        },
      },
      $ref: "https://example.com/test#/$defs/PaginatedTemplate",
    },
  },
};

ajv.addSchema(schema);

const validate = ajv.compile({
  $ref: "https://example.com/test#/$defs/PaginatedUserResponse",
});

const valid = validate({
  items: [{ id: "u1", email: "user@example.com" }],
  total: 1,
  page: 1,
  pageSize: 10,
});

console.log(valid);
console.log(validate.errors);
```

Expected: `true`

Actual on this PR: `false`, with `must NOT be valid`.

On current `master`, the failure mode is different: it appears to resolve to the root/current schema and reports missing pagination fields on the item (`items`, `total`, `page`, `pageSize`). On this PR, it resolves to the static fallback anchor (`not: {}`), which is an improvement, but it still does not apply the dynamic override from `PaginatedUserResponse.$defs.itemType`.

## Interpretation

My read is that this PR fixes static `$dynamicRef` target resolution, but the generics pattern also needs the dynamic-scope override to be available when evaluating through:

```
PaginatedUserResponse -> $ref: PaginatedTemplate -> $dynamicRef: #itemType
```

In that evaluation path, `$dynamicRef: "#itemType"` should resolve to `PaginatedUserResponse.$defs.itemType`, not to the fallback `PaginatedTemplate.$defs.itemType`.

Hyperjump 2020-12 validates this pattern as expected.

We also have a second fixture where the override is inline in an OpenAPI response schema rather than in a named wrapper schema:

- `fixtures/generic-schema-binding.yaml`
- `fixtures/paginated-response.yaml`

Both are in https://github.com/aqeelat/openapi-dynamicref-adoption-tracker and currently pass Hyperjump but fail AJV.

Happy to help turn these into AJV regression tests if useful.

nsteve-one:
Thanks for testing this against the tracker fixtures. I pushed an update in `5cef71f4` that covers the generic/type-binding behavior too.

What changed:
- `$ref`-only schema resources that declare dynamic anchors are no longer resolved through before they can contribute dynamic scope.
- Before following `$ref`, Ajv now binds dynamic anchors for the active base URI when `dynamicRef` is enabled, so the `PaginatedUserResponse -> PaginatedTemplate -> $dynamicRef: #itemType` path resolves to the wrapper resource's `itemType` anchor instead of the template fallback.
- Added regression coverage for both shapes you called out: the named wrapper schema and an inline referencing schema.

Local verification:
- `npm run build`
- `npx cross-env TS_NODE_PROJECT=spec/tsconfig.json mocha -r ts-node/register spec/dynamic-ref.spec.ts -R spec` => 6 passing
- `npx cross-env TS_NODE_PROJECT=spec/tsconfig.json mocha -r ts-node/register spec/resolve.spec.ts -R spec` => 36 passing
- `npm test` => 7616 passing, 350 pending

PR #2626, opened 2026-05-28
Fix: coercion issue with infinity

rajwinder-dev:
## **What issue does this pull request resolve?**

[found this issue form fastify library](https://github.com/fastify/fastify/issues/6718)

This PR fixes a validation issue where `ajv` incorrectly validates non-finite values (e.g., `"Infinity"`, `"9e600"`) as `integer` types when `coerceTypes` is enabled.

## **What changes did you make?**

I added an `isFinite(+${data})` check to the integer validation logic. This ensures that coerced values must be finite numbers to be considered valid integers.

## **Is there anything that requires more attention while reviewing?**

* Please ensure the `isFinite` implementation aligns with the existing coercion patterns used across the codebase.
* The added test cases cover edge cases for both positive and negative non-finite values.

### DIFF
--- a/lib/compile/validate/dataType.ts
+++ b/lib/compile/validate/dataType.ts
@@ -108,7 +108,7 @@ function coerceData(it: SchemaObjCxt, types: JSONType[], coerceTo: JSONType[]): 
         gen
           .elseIf(
             _`${dataType} === "boolean" || ${data} === null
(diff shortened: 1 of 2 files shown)


rajwinder-dev:
I've updated the PR to include a fix for type: "number" as well,  and  also added relevant test cases to cover these.

PR #2639, opened 2026-06-25
fix(if): collect annotations from "if" without "then"/"else"

spokodev:
## Problem

`if` is an applicator: when an instance validates against the `if` subschema, that subschema's annotations are collected and are visible to `unevaluatedProperties` / `unevaluatedItems`. Ajv returns early when neither `then` nor `else` is present, so those annotations are dropped and a schema that should pass is rejected.

```js
const Ajv = require("ajv/dist/2020")
const ajv = new Ajv({strict: false})

const validate = ajv.compile({
  type: "object",
  if: {properties: {b: {type: "string"}}, required: ["b"]},
  unevaluatedProperties: false,
})

validate({b: "y"})
// false — "must NOT have unevaluated properties", unevaluatedProperty: "b"
// expected: true, "b" is evaluated by the matching "if"
```

Measured on ajv 8.20.0. The same drop happens whenever the applied branch is absent or always-valid, not only when both are missing:

| schema | annotations from `if` collected |
|---|---|
| `if` + `then` with content | yes |
| `if` + `then: {}` | no |
| `if` + `else`, instance matches `if` | no |
| `if` alone | no |

## Fix

`lib/vocabularies/applicator/if.ts` no longer returns before evaluating `if`. The evaluated props/items names are declared first so the merge only assigns when `if` passes, which keeps the assertion semantics unchanged — `if` on its own still never fails validation, it only contributes annotations.

## Test

`spec/issues/if_without_then_else_annotations.spec.ts`, four cases across draft 2019-09 and 2020-12, `unevaluatedItems` and `unevaluatedProperties`. Against `master` they are 0 passing / 4 failing; with this change 4 passing.

Note the spec imports `../dist/2020`, so `npm run build` is required between editing `lib/` and running it — a stale `dist` will show green either way.


spokodev:
Correcting my own mistake: the description on this PR was wrong from the start — I pasted the body of an unrelated PR of mine (a `Buffer#write` fix in another package). It has been replaced with the actual problem, repro and test notes. Sorry to anyone who opened this and found it made no sense.

Also renamed the spec file: it cited issue 2607, which is a different issue (prototype pollution via `$data`). No issue tracks this one, so it now uses the unnumbered form already present in `spec/issues/`.

The code itself is unchanged and was always the `if`-annotations fix the title describes.

PR #2640, opened 2026-06-29
fix(core): null-prototype objects for the schemas and refs registries

mahirhir:

There's no separate issue. This is a direct follow-up to #2607 (and #2606), which changed `formats` to a null-prototype object so a `$data` format name that matches an `Object.prototype` member isn't resolved to an inherited function. The `schemas` and `refs` registries declared just above it still use plain `{}` and have the same problem. Happy to open an issue first if you'd prefer.

`schemas` and `refs` now use `Object.create(null)`, matching `formats`.

Both registries are keyed by schema id/ref, so a key that matches an `Object.prototype` member reads back an inherited value instead of `undefined`. On a fresh instance:

```js
const Ajv = require("ajv").default

new Ajv().getSchema("toString")
// throws: Cannot read properties of undefined (reading 'baseId')
// expected: undefined

new Ajv().addSchema({type: "string"}, "constructor")
// throws: schema with key or id "constructor" already exists
// the key is unused
```

`getSchema` reads `this.schemas[keyRef] || this.refs[keyRef]`, so both have to be null-prototype for it to return `undefined`; `_checkUnique` goes through the same lookup.

Added a regression test in `spec/issues/registry_prototype_keys.spec.ts`.

The keys involved are the inherited `Object.prototype` members (`constructor`, `toString`, `hasOwnProperty`, `valueOf`, `isPrototypeOf`, `propertyIsEnumerable`, `toLocaleString`). I scoped this to `schemas` and `refs` since those are reachable from the public API. `_loading` is the same kind of object but I left it out to keep the change minimal.

### DIFF
--- a/lib/core.ts
+++ b/lib/core.ts
@@ -279,8 +279,8 @@ export default class Ajv {
   logger: Logger
   // shared external scope values for compiled functions
   readonly scope: ValueScope
-  readonly schemas: {[Key in string]?: SchemaEnv} = {}
-  readonly refs: {[Ref in string]?: SchemaEnv | string} = {}
+  readonly schemas: {[Key in string]?: SchemaEnv} = Object.create(null)
+  readonly refs: {[Ref in string]?: SchemaEnv | string} = Object.create(null)
   readonly formats: {[Name in string]?: AddedFormat} = Object.create(null)
   readonly RULES: ValidationRules
   readonly _compilations: Set<SchemaEnv> = new Set()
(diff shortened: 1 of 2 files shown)


mahirhir:
Closing this myself: the repository has had no push in 116 days, so there is no one to review it right now. No hard feelings and no action needed. If the project picks up again and this is still wanted, I am happy to reopen or re-send it.

PR #2612, opened 2026-05-10
chore: bump fast-uri to 3.1.2

hackel:
**What issue does this pull request resolve?**
Fix for [CVE-2026-6321](https://github.com/advisories/GHSA-q3j6-qgpj-74h6) and [CVE-2026-6322](https://github.com/advisories/GHSA-v39h-62p7-jpjc).

**What changes did you make?**
Minor version bump of [fast-uri](https://github.com/fastify/fast-uri/releases) dependency.

**Is there anything that requires more attention while reviewing?**
No.

PR #2667, opened 2026-08-27
chore: upgrade fast-uri from v3 to v4

### LINKED ISSUE #2601: fast-uri - missing export URIComponent

### LINKED ISSUE #2658: CVE-2026-16221 vulnerability in fast-uri
fast-uri should be updated to 3.1.4 to fix the following security vulnerability:

https://github.com/advisories/GHSA-v2hh-gcrm-f6hx

### LINKED ISSUE #2660: Track fast-uri v4 migration (currently pinned to ^3.0.1)
**What version of Ajv are you using?**
Latest (8.x)

ajv pins `fast-uri` at `^3.0.1`. #2658 and PR #2654 track patching within the 3.x line (CVE-2026-16221, bumping to 3.1.4), but nothing tracks moving past the v3 major to fast-uri v4. Because ajv sits so deep in the dependency graph, this forces any downstream consumer that needs fast-uri v4 to carry a manual resolution/override pin instead of getting it through a normal ajv upgrade. This is the highest-impact single dependency override across our fleet — it currently affects 9 of our internal projects.

**What do you think is the correct solution to the problem?**
Open a dedicated tracking issue (separate from the 3.x patch work) for bumping to the fast-uri v4 major — acknowledging this may require a semver-major ajv release if fast-uri v4's API changed in breaking ways.

No — flagging from the downstream/consumer side; happy to help test a release candidate.

okuryu:

- Fixes: #2601 
- Fixes: #2602 
- Fixes: #2612 
- Fixes: #2654 
- Fixes: #2658 
- Fixes: #2660 

- Bumped the `fast-uri` dependency from `^3.0.1` to ~`^4.1.3`~ `^4.1.4`.
- Changed the `URIComponent` import to `import type` in `lib/compile/index.ts` and `lib/types/index.ts`, since it's used only as a type, matching the existing pattern already used in `lib/compile/resolve.ts`.

fast-uri v4 removed the previously deprecated `options`/`URIComponents` type aliases, but ajv doesn't use them, so there's no impact. The public API (`parse`/`serialize`/`normalize`/`resolve`/`equal`) is unchanged.

### DIFF
--- a/lib/compile/index.ts
+++ b/lib/compile/index.ts
@@ -14,7 +14,7 @@ import N from "./names"
 import {LocalRefs, getFullPath, _getFullPath, inlineRef, normalizeId, resolveUrl} from "./resolve"
 import {schemaHasRulesButRef, unescapeFragment} from "./util"
 import {validateFunctionCode} from "./validate"
-import {URIComponent} from "fast-uri"
+import type {URIComponent} from "fast-uri"
 import {JSONType} from "./rules"
 
 export type SchemaRefs = {
--- a/lib/types/index.ts
+++ b/lib/types/index.ts
@@ -1,4 +1,4 @@
-import {URIComponent} from "fast-uri"
+import type {URIComponent} from "fast-uri"
 import type {CodeGen, Code, Name, ScopeValueSets, ValueScopeName} from "../compile/codegen"
 import type {SchemaEnv, SchemaCxt, SchemaObjCxt} from "../compile"
 import type {JSONType} from "../compile/rules"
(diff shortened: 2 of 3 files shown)


henriknoren:
I would be nice if this was prioritized, this is preventing release to environments with strict CVE requirements

okuryu:
updated to `fast-uri@^4.1.4`

Issue #1573, opened 2021-04-28
handling of $dynamicRef on openApi v3.1 spec schema


seriousme:
**What version of Ajv are you using? Does the issue happen if you use the latest version?**
8.2.0
**Ajv options object**

```javascript
{ strict:false }
```

**JSON Schema**

```json
 {
  "type": "object",
  "properties": {
    "name": {
      "type": "string"
    },
    "in": {
      "enum": [
        "query",
        "header",
        "path",
        "cookie"
      ]
    },
    "required": {
      "default": false,
      "type": "boolean"
    },
    "schema": {
      "$dynamicRef": "#meta"
    },
    "content": {
      "type": "object"
    }
  },
  "required": [
    "in"
  ],
  "oneOf": [
    {
      "required": [
        "schema"
      ]
    },
    {
      "required": [
        "content"
      ]
    }
  ],
  "$defs": {
    "schema": {
      "$dynamicAnchor": "meta",
      "type": [
        "object",
        "boolean"
      ]
    }
  }
}
```

**Sample data**

```json
{
  "name": "limit",
  "in": "query",
  "description": "How many items to return at one time (max 100)",
  "required": false,
  "schema": {
    "type": "integer",
    "format": "int32"
  }
}
```

**Your code**

```javascript
const Ajv = require("ajv/dist/2020.js")

const ajv = new Ajv(options)

const validate = ajv.compile(schema);

console.log(validate(data));
console.log(validate.errors);
```
https://runkit.com/seriousme/ajv-dynamicref

**Validation result, data AFTER validation, error messages**

```
false
[
  {
    instancePath: '/schema',
    schemaPath: '#/oneOf/0/required',
    keyword: 'required',
    params: { missingProperty: 'schema' },
    message: "must have required property 'schema'"
  },
  {
    instancePath: '/schema',
    schemaPath: '#/oneOf/1/required',
    keyword: 'required',
    params: { missingProperty: 'content' },
    message: "must have required property 'content'"
  },
  {
    instancePath: '/schema',
    schemaPath: '#/oneOf',
    keyword: 'oneOf',
    params: { passingSchemas: null },
    message: 'must match exactly one schema in oneOf'
  }
]
```

**What results did you expect?**
I expected the validation to succeed.
https://json-schema.hyperjump.io/  says the data matches the schema.

[...]

PR #2669, opened 2026-09-05
fix: add opt-in resource-scoped dynamic references

### LINKED ISSUE #1745: Support $dynamicAnchor and $recursiveAnchor in any schema location (not only root)

jasoniangreen:

Fixes #1745 when the new `fullDynamicRefs` option is enabled. Related: #1573.

Ajv can resolve a nested dynamic anchor to the schema root instead of its declared location. This breaks the small example in the issue and some OpenAPI 3.1 schemas. It also affects generic schemas and references across schema resources. This PR fixes the reference handling; it does not add an OpenAPI-specific workaround.

Add an option, disabled by default:

```js
const Ajv2020 = require("ajv/dist/2020")
const ajv = new Ajv2020({fullDynamicRefs: true})
```

The option enables corrected `$dynamicRef` handling in `Ajv2020` and `$recursiveRef` handling in `Ajv2019`. With the option omitted or set to `false`, Ajv keeps the existing behavior, including its known limitations.

The enabled path:

- Finds anchors across each schema resource during registration. A resource is the root schema, or a nested schema with its own resource ID.
- Resolves the initial reference URI before choosing a dynamic anchor. If dynamic lookup does not apply, it uses the static target.
- Keeps outer bindings when validation enters another resource. Bindings do not leak into sibling paths or change a caller's context object.
- Shares compiled validators across pointer, anchor, and dynamic references. Forced compilation does not change whether a later plain `$ref` is inlined.
- Uses the same resolution code for normal validation and generated standalone validators.

Tests cover anonymous schemas, generic bindings, nested resources, evaluated properties/items, schema lookup and reuse, missing-schema loading, async callers with sync targets, and prototype-like anchor names.

The PR reuses the existing JSON Schema test-suite submodule. Both modern drafts now run with the option off and on. The enabled runs do not skip the cases that fail only because of the legacy reference handling. Small supplemental fixtures retain newer upstream cases. Current OpenAPI 3.1 document and `schema-base` fixtures are included, with provenance and licenses. No dependency or submodule update is needed.

For OpenAPI, the document schema and `schema-base` serve different purposes. The tests check both: the document schema can accept an embedded Schema Object without checking its keywords; `schema-base` also rejects invalid Schema Object keywords.

### API and limits

The public option name and its default are proposals for maintainer review. The default is `false` because correct dynamic scope adds work for some schemas.

This implements the published 2020-12 and 2019-09 rules, not future-dialect changes. The option does not change validation behavior in draft-07 or JTD, but their browser bundles also grow (see below). Async callers can reference synchronous targets. Async dynamic anchors and targets remain unsupported and are rejected during compilation. `compileAsync` can still load synchronous target schemas.

The existing `getValidate(KeywordCxt, SchemaEnv)` helper signature is preserved. The new option stays optional in the exported `InstanceOptions` type, even though Ajv normalizes it internally. With the option off, compiler-cache lookup and reference-call generation do not read extension-owned `resources` or `dynamicScope` metadata. Regression tests cover the old helper call with ready and deferred validators, options objects without the new property, and metadata getters that must not be accessed.

Enabling the option can change when custom keyword compilation hooks run, including for unvisited anchor definitions, and can share a validator across references to the same schema within a resource. Extensions must not depend on a particular compilation order/count or mutation of the caller's `dynamicAnchors` object. These enabled-mode implications are documented. Adding the option also adds a property to `ajv.opts`; exact snapshots of that object can change.

### Validation included in the PR

| Check | Result |
| --- | --- |
| Full spec suite | 11,727 passing; 560 pending |
| Full spec suite with `AJV_FULL_TEST=true` | 11,727 passing; 560 pending |
| Focused dynamic-reference suite | 1,146 passing |
| TypeScript build, repository formatting, and ESLint | Pass |
| Fresh `AJV_FULL_TEST=true npm run test-cov` run | 11,727 passing; 560 pending |

The coverage run reported 96.82% statements, 93.99% branches, 94.28% functions, and 97.77% lines. These are whole-suite figures, not coverage percentages for the new code alone.

The pending count includes existing unrelated exclusions repeated in the new enabled-draft runs. No new reference-related skip was added to make the enabled path pass.

The focused matrix covers 32 combinations: `allErrors` false/true; `inlineRefs` true/false/1/100; ES5/modern output; and code optimization off/on. It checks normal and standalone validators.

For the final pre-commit check, I ran `npm run json-tests`, `npm run build`, `npm run prettier:check`, `npm run eslint`, and `AJV_FULL_TEST=true npm run test-cov` separately. The local package self-reference was already configured, so I did not run the global `npm link` steps in `npm test` again.

### Extra local validation, not added to the patch

- Compared with a clean build of master at `f177fe323420ccb23e1a79445fd470cbf80aee7c`. Across 36 schema groups and 32 option combinations, omitted/false settings produced **2,112 byte-identical standalone modules** and **192 matching compilation errors**.
- Ran a separate probe against historical and current OpenAPI fixtures, the original reduction, generic examples, meta-schema use, and the complete reference cases from JSON-Schema-Test-Suite commit `c9510e3bf8a896c3cba4e08509cf752b4f30dff8`. All 456 assertions passed. This includes all 44 current dynamic-reference cases and 34 recursive-reference cases in normal and standalone modes.
- Ran the focused regression bundle in Chrome Headless 152: **1,145 passing**, with one explicit Node-only standalone-module test skipped. This was a focused browser run, not the whole browser suite.
- Reviewed generated code, cache reuse, failed-compilation retry behavior, and scope isolation. The shared-anchor regression confirms one compilation of a custom keyword, even when pointer, anchor, and dynamic references reach the same schema.

Research scripts, benchmark harnesses, duplicate full test fixtures, and historical OpenAPI files are not part of the patch. These extra results come from local checks, not from checks that this PR adds to CI. Local tests used Node 22.23.2 on macOS arm64. The results above were rerun after the compatibility follow-up at `b7b11aaba5b1ef7dcd120ac426951c372f139472`. The Node 18/20/22/24 CI matrix and Coveralls check also passed for that commit. The documentation site and the full browser suite were not built or run locally.

### Performance

Measurements used Node 22.23.2 on an Apple M2 Max. Seven fresh interleaved rounds compared clean master, omitted, false, and true settings in separate processes after the compatibility follow-up. No local builds or tests ran at the same time. Values below are medians, not production guarantees.

With the option off, generated validation code matched master in the comparison above. Valid-input throughput stayed within about 2.1% of master on the five benchmark workloads that the legacy resolver validates correctly. The earlier run was within about 0.7%; neither run proves zero overhead. Generated-code hashes for all nine benchmark schemas and all four modes also match the preceding PR implementation, including the enabled mode.

| Workload | Master: valid validations/s | Option on: valid validations/s | Option on: cold compilation |
| --- | ---: | ---: | ---: |
| Ordinary 2020-12 object | 63.07 million | 63.04 million | 0.365 ms |
| Static references | 74.38 million | 73.56 million | 0.186 ms |
| Three-node recursive tree | 14.32 million | 7.71 million | 0.314 ms |
| Original nested-anchor example | Incorrect legacy result | 15.26 million | 0.265 ms |
| Generic binding | Incorrect legacy result | 8.70 million | 0.408 ms |
| Current OpenAPI document | Incorrect legacy result | 434 thousand | 15.777 ms |
| Current OpenAPI schema-base | Incorrect legacy result | 265 thousand | 19.823 ms |

The recursive-tree case is about **1.86 times slower** with the option on. Incorrect legacy results are not used as speed comparisons. Cold-compilation timings include a new instance and dependency registration; they are not timings for an already-warmed compiler alone.

There is a larger cost for a resource placed directly inside an item loop, for example `items: {$id: "item", $dynamicAnchor: "item", type: "integer"}`. Scope can be rebuilt for every item. A separate fresh 200,000-integer array test measured **5.68 ms enabled**, **0.49 ms disabled with the same schema**, and **0.20 ms enabled without the anchor**. Scope construction is not moved outside the loop in this PR. This limit is documented. The repeated allocation does not occur with the option off.

A fresh compiler check after the compatibility follow-up used nine interleaved rounds, 500 warmups, and at least 300 ms per sample on four 2020-12 workloads. Omitted-setting medians ranged from 1.0% faster to 0.2% slower; explicit-false medians ranged from 0.8% faster to 1.7% slower. All exploratory timing intervals included zero, but some were wide enough to allow appreciable slowdowns. **Small default compilation overhead remains unresolved.** We do not claim zero overhead from these measurements.

### Browser bundle size

The option gates execution, not inclusion of compiler code. **The bundles grow even with the option omitted or false**, including draft-07 and JTD. The following sizes compare clean master `f177fe323420ccb23e1a79445fd470cbf80aee7c` with this PR at `b7b11aaba5b1ef7dcd120ac426951c372f139472`.

All sizes are bytes. Gzip means the minified bundle compressed with gzip.

| Bundle | Raw: master → PR | Minified: master → PR | Gzip: master → PR | Gzip increase |
| --- | ---: | ---: | ---: | ---: |
| Draft-07 | 269,873 → 275,511 | 127,605 → 130,459 | 36,187 → 37,021 | 834 (2.30%) |
| 2019-09 | 287,745 → 294,941 | 138,462 → 142,189 | 38,203 → 39,250 | 1,047 (2.74%) |
| 2020-12 | 288,857 → 296,053 | 139,379 → 143,106 | 38,346 → 39,408 | 1,062 (2.77%) |
| JTD | 263,138 → 268,786 | 122,396 → 125,260 | 35,844 → 36,668 | 824 (2.30%) |

Both checkouts were freshly built with the same dependencies: Node 22.23.2, Browserify 17.0.1, and Terser 5.51.2. Bundles use the repository's normal `scripts/bundle.js`; gzip uses level 9. Source map files are excluded, but the normal sourceMappingURL comment is included. These are Browserify bundle sizes, not predictions for every application bundler. Download, parsing, and startup times were not measured.

Raw and minified bundles passed smoke checks through their browser-global exports in Node's VM: ordinary validation with the option omitted/false/true, plus the original 2020-12 reproduction failing under legacy behavior and passing with the option enabled. These checks are separate from the earlier Chrome run; they are not an additional actual-browser test run. The measurement script and raw results remain outside the patch.

<details>
<summary>Additional timing details, generated-validator size, and npm package size</summary>

The compiler checks use paired 95% t intervals on log timing ratios, without adjustment for multiple testing. In the fresh run on `b7b11aaba5b1ef7dcd120ac426951c372f139472`, all intervals included zero; the widest upper bound allowed about 8.86% slower compilation. The earlier run on `b46f94811b08efb9cfe83a62e3f228c026f04bba` had two intervals excluding zero: omitted recursive compilation, about 0.25%–1.33% slower; and explicit-false generic compilation, about 0.24%–2.81% slower. The fresh run does not establish that the earlier signal is fixed. These are noisy local measurements, not proof of a specific cause or of zero regression.

The runtime benchmark warmed each validator before timing valid and invalid inputs separately. Compilation measurements used fresh instances with schema validation disabled. Built-in meta-schemas were loaded where the OpenAPI schema-base needed them. Heap measurements were also collected, but GC and JIT activity affect them. They are not allocation counts or evidence that memory overhead is zero.

The new resolver produces a larger OpenAPI schema-base validator: 320,324 bytes versus 206,573 bytes on master, which omits required checks. Sharing the compiled anchor validators reduced the new output from 325,475 to 320,324 bytes during review. It reduced the generic example from 6,078 to 5,146 bytes and the nested-anchor example from 4,310 to 3,256 bytes. These review reductions compare two versions of the new implementation, not the new code with master.

A local `npm pack --dry-run --ignore-scripts --json` check measured the package at 222,363 compressed bytes versus 217,611 bytes for clean master: an increase of 4,752 bytes, about 2.2%. Unpacked size increased by 21,010 bytes, about 2.0%. This is package size, not browser bundle size. Test fixtures are not included in the npm package.

</details>

### Prior work and credits

This implementation starts from master. No existing PR was used as its base or cherry-picked.

Thanks to `essential-randomness`, `seriousme`, `jdesrosiers`, and `epoberezkin` for the reports, reductions, explanations, and acceptance criteria; to `jasoniangreen`, `nsteve-one`, `Treasure520520`, and `aqeelat` for the earlier implementations and discussion in #2573, #2615, #2622, and #2642; and to `gregsdennis` for the generic-schema and specification explanations. Thanks also to the JSON Schema Test Suite and OpenAPI contributors for the conformance and integration fixtures.

These acknowledgements credit the work that informed this change. They do not claim that the earlier patches were incorporated or that those contributors co-authored this implementation.


jasoniangreen:
### Review guide

This guide covers `b7b11aaba5b1ef7dcd120ac426951c372f139472`, including the extension-compatibility follow-up. The PR description records validation, timing, and browser bundle measurements with their limits.

The key question is whether the corrected reference behavior and the default-off option are an acceptable trade-off. This is not an OpenAPI-specific workaround: the same resolution rules also apply to generic schemas and references across resources.

#### Why does the fix touch the compiler, not just `$dynamicRef`?

An anchor can be declared in a subschema that validation has not visited. The compiler must know which resource owns that declaration before generating the reference call. It must also preserve that resource's scope when code is inlined or reused.

| Change | Purpose | Regression to start with |
| --- | --- | --- |
| Resource indexing in `compile/resolve.ts` | Find unvisited anchors and distinguish resources with different base URIs. | [Unvisited anchors](https://github.com/ajv-validator/ajv/blob/b7b11aaba5b1ef7dcd120ac426951c372f139472/spec/full-dynamic-refs.spec.ts#L92), [shared objects under different base URIs](https://github.com/ajv-validator/ajv/blob/b7b11aaba5b1ef7dcd120ac426951c372f139472/spec/full-dynamic-refs.spec.ts#L499). |
| Scope generation in `compile/dynamic.ts` | Keep outer bindings without changing the caller's scope or leaking bindings into siblings. | [Sibling item loops](https://github.com/ajv-validator/ajv/blob/b7b11aaba5b1ef7dcd120ac426951c372f139472/spec/full-dynamic-refs.spec.ts#L411), [caller context](https://github.com/ajv-validator/ajv/blob/b7b11aaba5b1ef7dcd120ac426951c372f139472/spec/full-dynamic-refs.spec.ts#L480). |
| Compiled-target reuse in `compile/index.ts` | Compile the same anchor once without changing a later plain `$ref` inlining decision. The cache belongs to the schema resource, not the process. | [Compilation count and lookup identity](https://github.com/ajv-validator/ajv/blob/b7b11aaba5b1ef7dcd120ac426951c372f139472/spec/full-dynamic-refs.spec.ts#L272), [inlining order](https://github.com/ajv-validator/ajv/blob/b7b11aaba5b1ef7dcd120ac426951c372f139472/spec/full-dynamic-refs.spec.ts#L312). |
| Static resolution before dynamic selection in `dynamicRef.ts` | Use dynamic lookup only when the initial target permits it; preserve evaluated-property/item tracking through the selected validator. | Existing upstream reference cases in the matrix; runtime-selected [properties](https://github.com/ajv-validator/ajv/blob/b7b11aaba5b1ef7dcd120ac426951c372f139472/spec/full-dynamic-refs.spec.ts#L329) and [items](https://github.com/ajv-validator/ajv/blob/b7b11aaba5b1ef7dcd120ac426951c372f139472/spec/full-dynamic-refs.spec.ts#L388). |

#### What about custom keywords and existing integrations?

The follow-up preserves `getValidate(KeywordCxt, SchemaEnv)` and keeps `fullDynamicRefs` optional in the exported `InstanceOptions` type. With the option off, cache lookup and reference-call generation do not read the new `resources` and `dynamicScope` fields. [Four regression tests](https://github.com/ajv-validator/ajv/blob/b7b11aaba5b1ef7dcd120ac426951c372f139472/spec/full-dynamic-refs.spec.ts#L140) cover these boundaries, including ready/deferred validators and throwing metadata getters.

With the option on, anchor compilation order/count, validator reuse, and caller-scope mutation can differ from legacy behavior. Extension authors should not rely on those legacy details when enabling it. This is documented; preserving the old helper call does not promise that every internal behavior stays unchanged.

#### Does default-off mean no cost?

No. The extra generated scope operations are disabled. The refreshed local comparison found 2,112 byte-identical standalone modules with the option omitted/false, and validation throughput stayed within about 2.1% of master on the five legacy-correct benchmark workloads. Small compilation overhead remains unresolved; the PR description includes both the fresh timing results and the earlier compiler findings.

The compiler code is still bundled: minified-and-gzipped browser bundles grow by 824–1,062 bytes (2.30–2.77%), including draft-07 and JTD. The description records all sizes, tool versions, and measurement limits. The option does not remove this download-size cost.

#### Why leave scope allocation inside item loops?

Moving it outside the loop needs a separate design for scope lifetime across nested loops and conditional paths. This PR keeps scope isolation explicit and documents the cost. The refreshed 200,000-item case measured 5.68 ms enabled versus 0.49 ms disabled with the same schema. The repeated scope allocation is absent with the option off. Hoisting remains a possible follow-up, not a completed optimization.

#### What remains unsupported?

Async dynamic anchors and targets are rejected during compilation. Async callers with synchronous targets are supported and tested, as is loading missing synchronous schemas with `compileAsync`. The option name and default remain proposals for maintainer review.

PR #2672, opened 2026-09-20
fix: evaluate additionalProperties first when removeAdditional is failing

### LINKED ISSUE #897: Wrong results with {removeAdditional: 'failing', allErrors: false} and additionalProperties as failing schema

dyk1454683243-sudo:

Fixes #897

**What changes did I make?**

With `removeAdditional: "failing"`, evaluate `additionalProperties` before other object keywords (`required`, `minProperties`, `maxProperties`).

That restores the intended behaviour when `allErrors` is `false` and `additionalProperties` is a failing schema:

1. Extra properties that fail their schema are removed even if a later keyword such as `required` would fail.
2. Errors for those removed properties are discarded; `validate()` stays consistent with `errors`.
3. Property-count keywords see the data after extras are removed, so `minProperties` / `maxProperties` cannot pass or fail against the pre-removal object.

Regression tests cover the original #897 report and the min/maxProperties follow-on cases.

Keyword evaluation order changes only when `removeAdditional === "failing"`. The default order (no `removeAdditional`) is unchanged.


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

